PR #7, opened 2016-06-19
Enabled mouse scrolling and disabled downsampling of thumbnails

mrMoe:
greetings
moe

### DIFF
--- a/pdfshuffler/pdfshuffler.py
+++ b/pdfshuffler/pdfshuffler.py
@@ -320,8 +320,8 @@ class PdfShuffler:
         cnt_all = 0
         for row in self.model:
             cnt_all += 1
-        if row[1]:
-            cnt_finished += 1
+            if row[1]:
+                cnt_finished += 1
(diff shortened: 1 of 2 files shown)


jeromerobert:
Your patch contains too much noise so it cannot be read. Please let your commit contains only the changes you describes in the commit description (no white space or reindent changes).

mrMoe:
should I break the pull Request into three for 
Disabling downsampling,
Enabled mouse scrolling,
Rreduced newlines and Corrected indentation
?

jeromerobert:
Yes that would be better. Please don't do changes in new lines or indentations, because I want this repo to stay as close as possible as https://gna.org/projects/pdfshuffler/. Let Konstantinos Poulios (the original author) manage the source code format.

jeromerobert:
I'm back again on that PR. Those features are nice but I'm really unable to find the actual changes in the commit as they are too noisy. As it now quite old I close it. May you find time and energy to make it more readable.

PR #10, opened 2016-08-31
Reverse order

julianfoad:
Hi, Jerome. I found a need to reverse the order of a stack of bank statements that I had scanned, where I had kept the paper copies in reverse order, so I have written this patch to do it in PdfShuffler. I think it would be a useful addition.

It adds an item under 'Edit' menu and in the popup menu called 'Reverse Order', which is sensitive when multiple contiguous pages are selected, and it reverses their order.

I think bug fixes relating to selection and drag-and-drop are more important than this, and have been trying to work on those too, but this reverse-order enhancement is the first patch I am happy with.

Please let me know if this is something you could accept. Thanks.

### DIFF
--- a/data/pdfshuffler.ui
+++ b/data/pdfshuffler.ui
@@ -163,6 +163,16 @@
                         <signal name="activate" handler="clear_selected" swapped="no"/>
                       </object>
                     </child>
+                    <child>
+                      <object class="GtkImageMenuItem" id="imagemenuitem_reverse_order">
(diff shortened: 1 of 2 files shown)


julianfoad:
I have now submitted this as https://gna.org/support/index.php?3392 .

julianfoad:
Sorry, I thought "unofficial git mirror" implied you might be acting as an Github gateway to the project, but then I found your comment in https://github.com/jeromerobert/pdfshuffler/issues/9 where you say not. Closing.

PR #15, opened 2018-10-10
Created Debian/bionic

juanma1980:
Hi.
I've also forked pdfshuffler and added some changes:
- First I've renamed debian branch to debian/bionic. This packages the project for Ubuntu bionic but if you want packages for debian stable, unstable..etc... then the branches must be named debian/unstable, debian/stable and so on... This way is easiest to build packages for any distro/flavour. 
- Second I've updated "ca" and "es" translations and added a "generate_po.sh" that updates the pot file from the sources
- Third I've merged master and debian branches as the source was not the same in both branches (if sources aren't the same gbp buildpackage will fail when comparing the two branches)

I know that these changes may not go the way to your repo but I need them in order to build the package for the LliureX Distribution. Feel free to include them or not.

Cheers

Juanma1980.


jeromerobert:
That's too much stuff for a single PR. I just merged the valencian translation. see 2c36af70b1d33c6df6b6aef4dbde34f70917e54f

PR #25, opened 2018-11-08
Add .gitignore

dreua:


jeromerobert:
I think this is a too wide ignore. The only line which currently apply to this repo is `__pycache__`. Some are (currently) out of topic, other (like `.idea`) should go to `.git/info/exclude`.

dreua:
`ip install --user -e .` created some more files I wanted to ignore, so I just went for some template from the internet. But yeah, I agree it might have been a bit too much.

PR #26, opened 2018-11-08
Fix bytes conversion

dreua:
Remove `chooser.destroy()` where chooser is not defined
Remove pointless statement `self.model`

Fix the following error when dropping a page into the scrollview:
```
  File "[...]/pdfshuffler/pdfshuffler.py", line 930, in sw_dnd_received_data
    data = data.split('\n;\n')
TypeError: a bytes-like object is required, not 'str'
```

Use SelectionData::get_uris() and decode() where appropriate.
Remove now unneeded decode() call in `get_file_path_from_dnd_dropped_uri`.

I did various tests with Python 2 and Python 3 on Linux and found no errors.

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -927,7 +927,6 @@ class PdfShuffler:
 
         data = selection_data.get_data()
         if target_id == self.MODEL_ROW_EXTERN:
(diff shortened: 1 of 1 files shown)


jeromerobert:
Manually rebased

PR #27, opened 2018-11-08
Improve handling of SIGINT

dreua:
Up to now you couldn't stop pdfarranger by pressing Ctrl+C in the terminal or your IDEs Stop button, this PR fixes this for better user experience and faster development.

Tested with Python 2 and 3 on Linux.

`GLib.unix_signal_add` suggests that it might not work on Windows (Let's hope it's just a silent noop in this case). Furthermore I don't know if such functionality is needed or wanted on Windows, what do you think @carlos22 ?


jeromerobert:
Manually rebased and merged

PR #28, opened 2018-11-11
Flatpak manifest

Johnn3y:
Hi,

I wrote a [Flatpak](https://flatpak.org/) manifest. It contains all required dependencies and it works, but it's improvable. This could be the first step to add pdfarranger to [Flathub](https://flathub.org/) some time.


jeromerobert:
Hello @Johnn3y and thank you. Here are some questions (some of them because I'm not familiar at all with Flatpak):
* Could you confirm that openjpeg is a requirement of poppler (I never built it) ?
* Do you mind if I switch to Yaml (to add comments)
* Wouldn't be better to choose well tag versions of each dependencies instead of SHA1 ?
* And tarball instead of git ?
* After merging your commits, should I change `com.github.Johnn3y.pdfshuffler` ? This is just an ID for FlatHub right ?

Johnn3y:
* It's not  necessary as far as I know, but it is recommended (haven't try build without it). I don't know how it affects pdfarranger.
* No, I don't mind.
* Yes.
* And Yes. 
* Yes it's a [general convention](http://docs.flatpak.org/en/latest/conventions.html#application-ids) in Flatpak and [nessessary](https://github.com/flathub/flathub/wiki/App-Requirements#application-id) for Flathub so you should change it to `com.github.jeromerobert.pdfarranger` or something. And you should replace everything else with `pdfshuffler` to `pdfarranger` and use your git repository (or archive) as source.

PR #56, opened 2019-02-21
Needs input: Disable strict mode for PdfFileReader

dreua:
With strict = False exporting certain pdf files (e.g. https://github.com/jeromerobert/pdfarranger/issues/54) works. 
Warnings are printed to stdout so they can be easily missed and are not visible at all when the application is not started from a command line.

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -636,7 +636,7 @@ class PdfArranger:
         pdf_output = PdfFileWriter()
         pdf_input = []
         for pdfdoc in self.pdfqueue:
-            pdfdoc_inp = PdfFileReader(open(pdfdoc.copyname, 'rb'))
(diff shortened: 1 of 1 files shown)


jeromerobert:
Shouldn't we investigate the `warndest` and/or `overwriteWarnings` parameters to have those warnings into a pop up dialog ? Of course this would only be reasonable if there are not too many of them.

dreua:
Yes, I'm on that, but either it's not that straightforward or it's broken.

jeromerobert:
It's broken. <https://github.com/mstamy2/PyPDF2/blob/master/PyPDF2/pdf.py#L1129>  change the default argument but this default argument is never used. The `warnings` python module always call `showwarning` with `file=None`. Anyway a library which modify the default of another library is a violation of many programming rules. I will try an other way.

PR #62, opened 2019-03-29
change default version to latest tag

zocker-160:
I think it makes sense to use the latest tag instead of the master branch. For me personally the version from master didn't work, but the 1.1.1 did.

I have tested it with the new links and it works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ sudo dnf install poppler-glib python3-distutils-extra python3-pip python3-gobjec
 Then:
 
 ```
-pip3 install --user -r https://raw.githubusercontent.com/jeromerobert/pdfarranger/master/requirements.txt
+pip3 install --user -r https://raw.githubusercontent.com/jeromerobert/pdfarranger/1.1.1/requirements.txt
 ```
 
 # For developers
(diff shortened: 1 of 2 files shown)


dreua:
I personally would be annoyed by having to change these links on each release and try this: https://help.github.com/en/articles/linking-to-releases#linking-to-the-latest-release

zocker-160:
ok fair enough, you can close the PR then

dreua:
No, let's wait what Jerome thinks about it, I think it is a good idea but it could be improved by automatically linking to the latest release. (I can't close this anyway.)

dreua:
Oh, and I just realized that we should also try to fix the current master. Could you provide some information what's wrong?

zocker-160:
sure I'll create a new issue for that

zocker-160:
I've created a script based on [this](https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c)

```
RELEASE=`curl --silent "https://api.github.com/repos/jeromerobert/pdfarranger/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/'`

wget https://raw.githubusercontent.com/jeromerobert/pdfarranger/$RELEASE/requirements.txt
```
sounds like a better solution for me, what do you think?

jeromerobert:
I prefer to let master as default version. master is supposed to work smoothly. Without that you would not have found #63 and 1.2.0 would have been released with a critical bug.

PR #70, opened 2019-05-09
Update data files

fabaff:
Would no longer require to carry a patch for the Fedora package.

Thanks

P.S. Sorry, #69 was incomplete.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,9 +26,17 @@ import re
 from setuptools import setup
 from DistUtilsExtra.command import *
 
-data_files=[('share/applications', ['data/pdfarranger.desktop']),
-            ('share/pdfarranger', ['data/pdfarranger.ui']),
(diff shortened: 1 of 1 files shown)


jeromerobert:
Icon are installed by DistUtilsExtra so I prefer #69. Merged as 72fd589, closing.

fabaff:
The icons may be installed by DistUtilsExtra but they are not available during the RPM build process. It seems also that DistUtilsExtra is responsible that the l10n stuff can be picked-up by the RPM helper scripts.

jeromerobert:
Yes DistUtilsExtra is also responsible for the I10n. Is it also not working during the RPM process ? Is installation with `pip3 install --user .` also broken (i.e. icon missings) ?

fabaff:
No, a wild guess is that DistUtilsExtra is doing too much already.

PR #93, opened 2019-07-18
Update de.po

Etamuk:
Hello,
I did a update to this po. There were some little grammar issues and missing marks (f.e. "-").
Best regards from germany.
By the way:
**#: pdfarranger/pdfarranger.py:1166 : msgid "% of width" and msgid "% of height"**
are translated in po-file but not shown as translated text in UI.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -10,8 +10,8 @@ msgstr ""
 "Project-Id-Version: PDF-Shuffler 0.7\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2019-02-03 17:29+0100\n"
-"PO-Revision-Date: 2017-02-26 16:16+0000\n"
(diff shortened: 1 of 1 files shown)


jeromerobert:
Thanks for the update.
I have it at line 1043: https://github.com/jeromerobert/pdfarranger/blob/master/pdfarranger/pdfarranger.py#L1043
And for me it's shown translated:
Untitled

Etamuk:
Hi,
ok. Strange. 
This is what I see:
Bildschirmfoto von 2019-07-18 22-42-16
Also the text in top of the window is missing here.

jeromerobert:
> I am using version 1.1.1 of pdf arranger from official Ubuntu Repo. Maybe not the lastest version(?)

Ok it does explains why we have differences. I redesigned this dialog box in 1.2.0.

Issue #95, opened 2019-07-23
Feature Request: Save selection to seperate file


kheaactua:
It'd be great if I could select a few pages and save them to a PDF, select another range, save it to another.

This would fantastically speed up my scanning:  right now I scan mail one letter at a time, then use PDFShuffler to remove some pages in each resulting PDF.  If I could just scan all my mail at once, and then use this app to partition the mail into separate files, that'd be great.

Unless I missed this functionality somehow, I tried selecting a range, but when I pressed save it saved everything.

Thanks!

dreua:
It is possible via the right click menu, Export selection. 
The option is missing in the edit menu, so thank you for reporting this!

kheaactua:
Oh nice, even just this should save me a bunch of time!

dreua:
I was about to add this function to the edit menu but now I'm wondering: Should it be in the file menu rather then in the edit menu (that's where you would have seen it, right?) or even have it's own icon (is there some icon we could use?). For power-users a keyboard shortcut would allow for the highest speedup. (Which would be suitable?)

kheaactua:
I think a keyboard shortcut regardless would be great in any case.

But, for an item, to me it would be intuitive if this, say, `Export Selection` was under the file menu.  An icon is cool too.

dreua:
Luckily I just implemented it the way you said before seeing your comment. Haven't looked for an icon though and now I switched to a train without WiFi :/ Feel free to test my PR!

kheaactua:
Hey, I tried it out, and it seemed to work.

It did fail on one PDF though, it was a PDF of some unit testing article.  The error was:

![error](https://user-images.githubusercontent.com/2363406/61891519-d10cf600-aed7-11e9-9040-94225d9aa389.png)

Maybe the PDF is weird?  I can upload it too if you want to look into it.

jeromerobert:
@kheaactua I already fixed bugs like that but I guess you are not using the master version. Anyway, could you open a new bug and attach the culprit PDF, so I could confirm the bug is already fixed ?

kheaactua:
Yep, I was using 5503d73d44fb85c84818b4c3168a9c0e5e15b69c on dreua/export_selection

Sure, I'll create a new issue now

PR #96, opened 2019-07-23
Add export selection menu item to file menu and add shortcut.

dreua:
I think Ctrl+e is the obvious choice for a shortcut here.
The file menu seems to be more appropriate than the edit menu
since people want to save something.
I also added the export to the list of functions that are
disabled when nothing is selected.

Inspired by #95


dreua:
I forgot to test the translation of the new menu entry. Maybe it should also be reworded to be more in line with the other file menu items.

PR #107, opened 2019-09-08
Add feature: Duplicate selection

dreua:
Closes https://github.com/jeromerobert/pdfarranger/issues/18

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -235,6 +235,7 @@ class PdfArranger(Gtk.Application):
         self.window.add_action_entries([
             ('rotate', self.rotate_page_action, 'i'),
             ('delete', self.clear_selected),
+            ('duplicate', self.duplicate),
             ('crop', self.crop_page_dialog),
             ('export-selection', self.choose_export_selection_pdf_name),
             ('reverse-order', self.reverse_order),
@@ -911,7 +912,7 @@ class PdfArranger(Gtk.Application):
         selection = self.iconview.get_selected_items()
         ne = len(selection) > 0
         for a, e in [("reverse-order", self.reverse_order_available(selection)),
-                     ("delete", ne), ("crop", ne), ("rotate", ne),
+                     ("delete", ne), ("duplicate", ne), ("crop", ne), ("rotate", ne),
                      ("export-selection", ne)]:
             self.window.lookup_action(a).set_enabled(e)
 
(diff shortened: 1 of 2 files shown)


pdfarranger/pdfarranger.py
```diff
@@ -1089,6 +1090,29 @@ def crop(self, selection, newcrop):
...
+        """Duplicates the selected elements"""
+        self.set_unsaved(True)
+        self.undomanager.commit("Duplicate")
+
+        model = self.iconview.get_model()
+        selection = self.iconview.get_selected_items()
+        # selection is a list of 1-tuples, not in order
+
+        # adapted from iv_dnd_receive_data / MODEL_ROW_INTERN
+        before = False
```
jeromerobert: Unless I missed something `before` should be removed and the code below adapted. Right ?

dreua: Yeah, I forgot about this: There are different ways on where the duplicated page(s) can be inserted and I don't know which one we want. Let's say a, b, c are the selected pages and A, B, C are their duplicates. 1, 2, 3 are just other pages. With before = False we currently get this behavior:
1 a 2 b 3 c --duplicate--> 1 a 2 b 3 c A B C
With before = True we could easily switch to:
1 a 2 b 3 c --duplicate--> 1 A B C a 2 b 3 c
(There are even more possibilities, for example 1 a A 2 b B 3 c C...)
 @darkdragon-001 are you still interested in this feature and have ideas about this?

darkdragon-001: My original concern was about duplicating a single page at a time and didn't think about multiple pages.

My first thought for your example would be `1 a A 2 b B 3 c C` but I can imagine that the more commonly wanted behavior for _continuous_ page ranges would be to insert the whole bundle before/after (the result is the same).
I would suggest that your case (when somebody wants _non-continuous_ pages bundled together) would be to _export_ them to a new pdf file.

dreua: For the export use-case the "export selection" function would be the superior solution compared to duplicating, deleting the rest and exporting, so actually I can't think of a common use case of duplicating multiple pages one way or the other.

Since duplication of single pages works either way, I'd suggest we keep the functionality as is, I clean the code up and we can close this issue?

jeromerobert: I agree with @darkdragon-001, `1 a A 2 b B 3 c C` is the way to go as it is the most intuitive. I would say that there are 3 ways to consider that feature (from the less to the most generic):

1. Only allowed for single pages
2. Only allowed for continuous pages rages
3. `1 a A 2 b B 3 c C`

> Since duplication of single pages works either way, I'd suggest we keep the functionality as is, I clean the code up and we can close this issue?

I would prefer that 2 or 3 get implemented first. I could do it later if you are too busy (and keep your PR on hold for the time being).

darkdragon-001: I agree with @jeromerobert that 3 should be the standard implementation since it is most intuitive and will always work.

Optionally you could implement a case switch: if continuous page range then use currently implemented behavior of before/after, else use default option 3 above.

dreua: From @jeromerobert 's options I like 3 the best. I guess I can implement it on the weekend, but feel free to take over. Currently I'm busy with the pdfarranger package for Fedora.
I'm not that fond of the case switch because it might be more confusing than useful for users.
BTW: Pdfarranger-power-users can still use `ctrl` + drag and drop to get their ... A B C ... wherever they like ;)

dreua:
I changed the behavior as discussed in the review above and rebased on the current master.

dreua:
Reviewing my own code I noticed some variable names didn't make sense any more. Cleaned that up and squashed it all together.

PR #116, opened 2019-10-10
WIP: Refactor unused parameters and formatting

dreua:


pdfarranger/pdfarranger.py
```diff
@@ -429,7 +429,7 @@ def do_activate(self):
         self.__create_actions()
 
     @staticmethod
-    def set_cellrenderer_data(column, cell, model, it, data=None):
+    def set_cellrenderer_data(_column, cell, model, it, _data=None):
```
dreua: We need those =None for functions which get called not only from Gtk code but also from our own code.

pdfarranger/pdfarranger.py
```diff
@@ -470,7 +485,7 @@ def render(self):
 
         if self.progress_bar_timeout_id:
             GObject.source_remove(self.progress_bar_timeout_id)
-        self.progress_bar_timout_id = \
+        self.progress_bar_timeout_id = \
```
dreua: This code was broken since 2011. I don't see any change in behavior though

dreua:
This my biggest pull request so far, please let me know if I should do anything to make it easier to review.

dreua:
I just noticed _ is used for string translation so I need to revert the cases where it's used as a dont care variable. Don't have time for this now but need to remember.
I will also try to clean up the import section a bit but I belive it's not possible to get it warning-free.

pdfarranger/pdfarranger.py
jeromerobert: I now get:

```
pdfarranger.py:487: Warning: Source ID 18 was not found when attempting to remove it
  GObject.source_remove(self.progress_bar_timeout_id)
```

I will investigate before merging your PR.

jeromerobert:
Merged: ceb730b

PR #117, opened 2019-10-15
Update Japanese translation

sicklylife-jp:


jeromerobert:
Merged as e705e4b. thanks.

PR #128, opened 2020-01-21
Fix German translation

timoschwarzer:
Build was failing because of `msgfmt` not being happy with the comment between `msgid` and `msgstr`.

```
msgfmt po/de.po -o build/mo/de/LC_MESSAGES/pdfarranger.mo
po/de.po:37: missing 'msgstr' section
po/de.po:39:7: syntax error
```

btw @dreua: "CreatorTool" specifies the tool the PDF file was generated with, so "Erstellungswerkzeug" should be fine :)

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -35,7 +35,6 @@ msgstr "Hersteller"
 
 #: pdfarranger/metadata.py:33
 msgid "Creator tool"
-# not sure what that is and how to translate it
 msgstr "Erstellungswerkzeug"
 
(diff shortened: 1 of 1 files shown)


dreua:
Oh ,I am sorry for not testing my PR. Seemed so trivial but of course one should always test. Maybe we should think about a CI with at least an automated builld?

timoschwarzer:
@dreua yeah, CI builds would be great!

A simple Travis config which builds the project should be sufficient for the beginning.

Issue #130, opened 2020-02-01
Edit properties: Changes to current field lost when clicking OK


dreua:
It seems that the field content is not saved when clicking ok, only when pressing enter or clicking another field the content is saved.

dreua:
Additionally the Creator field must be formatted as a list of strings (e.g. `["aaaa-Creator", "asdfasf"]`) or it will not be saved. Is this intended behaviour or am I hitting a bug here?

And one more thing: The Producer field seems to be overwritten to `pikepdf 1.8.3` not matter what I enter. Thats no big deal for me but maybe we shouldn't offer it for editing at all then.

dreua:
For the record: Writing the Producer meta data has been fixed in https://github.com/jeromerobert/pdfarranger/commit/e39770541dc6c4fc0bcb21517392a348b53c106b by @jeromerobert , thank you - not only for that fix but since I've been struggling to find the right place to say that I'll do it right here, right now: Thank you so much for taking care of pdfarranger! It's great to see this tool developing from the python2 death sentence to getting more features, fixes and being actively maintained. :+1: :+1: :+1:

PR #136, opened 2020-02-08
 Always save edited values of metadata 

dreua:
I wish PyGtk would have a better way to do this, but I'm quite sure that is the best we con do. It is not possible to cancel a single edit by pressing escape but since you still have the cancel button to cancel the whole edit I think that is no big deal. It definitely feels better for me if "Apply" is not reverting your current edit any more.
Looking forward to the 1.4.1 release 

Fixes: #130

### DIFF
--- a/pdfarranger/metadata.py
+++ b/pdfarranger/metadata.py
@@ -80,7 +80,7 @@ def _metatostr(value, name):
 
 
 def _strtometa(value, name):
-    if len(value) == 0:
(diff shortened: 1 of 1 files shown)


jeromerobert:
Thank you ! I would not have been able to find that.

1.4.1 is now just waiting for the Windows bundle, most probably this evening or tomorrow evening.

Issue #140, opened 2020-02-09
cut/copy/past actions


kbengs:
Would be nice to have option to cut / paste selected pages when right clicking. It is now a bit hard to move for example the last page to the beginning when there is a lot of pages..

jeromerobert:
Yep, this was on my (hidden) to do list.

The problem is not that we don't have cut/copy/paste actions in the menu, the problem is that we don't have them at all. The current drag & drop must be reimplemented to separate the drag & drop behavior from the cut/copy/paste code, so it could be reused in menus (and have ctrl+cxv shortcuts).

It would also make #67 easier to implement.

dreua:
Before thinking about code I see a UX problem here: We currently do not have a cursor *between* two pages but only a selection of pages. Think of text editing: The cursor is usually not *on* but *between* two characters and you can move it in front of the first and after the last character. We could decide to paste pages before or after the current selection but either way it wouldn't be possible to paste as the first and the last page.

jeromerobert:
As I'm not that comfortable with UX I was thinking to something like:

* A *past* action (either before or after but not both) which would be bound to `ctrl+v`
* A *special past* action which would manage before, after, and interleave (for #67)

I don't know yet if special past should popup a dialog or be split into different actions (into a submenu).

kbengs:
One way of doing it using partly already available functions (I suppose) would be first to do a modified version of the "export selection" that would save selected pages to temporary location (without having to supply a place and file location). Then delete selected pages. Then you would select the page where you would like to paste. Then choose "paste", and it would paste for example after selected page.

Another thing a bit related to this, I would suggest in pdfarranger.py on line 853 "autoscroll_area = 40" would be increased to 80.

kbengs:
I have been working on this. If someone wants to see it is in kbengs/pdfarranger/kben/raw fork/branch. It paste after selected item(s) (after last selected item). Paste to other instance does not work. But otherwise it seems to work. Should I continue or stop? :)

dreua:
I just gave it a try and it definitely feels like a good addition! :+1:  Like I mentioned before we now have the problem, that you can't paste something before the first page. I wonder if we could add a selectable dummy before the first page that could be selected by arrow buttons or clicking so that pages pasted there would end up at the beginning. (I may be aiming high here, a "paste before" action would be probably easier to implement and sufficiently usable when mapped to Ctrl+Shift+V.)

[...]

PR #143, opened 2020-02-16
Drag pages(s) from one instance to empty instance now possible

kbengs:
Also autoscroll area increased

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -850,7 +850,7 @@ class PdfArranger(Gtk.Application):
     def iv_dnd_motion(self, _iconview, _context, _x, y, _etime):
         """Handles the drag-motion signal in order to auto-scroll the view"""
 
-        autoscroll_area = 40
+        autoscroll_area = 80
         sw_vadj = self.sw.get_vadjustment()
         sw_height = self.sw.get_allocation().height
         if y - sw_vadj.get_value() < autoscroll_area:
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -795,11 +795,10 @@ class PdfArranger(Gtk.Application):
                 if len(model) > 0:  # find the iterator of the last row
                     row = model[-1]
                     ref_to = Gtk.TreeRowReference.new(model, row.path)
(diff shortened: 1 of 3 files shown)


jeromerobert:
Thank you ! Your `else` block is almost a copy/past of the `elif target == 'MODEL_ROW_EXTERN'` block which is just above. I think both should be merged.

pdfarranger/pdfarranger.py
```diff
@@ -798,39 +798,41 @@ def iv_dnd_received_data(self, iconview, context, x, y,
             if ref_to:
```
dreua: Is this if clause still useful?

pdfarranger/pdfarranger.py
```diff
@@ -798,39 +798,41 @@ def iv_dnd_received_data(self, iconview, context, x, y,
...
-
+                while data:
+                    tmp = data.pop(0).split('\n')
+                    filename = tmp[0]
+                    npage, angle = [int(k) for k in tmp[1:3]]
+                    crop = [float(side) for side in tmp[3:7]]
+                    pageadder.addpages(filename, npage, angle, crop)
+                if pageadder.commit() and context.get_actions() & Gdk.DragAction.MOVE:
+                    context.finish(True, True, etime)
+
```
dreua: Unnecessary whitespace. I personally can't see it but Codacy can and it's unhappy about it.

pdfarranger/pdfarranger.py
kbengs: I tested to remove it and also on line 823 and dragged some pages and I dont see any problem with it. So maybe it does not need to be there..

jeromerobert:
Hello kbengs. Why did you closed that PR ? IMHO it was good.

kbengs:
Hello I found out that it was still a problem with it, when dropping to a a empty instance the pages came in wrong order. So I though I should delete this fast before it gets merged. :)
And I though I would make a PR from #140. Drag to empty instance is also possible there. Have you looked at bengs/pdfarranger/kben/raw fork/branch? Should I make a PR from that?

jeromerobert:
Ok I got it. See my comment on #140.

PR #145, opened 2020-02-16
Run Actions on pull request

dreua:
I don't think that there are any credentials or anything else in the build process or the build container that could be exploited by malicious pull requests therefore this will be only beneficial to catch errors even earlier.


dreua:
It is strange that the work flow already runs on this pull request itself. Isn't it?

PR #147, opened 2020-02-24
Fix crop / rotation on duplicated pages

dreua:
Fix: https://github.com/jeromerobert/pdfarranger/issues/141


jeromerobert:
Brillant. On the way to 1.4.2.

PR #162, opened 2020-03-29
Preserve selection when dropping dragged pages

kbengs:
All other actions (crop, duplicate, ..) preserve selection when action is done. I think also drag and drop should do that. It is more easy to see where the pages went when they are selected.


jeromerobert:
Merged as 8f69a14

PR #171, opened 2020-04-12
Add menu.ui to installation

dreua:


dreua:
I needed this to make my local pip install work.

PR #177, opened 2020-04-14
Set outline css_data only for iconview

kbengs:
Sometimes when selecting import the thick dotted line appear:
Dotted line

..and when selecting edit properties this appear:
Dotted line 2

I saw this css sideeffect now when i tested the released 1.5.0


dreua:
Funny, I noticed these outlines but thought they were just a quirk in Gtk or something, not a big deal imo but good to fix anyway. :+1:

Issue #178, opened 2020-04-15
Error handling for non-PDFs


dreua:
When opening a file that is broken or not a PDF at all, the following error gets printed to command line:
```
Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/pdfarranger/pdfarranger.py", line 928, in sw_dnd_received_data
    msg = pageadder.addpages(filename)
  File "/usr/lib/python3.7/site-packages/pdfarranger/pdfarranger.py", line 1210, in addpages
    pdfdoc = PDFDoc(filename, self.app.tmp_dir)
  File "/usr/lib/python3.7/site-packages/pdfarranger/pdfarranger.py", line 1177, in __init__
    self.document = Poppler.Document.new_from_file(uri, None)
gi.repository.GLib.Error: poppler-quark: PDF document is damaged (4)
```
This is confusing to users (if they see it at all) and additionally this triggers Fedora's automatic bug reporting tool ([ABRT](https://developer.fedoraproject.org/tools/abrt/about.html)).
Note that this can be triggered via drag & drop, by command line argument and by the "add file" button. (In the last case, a warning dialog is shown.)

Originally reported by [Fedora user  sukhamay ganguly (bug not publicly visible)](https://bugzilla.redhat.com/show_bug.cgi?id=1824017) who tried to open a picture in pdfarranger (which is a fair thing to do imo, see #45)

### Solution
Show a proper warning dialog and print a user and abrt friendly message to the console in all three possible ways to add a file:

- [x] Drag and Drop
- [x] Add file
- [x] Command line argument (alternatively don't show gui at all, print only to console and exit)

Related: #101 / https://github.com/jeromerobert/pdfarranger/commit/53cc05f7408f71c0da86651b01d5cc307f5fad2b

PR #184, opened 2020-04-17
Fix paste exception type and use private clipboard in linux

kbengs:
The clipboard data is only useful to pdfarranger, so I think it is best if it is not visible to other applications. We can also cause many errors by pasting to a text editor and manipulate the data, and paste that data back to pdfarranger. (#178 can also happen this way)

So this line:

self.clipboard = Gtk.Clipboard.get(Gdk.Atom.intern('_SELECTION_PDFARRANGER', False))

gives a "private" clipboard that can not be seen by other applications.
But unfortunately is does not work in windows. It doesn't throw any exceptions or anything, the clipboard just seems to be empty. So that's why windows does not use the same clipboard.

Could the reason be that some library or something is missing in the windows installation? Now when I have tested this in windows I have just replaced the pdfarranger.pyc with pdfarranger.py.

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -831,7 +831,7 @@ class PdfArranger(Gtk.Application):
         while data:
             try:  # Clipboard can contain not expected data
                 self.data_to_pageadder(data, pageadder)
-            except:
+            except (ValueError, FileNotFoundError):
                 return False
         return pageadder.commit(select_added, add_to_undomanager=True)
 
@@ -848,7 +848,7 @@ class PdfArranger(Gtk.Application):
         while data:
             try:  # Clipboard can contain not expected data
                 self.data_to_pageadder(data, pageadder)
-            except:
+            except (ValueError, FileNotFoundError):
                 return
 
             pageadder.move(ref_to, before)
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -272,7 +272,10 @@ class PdfArranger(Gtk.Application):
         self.export_file = None
 
         # Clipboard for cut copy paste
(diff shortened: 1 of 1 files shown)


kbengs:
@jeromerobert 
I have built the msi installer on a Windows 7 virtual machine. (following the win32.md guide) Installation of msi seems to go fine.  But when I try to start pdfarranger this message comes up:
Error when start
Do you know what I should do to solve it?
My intention was to test if the "private" clipboard could work in Windows when everything is rebuilt. My idea was that maybe the Windows installation is missing something so that it does not know what Gdk.Atom.intern means)

jeromerobert:
* `/mingw64/bin/python3 -m pip install --user pikepdf` should have created `c:/Users/Win7Pro/.local/lib/python3.8/site-packages/pikepdf/_qpdf-cpython-38.dll`. Could you check that ?
* Before trying to run the MSI installation you can also try to run `/mingw64/bin/python3 - m pdfarranger` from the MSYS2 command line
* Do you have the same issue with the .zip ?
* Do you have the same issue when running the MSI `pdfarranger.exe` file from the command line (and from its parent folder) ?
* I'm a bit puzzled by the *You may installed Microsoft Visual C++ 2015-2019* message. Mingw64 python3 and qpdf should not request that. May you have an official python installed (it's built with Visual C++) and it conflict with the one of Mingw64.

kbengs:
The pikepdf command did not create the dll because it already found some files there. So I renamed the site-packages to OLD and did a new try with pikepdf. That was successfull and installation worked after that. (I had already tried to build pdfarranger a few months ago and this probably had caused the problem)
Good to know about /mingw64/bin/python3 - m pdfarranger !
Thanks for help!

PR #187, opened 2020-04-19
Improve user experience

zsugabubus:
I try to address and fix some user experience issues. Commit messages should speak for themselves.

### DIFF
--- a/data/querysavedialog.ui
+++ b/data/querysavedialog.ui
@@ -1,0 +1,83 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Taken from LibreOffice. -->
+<interface>
+  <requires lib="gtk+" version="3.18"/>
+  <object class="GtkMessageDialog" id="querysavedialog">
+    <property name="can_focus">False</property>
+    <property name="title" translatable="yes" context="querysavedialog|QuerySaveDialog">Save changes?</property>
+    <property name="resizable">False</property>
+    <property name="modal">True</property>
+    <property name="type_hint">dialog</property>
+    <property name="skip_taskbar_hint">True</property>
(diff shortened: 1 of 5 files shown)


jeromerobert:
Thank you. Just some preliminary thinking:
* I'll take 308f652 as is
* In 2155417 I have to think about the `...`. It should be done everywhere and in all translation file in the same time. This was already requested by a translator.
* In b2c717f I wonder if we need a large `.ui` file for just a `GtkMessageDialog`.

zsugabubus:
> I'll take 308f652 as is

Thanks.

> In 2155417 I have to think about the .... It should be done everywhere and in all translation file in the same time. This was already requested by a translator.

I can update all if you would like to.

> In b2c717f I wonder if we need a large .ui file for just a GtkMessageDialog.

Not an expert, but as I looked, standard dialogs just have ‘Yes’, ‘No’, ‘Cancel’… etc buttons and I could not find any way to add custom ones. I think it would be nice if users would have a yes, I want to ‘Save’ it now and an explicit ‘DO NOT SAVE’. But it’s absolutely up to you. If you say we don’t need it, I can use a simple dialog too.

jeromerobert:
308f652 and 2155417 merged and will be in 1.5.2

zsugabubus:
Great to here.

Note that as part as of that quit confirmation dialog I rearranged the save logic and fixed a bug that can cause files marked as saved meanwhile save failed. The root cause of it is that files marked saved before they really got saved. [See here.](https://github.com/jeromerobert/pdfarranger/pull/187/files#diff-c540994ee693d2f56b5f47b884b94fdeL724) Probably you could fix that even if you do not take that commit.

Issue #182, opened 2020-04-16
Drag and drop doesn't work (Windows 10)


StefRe:
Drag and drop doesn't work (neither between file explorer and pdfarrange, nor between two instances of pdfarrange) - just nothing happens:
![drag_and_drop](https://user-images.githubusercontent.com/19879328/79476179-ed723500-8008-11ea-96d7-988b7a9e6a87.gif)
(the file is a valid pdf and can be imported by the menu as you can see in the second half of the video)

Copy and paste doen't work either (not sure if this is supposed to work from file explorer though):
![copy_and_paste](https://user-images.githubusercontent.com/19879328/79476259-0aa70380-8009-11ea-8598-b577de5e94fa.gif)
 (the file was indeed copied to the clipboard as you can see in the second half of the video where it get's pasted into file explorer).

pdfarranger version 1.5.0, Windows 10.

(oh, I forgot to say: great program with nice user interface! Keep it up!)

kbengs:
The first case (drag file from explorer to pdfarranger) I have looked at and got it working. I can make a PR for it later.

jeromerobert:
It was already not working in 1.4.2. It has probably never worked.

jeromerobert:
Fixed by c1662a3, thank you kbengs !

kbengs:
@jeromerobert, drag pages between two instances was also mentioned here. And that is still unsolved. :) (And add pdf file by right clicking on a pdf file in Explorer and paste to Pdfarranger..)

jeromerobert:
oops

dreua:
> And add pdf file by right clicking on a pdf file in Explorer and paste to Pdfarranger..

You mean by copy and paste from file explorer? That is not even supported on Linux. Would be cool though.

StefRe:
Thank you @kbengs! Drag and drop from file explorer was my main concern. Dragging between two instances of `pdfarranger` I only tried because I saw the three possible IDs [here](https://github.com/jeromerobert/pdfarranger/blob/c1662a39db2a9b70ab88035df4af6e05de2ea9f6/pdfarranger/pdfarranger.py#L216). And trying to insert a file from clipboard that was copied to clipboard in the file explorer was just out of curiosity.

kbengs:
If this line:
"os.environ['GDK_WIN32_USE_EXPERIMENTAL_OLE2_DND'] = 'any text'"
[...]

PR #191, opened 2020-04-25
Add paste pdf files feature & paste data error handling in Windows

kbengs:
Had to try to implement the feature mentioned in #182

Second commit should make it impossible to "manipulate" clipboard data with illegal data and cause errors in Pdfarranger in Windows. (copy pages in Pdfarranger -> paste to text editor -> change for example angle values to 45 degrees -> paste that data back to pdfarranger)

Tested with Dolphin, Nautilus and Thunar in Linux and File Explorer in Windows.

In Windows File Explorer a filepath can be copied with shift + right-click -> copy as path. Other methods to do it can be found here:

https://superuser.com/questions/1126793/create-copy-path-keyboard-shortcut-windows-10

Maybe not the most important feature to have, but quite useful anyway as it makes it possible to add files directly at desired location, with all four paste modes. (also multiple files can be copied at once) You may decide if it is worth to include it.

### DIFF
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -83,3 +83,13 @@ def export(input_files, pages, file_out, mdata):
         for k, v in mdata.items():
             outmeta[k] = v
     pdf_output.save(file_out)
+
+def num_pages(filepath):
+    """Get number of pages for filepath."""
+    try:
+        pdf = pikepdf.Pdf.open(filepath)
+    except pikepdf._qpdf.PdfError:
+        return None
+    npages = len(pdf.pages)
+    pdf.close()
+    return npages
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -211,7 +211,7 @@ def warn_dialog(func):
     return wrapper
 
 
-def get_file_path_from_dnd_dropped_uri(uri):
+def get_file_path_from_uri(uri):
     """Extracts the path from an uri"""
     uri = uri[5:]  # remove 'file:'
     path = url2pathname(uri)  # escape special chars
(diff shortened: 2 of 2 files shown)


jeromerobert:
Again an impressive amount of work ! I'll need some time to merge all that ;-). Some comments:
* I would prefere to keep pikepdf stuff in a dedicated file. Moreover `pdfarranger.py` is already bloated. To do so `export.pdf` could host a `num_pages(filename)` function.
* Could we have names for paste modes ? May be something like `PasteMode=enum.Enum('PasteMode', 'AFTER BEFORE ODD EVEN')`
* I'm not 100% sure, but it looks like if someone do an interleaved past with a 100 .jpg files (for example), he'll get 100 dialog box saying that the format is not valid. May be you should just stop at the first error.

kbengs:
Yeah, I have almost had to stop sleeping to have time to progam. :-) But I think it will slow down to a more slow pace soon.

> I'm not 100% sure, but it looks like if someone do an interleaved past with a 100 .jpg files (for example), he'll get 100 dialog box saying that the format is not valid. May be you should just stop at the first error.

Yes in Windows that is true. The get exactly same behavior on all paste modes I think it would be best if we don't paste anything if there is some fault. We will also have 100 dialog boxes with "PDF document is damaged" (from PageAdder class) if we paste after/before and actually with all other ways to add files to iconview also. So that is something that should be fixed too.

Maybe best if I add the fixes in new commits before you merge anything.

kbengs:
You may choose if you like enums or dictionary for paste modes...

jeromerobert:
Could you remove the history and keep only the final changes. I have trouble to review so many changes. I though I could just squash all commit together, but we have two features here.

Could you also be more precise in the 0c7a5fc commit message. I'm not sure to understand what it exactly do (#182 contains a lot of stuff).

kbengs:
Okay, here everything is in two commits. I had to do small changes so it works with images. I removed the code that would stop paste on first "PDF document is damaged", which I had included before. (if user tries to paste 100 damaged pdf files he will have 100 dialog boxes saying "PDF document is damaged")

jeromerobert:
Almost nothing to report. I would just do
```patch
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -964,7 +964,7 @@ class PdfArranger(Gtk.Application):
                         filepaths.append('\n'.join([filepath, str(1)]))
                     else:
                         num_pages = exporter.num_pages(filepath)
-                        if num_pages == None:
+                        if num_pages is None:
                             message = _('PDF document is damaged: ') + filepath
                             print(message, file=sys.stderr)
                             self.error_message_dialog(message)
@@ -992,8 +992,6 @@ class PdfArranger(Gtk.Application):
         elif os.name == 'nt' and data.startswith('pdfarranger-clipboard\n'):
             data = data.replace('pdfarranger-clipboard\n', '', 1)
             data = data.split('\n;\n')
-            if not self.is_data_valid(data):
-                data = []
         else:
             data_is_filepaths = True
             if os.name == 'posix' and data.startswith('x-special/nautilus-clipboard\ncopy'):
```
to silent LGTM and to make the first commit coherent (i.e. not use function from the second commit). I let you remove the *is not a supported format* dialog and I'll merge.

kbengs:
Okay now it's fixed.

Issue #101, opened 2019-08-09
Filetype: "application/wps-office.pdf" not supported! (Ubuntu)


ftharyanto:
![image](https://user-images.githubusercontent.com/17604130/62773848-379c3180-bade-11e9-93ef-71822f491b07.png)

OS: Ubuntu 19.10
PDF Arranger version 1.3.0

When I tried to open pdf, it failed and showing error message like in the picture. After I uninstall WPS, the PDF Arranger runs normally and could do the task. Is this bug could be fixed?

jeromerobert:
Could you give the output of `mimetype -D -a anypdffile.pdf` when WPS is installed?

jeromerobert:
Nevermind. I think it's useless to test the MIME type before opening a file. *pdfarranger* should try to always open the file (with Poppler). If this is not a PDF file it will raise a [`GLib.Error`](https://lazka.github.io/pgi-docs/#GLib-2.0/classes/Error.html#GLib.Error) with `domain=poppler-quark`, `code=4` and `message=PDF document is damaged`. I should just replace the message by `File is not a PDF document or is damaged`.

mimi89999:
Actually, is it OK to just change the MIME map like that? It doesn't seem so to me. Other people are also having this issue: https://askubuntu.com/questions/1171311/chrome-doest-recognize-pdf-files

It also isn't listed on IANA: https://www.iana.org/assignments/media-types/media-types.xhtml

I think that this is actually an issue with WPS and pdfarranger should continue to check the MIME.

jeromerobert:
Ok the bug was in WPS Office not pdfarranger. Let's reuse `mimetypes`.

PR #202, opened 2020-05-15
Add support for image files

mimi89999:
TODO:
- [x] Identify supported fileformat
- [x] Add setup changes
- [x] The pot file should be regenerated in a separate commit because there are unrelated changes
- [x] Don't keep 2 temp files

@jeromerobert I don't quite understand what you mean by _optional (in [...] and in the setup.py)_.

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -20,6 +20,7 @@ import sys  # for processing of command line args
 import threading
 import tempfile
 import signal
+import mimetypes
 import pathlib
 import platform
 import configparser
@@ -29,6 +30,13 @@ import locale  # for multilanguage support
 import gettext
 import gc
 from urllib.request import url2pathname
+
+try:
+    import img2pdf
+    img2pdf.Image.init()
+    img2pdf_supported_img = [i for i in img2pdf.Image.MIME.values() if i.split('/')[0] == 'image']
+except ImportError:
+    img2pdf = None
 
 sharedir = os.path.join(sys.prefix, 'share')
 basedir = '.'
@@ -751,6 +759,15 @@ class PdfArranger(Gtk.Application):
         filter_all.add_pattern('*')
         chooser.add_filter(filter_all)
 
+        if img2pdf:
+            filter_image = Gtk.FileFilter()
+            filter_image.set_name(_('Supported image files'))
+            for mime in img2pdf_supported_img:
+                filter_image.add_mime_type(mime)
+                for extension in mimetypes.guess_all_extensions(mime):
+                    filter_image.add_pattern('*' + extension)
+            chooser.add_filter(filter_image)
+
         filter_pdf = Gtk.FileFilter()
         filter_pdf.set_name(_('PDF files'))
         filter_pdf.add_pattern('*.pdf')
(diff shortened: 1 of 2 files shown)


dreua:
Hey, great to see someone working on this!

Related issue: https://github.com/jeromerobert/pdfarranger/issues/45#issuecomment-454972813

jeromerobert will probably speak for himself soon, but I think what he means and what I also think of this topic is that pdfarranger should continue to build and work without the additional dependencies needed for image support. (`setup.py` builds pdfarranger. Most people don't run it directly but through `pip install`.)

A few more of my thoughts to this topic:
- openjpeg is already a transitive dependency through poppler. Not sure if we can use it from python and/or if it helps with this task in any way though. 
- In general, if we can make it work without or with commonly available dependencies that would be appreciated. img2pdf seems to have a good coverage in the distributions with official pdfarranger packages ([repology](https://repology.org/badge/versions-matrix.svg?projects=pdfarranger,python:pikepdf%3E=1.7,img2pdf,python:img2pdf&header=pdfarranger%20packaging%20status) - note that python:img2pdf and img2pdf are the same).
- Bigger libraries often have a configure step where you tell the build system which build options you want to enable/disable. I am not sure if we can easily implement this with our current system or if it is needed at all. Alternatively pdfarranger can just check the availability of the needed dependencies at runtime and enable/disable functionality based on that.
- Keep user experience in mind, feel free to discuss it here before implementation. For example: What should happen when image dependencies are not available? Are there grayed out menu entries or are they not visible at all? Should the user be informed of this task? If yes, how? Pro: They know why the feature is missing. Con: Distracting. Non actionable for many users. Might make them unhappy. (I don't have an opinion on these questions yet, might also depend on other details of the implementation.)


Ping me for German translations when needed.

dreua:
*Note to self: Have a thorough look at the code before writing too much :see_no_evil: *

It looks like you already have working code and have thought about most of the stuff I mentioned above. I was mislead by the Draft status and all the unchecked boxes, sorry about my redundant comments. I need to sleep now, will try to test your code and be more concentrated tomorrow.

jeromerobert:
* By optional I mean that it must work without img2pdf which is what you almost did. We are just missing a `img2pdf=None` in the except to avoid `img2pdf not defined`.
* It should also be optional in `setup.py`. See https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) to support `pip install pdfarranger[img2pdf]`.
* `PDFDoc` is already creating a temporary file. I think we don't need to have 2 temporary files (especially when the first one is not deleted). May be you could add an other PDFDoc constructor that get a buffer as input ?

Just some remark
* I removed `mimetypes` usage because of #101 but I don't mind readding it for images
* Having img2pdf in the Windows installer will require some work and I don't plan to do it. img2pdf will be for Linux user's only until someone do it.
* I think people will want a way to specify the DPI of the image (but let's keep that for an other issue/PR)

mimi89999:
>     * By optional I mean that it must work without img2pdf which is what you almost did. We are just missing a `img2pdf=None` in the except to avoid `img2pdf not defined`.
Sorry, I missed that. Thanks for pointing that out.
>     * It should also be optional in `setup.py`. See https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) to support `pip install pdfarranger[img2pdf]`.
Done
>     * `PDFDoc` is already creating a temporary file. I think we don't need to have 2 temporary files (especially when the first one is not deleted). May be you could add an other PDFDoc constructor that get a buffer as input ?
On my TODO

I also made the MIME handling way more robust, but the code isn't as nice now.

mimi89999:
It should be ready now. @jeromerobert could you please review?

jeromerobert:
* `setup.py` is missing the img2pdf dependency
* Please don't use MIME for pdf type (see  #101). Your previous version was better (check mime for image, else just try to open the PDF).

jeromerobert:
You checked *Add translations (can do fr and pl)* but did not added those translation. If this is because you are concerned by the POT file synchronisation, don't bother. You can just regenerate the .pot file for you and commit only the line of the `fr.po` and `pl.po` files you modified.

mimi89999:
>     * `setup.py` is missing the img2pdf dependency
Added
>     * Please don't use MIME for pdf type (see  #101). Your previous version was better (check mime for image, else just try to open the PDF).
I think that it's an issue with WPS Office and that MIME map shouldn't be changed.

> You checked Add translations (can do fr and pl) but did not added those translation.

Removed that task.

jeromerobert:
Ok I agree that we should keep using `mimetypes`.

Now I get `NameError: name 'img2pdf_supported_img' is not defined` when trying your branch. You are missing a `img2pdf_supported_img=[]` right ?

mimi89999:
L762-768 should be wrapped in `if img2pdf:`. Missed that. Sorry

mimi89999:
I will push the change this evening. Is it OK besides that?

jeromerobert:
> Is it OK besides that?

I think so.

mimi89999:
Done

PR #203, opened 2020-05-16
Update pdfarranger.pot

mimi89999:
It needed to be regenerated and https://github.com/jeromerobert/pdfarranger/pull/202 got the change, but it should be separate.


mimi89999:
@jeromerobert Could you please review this, so I can move it from #202 and continue work on it?

jeromerobert:
Not needed. I only update POT when adding new translation or release. You don't need to change the POT file in your PRs.

PR #209, opened 2020-05-24
Set cell size

kbengs:
Setting iconview cell height has a negative effect though: The thumbnails that has smaller height than the others might be a bit blurred. This effect has also existed before, but only if the thumbnail is on same row as another thumbnail that has more height (in picture below on left window, top-right page)

Old behavior in left window, new to the right:
Set cell height OLD-NEW

So you may decide if this does more good than bad..

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -606,12 +606,16 @@ class PdfArranger(Gtk.Application):
             # scale*page_width*(1-crop_left-crop_right)
             item_width = int(max(0.5 + int(row[4] * row[11]) * (1. - row[7] - row[8])
                                  for row in self.model))
+            # scale*page_height*(1-crop_top-crop_bottom)
+            item_height = int(max(0.5 + int(row[4] * row[12]) * (1. - row[9] - row[10])
+                                 for row in self.model))
             item_padding = self.iconview.get_item_padding()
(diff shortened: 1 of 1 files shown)


jeromerobert:
Sorry but I do prefer the current layout, as it is more space efficient. Also, I cannot reproduce the blurred effect you are talking about.

kbengs:
Okay. Did you test the first commit? (set fixed cell width) It has only positive effects. 
If you try to drag more pages to iconview, or try to zoom in or out, you will sometimes see that the items are first too narrow or wide (which causes that the cells move too much to the left or right) and when page rendering is finished they move back to correct place.
The commit removes this unnecessary extra left-right shifting by placing the items directly where they should be.

jeromerobert:
> If you try to drag more pages to iconview, or try to zoom in or out, you will sometimes see that the items are first too narrow or wide

I've not been able to see that but as this commit does not hurt, I merged it.

kbengs:
Thanks. I can make a video for you later so you can see the positive effect. :-)

kbengs:
On this video the unwanted effect can be seen when cell width and height changes which causes the icons to temporary move left-right and up-down. The effect is eliminated when we set fixed sizes.
[Uploading Fixed width not enabled.zip…]()

Issue #106, opened 2019-09-08
Idea: Insert pages of a PDF at drop location


dreua:
Currently when dropping a pdf file on pdfarranger, all pages are inserted at the end. I think the user experience might be better with the pages inserted at the drop location. This is already possible when moving a selection of pages from one pdfarranger instance to another, so I believe it shouldn't be to hard to do it for pdf files as well.
I will wait for feedback before I might start working on this.

jeromerobert:
In that case, the `Gtk.IconView` does not show a cursor where the drop should happen but it does not matter.  `iconview.get_dest_item_at_pos(x, y)` does work so I agree that it should be easy to implement. It's about of mimic `iv_dnd_received_data` in `sw_dnd_received_data`.

dreua:
> In that case, the `Gtk.IconView` does not show a cursor

Do you know why that is the case? Is is just not possible if the drag event comes from another window or is it somewhere in the code? 
I would also love to have a better user experience with drag and drop in general, because quite often the drop location is one off the point that I was expecting. I'm thinking about moving the existing pages slightly out of the way (maybe animated) to signal the current drop location. I have no idea though, whether this is possible or how much work it would be, need some research on that.

jeromerobert:
> Do you know why that is the case?

Sadly, I don't know. But as a starting point I would look for `GTK_IS_TREE_DRAG_SOURCE` in https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-24/gtk/gtkiconview.c. May be also:

<https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/TreeDragDest.html#Gtk.TreeDragDest.row_drop_possible>

is involved (https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-24/gtk/gtktreednd.c).

> Is is just not possible if the drag event comes from another window or is it somewhere in the code?

It's the default GTK behavior when the source of the drag is not a Gtk tree (whatever it is in the current process or not). I don't know if that can be changed. Comments in <https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-24/gtk/gtktreednd.c> tend to say it is.

jeromerobert:
Done by kbengs at c807a86 and d420186.

PR #211, opened 2020-06-02
Changes in drag location indicator behaviour & solution for #106 & #182

kbengs:
**Change drag & drop location look and behaviour:**
I think drop above, under and into are a bit confusing. We are only dropping to left or right so best if we show only thouse drag locations. Drop reject between items is also not an ideal solution. Better if we dont have any "no mans land" between items.

**Add support for insert file at drop pos (Linux):**
Previous commit made drag location indicator visible also when drag file to app. So now we kind of have to implement #106

**Use ole2 dnd in Windows:**
Remaining thing in #182.

**Hold drag icon in the middle:**
When we start a drag we probably grab the page somewhere in the middle. So I think it feels more intuitive that we also hold the icon in the middle.

### DIFF
--- a/Win32.md
+++ b/Win32.md
@@ -28,7 +28,7 @@ and
 
 ```
 pip install --user https://launchpad.net/python-distutils-extra/trunk/2.39/+download/python-distutils-extra-2.39.tar.gz
-/mingw64/bin/python3 -m pip install --user img2pdf pikepdf https://github.com/jeromerobert/cx_Freeze/zipball/pdfarranger
+/mingw64/bin/python3 -m pip install --user keyboard img2pdf pikepdf https://github.com/jeromerobert/cx_Freeze/zipball/pdfarranger
 ```
 
 Get the pdfarranger sources from a MSYS2 shell:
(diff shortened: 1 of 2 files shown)


jeromerobert:
Again an impressive piece of work, thank you. Reading the commits it looks good. I'll just do some basic tests and merge it.

kbengs:
I found a way to do move-drag to other instance in Windows. I add the change soon. I rebase -i the "Use ole2 dnd in windows" commit. And I can fix the lgtm alert.

kbengs:
Its fixed. Changes I made are on lines 82, 1232, 1246 and 1251

kbengs:
Some more improvements, sorry :-)
* Position in center of cell where it decides if cursor should be on left or right side wasn't exactly in the middle. For some reason it was dependent of margin width. Also made this position so it has a bit of hysteresis.
* It was possible to sneak in with a dragged page between both rows and columns without drag-cursor showing up.

Do you want me to rebase -i so the history is clean?

kbengs:
Some more improvements again in first commit. Now it really should be ready for review. :-)

Issue #216, opened 2020-06-26
1.6.0 almost ready, please test master


jeromerobert:
We now have a lot of (and enough ;-) ) new features in master that needs to be tested. New developments will wait for 1.7.0.

I plan to release 1.6.0 by mid-July if no annoying bugs are found by then.

jeromerobert:
https://github.com/jeromerobert/pdfarranger/milestone/10?closed=1

jeromerobert:
I usually prepare the release notes in the draft of the release but this has the drawback of not being publicly visible. This time I will write (and update) them here:

* Improve iconview look and feel (#201, #209)
* Support image file import (#45)
* Ask user confirmation before quit (#187)
* New application icon (#205)
* Insert pages of a PDF at drop location (#106)
* Support page drag and drop on Windows (#182)
* Support Copy/Paste from a file explorer (#182, #191)
* Recommend pikepdf >= 1.15.1 (#198)

dreua:
Wow, we can now drop PDFs directly to the correct location and even paste a PDF file from the file explorer (nautilus in my case)! Great work @kbengs !

kbengs:
One thing I noticed is that it is possible to save a pdf before any pages are added to iconview. It results in a pdf with 0 pages.

dreua:
Haha, I wouldn't have thought its possible but here we go, a PDF with all the metadata but no actual pages :laughing: 
![Screenshot from 2020-06-30 10-47-38](https://user-images.githubusercontent.com/15114715/86105101-36417200-babf-11ea-84e1-3901ccc0f67a.png)

jeromerobert:
What about saying it's a feature ? (and not a bug)

kbengs:
Maybe it could be seen as a feature. Some errors are printed to terminal though if we try to open the file later in pdfarranger.

If you like to I could test the windows release, if I could have it somehow. Would it be any idea to release that as a beta?

jeromerobert:
> Some errors are printed to terminal though if we try to open the file later in pdfarranger.

Hum ok. Then I disabled it (8ae51f0).

> If you like to I could test the windows release, if I could have it somehow. 

Done: https://github.com/jeromerobert/pdfarranger/releases/tag/1.6.0rc1

kbengs:
A minor issue:
![save_text](https://user-images.githubusercontent.com/60842129/86525458-f5b37100-be8f-11ea-88fb-0ba3fff3614e.png)

[...]

PR #219, opened 2020-07-12
hu locale update

ovari:


jeromerobert:
Merged as fd6e73e

ovari:
Thank you for promptly accepting hu locale updates. Can you please advise when [Flathub](https://flathub.org/apps/details/com.github.jeromerobert.pdfarranger) will have the update? Thank you once again.

dreua:
Szia ovari, köszönöm szépen! - Thank you for the translation! (Unfortunately that is where my Hungarian skills come to an end.)

Since I take care of the Flatpaks I'll answer this: The flatpaks on Flathub contain only the released versions. I believe that Jerome will include this change in 1.6.0 which should be released in the near future (days to weeks), see  #216 for more on that. Flathub will be updated as soon as I find the time, usually between hours to a few days after release. Feel free to watch the [flatpak repository](https://github.com/flathub/com.github.jeromerobert.pdfarranger) to get notified of updates and/or ping me if I should ever forget one.

I can also make a test build from current master if that helps you. (Downsides: They seem to be removed after less than 20 days and probably don't update automatically, so for testing only. - See [this pr](https://github.com/flathub/com.github.jeromerobert.pdfarranger/pull/5) for example, which is a 404 now.)

While I can't comment on the actual translation, I tested and it looks all translated and good to me, just fixed three underscores showing up.

jeromerobert:
> I believe that Jerome will include this change in 1.6.0 which should be released in the near future (days to weeks)

I'll release 1.6.0 tomorrow.

ovari:
Does `locale hu (Hungarian - magyar) update (#219)` get added to the release notes at https://github.com/jeromerobert/pdfarranger/issues/216#issuecomment-650080492? Thank you

PR #221, opened 2020-07-14
Updated italian translation

MichelePezza:


po/it.po
```diff
@@ -1,297 +1,314 @@
...
+msgstr "Annulla"
+
+#: pdfarranger/pdfarranger.py:670 data/menu.ui.h:2
+msgid "_Save"
+msgstr "Salva"
+
+#: pdfarranger/pdfarranger.py:710
 msgid "Export…"
-msgstr "Esporta…"
+msgstr "Esporta..."
```
dreua, line 79: I think all of these ... were changed to the single character on purpose, any reason you changed it back?

po/it.po
```diff
@@ -1,297 +1,314 @@
...
-
 #: data/menu.ui.h:3
 msgid "Save _As…"
-msgstr "Salva Come…"
+msgstr "Save come…"
 
 #: data/menu.ui.h:4
 msgid "E_xport Selection…"
-msgstr "E_sporta Selezione…"
+msgstr "Esporta Selezione…"
```
dreua, line 222: Why are you removing the accelerators? They are useful for keyboard navigation imo. Press/hold the Alt key in menus to see them in action. 

This is just the first occurrence, there are more changes like this below.

jeromerobert:
I'll release tomorrow 1.6.0. If Michele does not fix its PR in the meantime I'll just pick the additional translations from its commit and ignore other changes.

po/it.po
```diff
@@ -1,297 +1,314 @@
...
+msgstr "_Taglia"
 
 #: data/menu.ui.h:16
 msgid "_Copy"
 msgstr "_Copia"
 
 #: data/menu.ui.h:17
 msgid "Paste _After"
-msgstr "Incolla_Dopo"
+msgstr "Incolla prima"
```
dreua: Did you switch prima/dopo on purpose? Just using Google translate, I think after = dopo; before = prima was correct, but I trust you if you tell me it was done on purpose.

dreua:
Thanks for the quick fix!

Just tested it and looks all translated and good to me, like in Hungarian I can't comment on the actual translations except for the prima/dopo change that is looking suspicious in the diff. 

### General issue: Accelerator duplication

Duplicated accelerators can happen quite easily and you need to double and triple check in the app to make sure you got them all. I did this for German and needed a few iterations until I got it right. Italian has some, e.g. d and p in the "Modifica" sub-menu. For Hungarian I didn't check. I'm not sure how much priority we should give this tbh as the amount of user noticing that will probably be rather small. To get it right, we would need some automation (actually I believe Gtk would need some automation first) to report these duplicates. In the meantime you can still use the app by keyboard, but you need multiple key presses to iterate through the items with the same accelerator.

po/it.po
dreua: This is not "outdated" it's just more obvious in the first commit.

jeromerobert:
Merged: 68f9ae8

jeromerobert:
@dreua I was also concern by the duplicate accelerator in the french translation. I removed all of them. Then I received 8d4a028 from an experienced gnome translator. By email he says that it's not an issue to have duplicated accelerators. In that case you just have to press the accelerator key multiple time.

I guess that the logic is to have accelerators has coherent as possible between applications and that this is more important than avoiding duplicate accelerators. I almost never use accelerators in menus so I don't know what other applications do but I think we should do the same. We should ask other gtk/gnome/translator veteran what is the common practice.

dreua:
> have accelerators has coherent as possible between applications

Huh I haven't thought that would be a goal, but yeah, might make sense if you actually use them regularly. If you could get some input from that veterans that would be great, placed into po/README.md this will be shown by Github when viewing the po folder. (BTW I just called them accelerators without research, not sure if mnemonics would be the better term.)

afranke:
Veteran translator here (started 16 years ago, worked for a long time in Ubuntu and GNOME, currently coordinator for the French GNOME translation team as well as internationalisation coordinator for GNOME).

We do have guidelines for mnemonics and they may have slight differences for some languages, but the gist of it is, ordered by priority:
* be consistent with other GNOME software
* pick a unique letter **within that given menu** if possible
* pick the same letter as the original string if available
* pick a strong letter (e.g. in "Search and replace" I would rather pick `s`, `r` or `p` than `a`)

Given all of that, it should be **pretty** rare that duplicates can’t be avoided and consistency should *not* prevent distinct mnemonics as the existing ones have already been chosen with uniqueness in mind.

I’m happy to help more if needed.


P.S.: do note that with the current trends in GNOME app design, menus are becoming less common and we have a lot less of them to translate.

PR #229, opened 2020-07-19
Added some info about translations in README.md

milotype:


README.md
```diff
@@ -62,3 +62,11 @@ python3 -m pdfarranger
 ```
 
 For Windows see [Win32.md](Win32.md).
+
+## Translations
+
+*pdfarranger* has been translated into several languages already:
+
+ca@valencia, cs, de, el, es, fr, hr, hu, id, it, ja, nl, pl_PL, pt_BR, tr, zh_CN, zh_TW.po
```
dreua, line 70: Extra .po: zh_TW.*po*

milotype: Thank you for noticing ... I've corrected it ...

dreua:
I'm not exactly sure if we need this information in Readme, but I'm also not against it, so that is up to Jérôme to decide anyway.

I'm missing the information on *how* to translate, I think this could go in a `po/README.md` (to be created) and linked here to not overload the the main Readme and maximize visibility for potential translators. Some information that it should contain (untested from the top of my head):
- Install from a local git checkout with your translation: `pip3 install --user .` (The dot at the end is important, run in the checked out root directory. I'd recommend to uninstall other pdfarranger packages beforehand but keep the dependencies.)
- Previous command must be re-run after each source change to take effect.
- Test different languages: `LANG="xx_YY" pdfarranger` (You may need to add `~/.local/bin/` to you PATH or just run it directly from there.)
- You may want to uninstall pdfarranger after testing and re-install your distro's package for automatic updates: `pip3 uninstall pdfarranger`
- Displayed name and description can be translated in the desktop file.
- Some explanation about the underscores and how to use them (TBD). Only use them in strings where the english version also contains an underscore.

dreua:
- When adding new translations, please also update the main README.md

milotype:
Before complicating it too much, you may consider implementing a translation platform like [https://weblate.org/](url). (I have no connection to Weblate, I simply translate diverse project on the platform.) There are other platforms, but I like this one.

Weblate offers free hosted or self-hosted plans for open source projects and it's a great way to reach a much broader community. As it has a good documentation for translators, there would be no need for documenting "how to translate" here at pdfarranger.

Weblate even offers widgets for translation statuses (for example: https://hosted.weblate.org/widgets/pdf-mix-tool/?lang=&component=translations), which can be added everywhere – in the `README.md` file, or maybe even at https://flathub.org/apps/details/com.github.jeromerobert.pdfarranger.

jeromerobert:
I've already tried Transifex and it was a nightmare to manage. It's like having 2 git repositories that you have to synchronize manually. I don't expect Weblate to be better.

I don't see any reasons to add `ls po/` to `README.md`, it's an information that only developers or translators will understand and they already know where to find it. If this was for end users we would need full language name (Français, Deutsch, hrvatskom, ...) and that would be too much work for little added value.

I agree that we need a "how to contribute translation" README (or add a *For translators* section after *For developers* in the main README). I'll write it down from the ideas @dreua listed.

milotype:
@jeromerobert I'm sorry, I can't help with the technical stuff, as I'm not a programmer. But maybe you can spare some time to check Weblate. It's definitely better for translators and the mentioned translation widgets update automatically! And you wouldn't have to specify "how to contribute translation" in the README file.

As a matter of fact, **the main idea is actually to inform end users** – not translators or developers! But end users may be translators or developers as well. Either way, it would help _pdfarranger_ …

My problem is, that nowadays many websites are/may be machine translated into user's locale. But a machine translated page does not imply, that the application itself is localized …

Issue #230, opened 2020-07-19
Add a "how to translate" README


jeromerobert:
I'm missing the information on *how* to translate, I think this could go in a `po/README.md` (to be created) and linked here to not overload the the main Readme and maximize visibility for potential translators. Some information that it should contain (untested from the top of my head):
- Install from a local git checkout with your translation: `pip3 install --user .` (The dot at the end is important, run in the checked out root directory. I'd recommend to uninstall other pdfarranger packages beforehand but keep the dependencies.)
- Previous command must be re-run after each source change to take effect.
- Test different languages: `LANG="xx_YY" pdfarranger` (You may need to add `~/.local/bin/` to you PATH or just run it directly from there.)
- You may want to uninstall pdfarranger after testing and re-install your distro's package for automatic updates: `pip3 uninstall pdfarranger`
- Displayed name and description can be translated in the desktop file.
- Some explanation about the underscores and how to use them (TBD). Only use them in strings where the english version also contains an underscore.

_Originally posted by @dreua in https://github.com/jeromerobert/pdfarranger/pull/229#issuecomment-660672434_

ovari:
Should https://github.com/jeromerobert/pdfarranger/blob/master/data/com.github.jeromerobert.pdfarranger.desktop also be added to the files that should contain translations?

ovari:
Can https://github.com/jeromerobert/pdfarranger/blob/master/data/com.github.jeromerobert.pdfarranger.metainfo.xml be translated too?

dreua:
### Desktop file
Yes, I already included that but in a very short form: "Displayed name and description can be translated in the desktop file." This should obviously be expanded and include a link to the actual file.

### Metainfo
Good catch, yes that can also be translated. Here is a random example from my harddrive: 
[network-manager-pptp.metainfo.xml.txt](https://github.com/jeromerobert/pdfarranger/files/4958802/network-manager-pptp.metainfo.xml.txt) (Yikes, Github does not allow .xml files to be attached.)

This is what it looks like. Note that most lang identifiers are two letter codes, but there is also `fur`, `pt_BR` and `sr@latin`. Research/Testing is needed to make sure this is compatible with the codes we use at the moment.
```xml
    <p>Compatible with various PPTP servers including Microsoft.</p>
[...]
    <p xml:lang="fur">Compatibil cun varis servidôrs PPTP includût Microsoft.</p>
[...]

PR #231, opened 2020-07-22
 Remove out-dated lines from po

ovari:


jeromerobert:
Thank you @ovari I'll merge it.

But the problem with such PR is that I will spend more time to check it and reshape it, than to do it myself. If you don't know how to use git from command line please avoid such PR. With git you would have been able to do this in one commit and have a meaningful commit message. Moreover it was probably not very nice for you to write all that changes in github instead of your prefered text editor. Unfortunately learning git is often a too high investment for non developers.

jeromerobert:
Reminder for me: Do not forget my above blah when doing #230

PR #233, opened 2020-07-22
Updates for German and Croatian

milotype:
- updated/completed German and Croatian translations with newest .pot file
- some minor corrections in German
- changed the mention of "PDF-Shuffler" to "pdfarranger" in German
- remove out-dated lines in German

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -290,11 +290,11 @@ msgstr "Einfügen _vor Auswahl"
 
 #: data/menu.ui.h:19
 msgid "Paste Interleave _Odd"
-msgstr "Leere _ungerade Seiten einfügen"
+msgstr "Abwechselnd einfügen _ungerade"
 
 #: data/menu.ui.h:20
 msgid "Paste Interleave _Even"
-msgstr "Leere _gerade Seiten einfügen"
+msgstr "Abwechselnd einfügen _gerade"
 
 #: data/menu.ui.h:21
 msgid "Zoom _In"
--- a/po/de.po
+++ b/po/de.po
@@ -1,103 +1,139 @@
-# PDF-Shuffler translation files.
-# Copyright (C) 2008-2017 Konstantinos Poulios
-# This file is part of PDF-Shuffler which is released under the GNU General
-# Public License version 3 or later.
-# See file COPYING or go to <http://www.gnu.org/licenses/> for full license details.
+# German translation for pdfarranger.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the pdfarranger package.
 # Translators:
 # Konstantinos Poulios, 2008-2017
 # David Auer, 2020
+# Milo Ivir <mail@milotype.de>, 2020
 msgid ""
 msgstr ""
-"Project-Id-Version: PDF-Shuffler 0.7\n"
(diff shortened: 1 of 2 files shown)


po/de.po
```diff
@@ -128,7 +164,23 @@ msgstr ""
 "%s ist ein Programm zum Umsortieren und Modifizieren von PDF-Dokumenten. "
 "Geschrieben in GTK+ und Python."
 
-#: pdfarranger/pdfarranger.py:1449
+#: pdfarranger/pdfarranger.py:1631
+msgid "Maintainers and contributors"
+msgstr "Verwalter und Mitwirkende"
```
dreua, line 169: I would prefer Maintainer here. 
https://de.m.wikipedia.org/wiki/Maintainer

po/de.po
```diff
@@ -242,11 +290,11 @@ msgstr "Einfügen _vor Auswahl"
 
 #: data/menu.ui.h:19
 msgid "Paste Interleave _Odd"
-msgstr "Verzahnt einfügen _ungerade"
+msgstr "Leere _ungerade Seiten einfügen"
 
 #: data/menu.ui.h:20
 msgid "Paste Interleave _Even"
-msgstr "Verzahnt einfügen _gerade"
+msgstr "Leere _gerade Seiten einfügen"
```
dreua, line 297: Verzahnt is not great but with Leere I wouldn't have clue what this is doing.

po/de.po
milotype: I don't see a reason to use English here. But "Verwalter" isn't nice, I agree.
Why not Entwickler or Hauptentwickler?

po/de.po
milotype: As far as I understand it, it's an option to include even or odd empty pages.

po/de.po
dreua: Hauptentwickler is a good idea! In context, I'd say going with "Entwickler" (plural) only would actually be the nicest.

milotype: OK, I'll change it to Entwickler.

po/de.po
dreua: No, I'm not aware of a function to add empty pages at all. This is about pasting the pages from clipboard in every second place. This is useful for example if you scanned all the fronts of your pages first and than all the backs and you want to put it back in a PDF in a way where each backside comes after it's front side.

Pages in Clipboard: A B C D
Pages displayed: 1 2 3 4
After paste interleave odd: 1 A 2 B 3 C 4 D

milotype: Sorry, I misinterpreted it then. Thank you for explaining.
How about **Verschachtelt** instead of Verzahnt?

milotype: Two more alternatives:
**Alternierend** or **Abwechselnd**?

dreua: I think Verzahnt is better than Verschachtelt.
I'm okay with Alternierend, Abwechselnd or Verzahnt, not sure what is best.

milotype: In that case I'd propose **Abwechselnd**.
But if you want, you may decide ;-)

milotype:
@dreua 
I've made the necessary corrections … files could be merged now …

jeromerobert:
Merged: 85227fd. Thanks.

PR #234, opened 2020-07-24
Update README.md

ovari:
What are you using? en-US or en-GB? `synchronized` or `synchronised`? Thank you


jeromerobert:
> What are you using? en-US or en-GB?

I really don't care to even mix both of them. I hope that Shakespeare won't mind.

Merged 03b44dd

Issue #245, opened 2020-07-24
"Past after" & "Past special" options should be inactive if no "Cut" or "Copy" action already done.


Nokia808:
Hi.

I noticed that, currently, "Past after" & "Past special" options are always active in right click list even if user did not perform "Cut" or "Copy" option ! This is not correct ! These past options should only be active after user perform "Cut" or "Copy" operations. Otherwise, they should appearing in right click list as  inactive

jeromerobert:
f45bd7f19cdd3e5bace8786b655d2e4c8489a2a7

PR #250, opened 2020-07-26
Grey out paste actions if nothing in clipboard

kbengs:
A possible solution for #245 

This is still not perfect as it does not check if data in clipboard is something that pdfarranger supports (undamaged PDF file or supported image if img2pdf installed, or pdfarranger page data in Windows) But maybe good enough? I tested it also on Windows.


jeromerobert:
It looks good, thank you. I will wait before merging it because I think 1.6.1 will be a translation only release. #245 is not critical so it can wait for 1.7.0.

jeromerobert:
Just changed my mind ;-)

PR #260, opened 2020-08-13
Resolves #240

### LINKED ISSUE #240: Add "Export selection to individual files"
Hi. Currently the "Export selection" option working to export all selected page into single output PDF.

Can you, kindly, added option to "Export selection" into individual PDF files (each selected page should exported into independent PDF file) ?

angsch:
Add two more export options, 'export selection to individual files' requested in #240 and 'export all pages to individual files'.

### DIFF
--- a/data/menu.ui
+++ b/data/menu.ui
@@ -35,11 +35,27 @@ along with PDF-Arranger.  If not, see <http://www.gnu.org/licenses/>.
         <attribute name="label" translatable="yes">Save _As…</attribute>
         <attribute name="action">win.save-as</attribute>
         <attribute name="verb-icon">document-save-as-symbolic</attribute>
+        <attribute name="target" type="i">0</attribute>
       </item>
-      <item>
-        <attribute name="label" translatable="yes">E_xport Selection…</attribute>
-        <attribute name="action">win.export-selection</attribute>
-      </item>
+      <submenu>
+        <attribute name="label" translatable="yes">E_xport…</attribute>
+        <section>
+          <item>
+            <attribute name="label" translatable="yes">E_xport Selection to a Single File</attribute>
+            <attribute name="action">win.export-selection</attribute>
+            <attribute name="target" type="i">2</attribute>
+          </item>
+          <item>
+            <attribute name="label" translatable="yes">Export Selection to _Individual Files</attribute>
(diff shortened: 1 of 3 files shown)


Nokia808:
@angsch 
Thank you for this PR !
Only one note: with "export to individual files" options, the output files should be without the outlines (TOC) that may some time already existing in original (input) PDF. For example, if I have a PDF has outlines (TOC) for it's 17 chapters & their sub-chapters, it will be meaningless if each of extracted individual files (that each of them represent a single page of original PDF) has TOC of original file.

I'm nearly sure you are already addressed this, but since I'm not programmer I give this note to avoid post implementation fix ....

Best regards.

kbengs:
Some observations when I tested it:
* When starting this is printed to terminal: "SyntaxWarning: "is" with a literal. Did you mean "=="? if exportmode is 'ALL_TO_SINGLE':"
* Mnemonics accelerators are missing in menu.ui (Keyboard shortcuts, Alt + X) Could be for example "Export Specia_l…"
* "export selection" and "save as" are always greyed-out in hamburger menu (or am I missing something?)
* Export options are only available in right-click menu, shouldn't they be in hamburger menu also?

But overall I think it looks good.

angsch:
@kbengs 
Thanks for testing. It is really good that you point of the thing with the hamburger menu. I am gonna fix your points. Thanks, again.

jeromerobert:
Thank you @angsch. You PR looks good. Here is what I would add to kbengs comments:
* I think "Export selection" is already something "special". So I would prefer that you rename "Export Special..." to "Export..." and move "Export Selection" to it.
* I also think that the new export submenu should be in the hamburger menu.
* Could you also rebase your branch to master and remove the merge commit ?

angsch:
The code has been revised according to your comments. I think it is ready for being merged.

kbengs:
Two small things I still notice:
* You probably forgot to remove the line: "attribute name="target" type="i">0</attribute" from the "save as" section in menu.ui
because save as in hamburger menu is still always greyed-out
* Empty line in menu.ui line 320

kbengs:
CTRL+F is usually shortcut for find, maybe something else should be used here?

jeromerobert:
I merge it, we'll fix that kind of details.

jeromerobert:
56b3150

PR #263, opened 2020-08-16
Switch select odd and even actions

kbengs:
@jerome, could I have:
git push upstreamPdfarrangerTeam master FixSelectOptions
here? I mean is that the right command/way to push obvious fixes?

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -1127,13 +1127,13 @@ class PdfArranger(Gtk.Application):
         elif selectoption == 'DESELECT':
             self.iconview.unselect_all()
         elif selectoption == 'ODD':
-            for page_number, row in enumerate(model):
+            for page_number, row in enumerate(model, start=1):
                 if page_number % 2:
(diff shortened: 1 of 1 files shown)


jeromerobert:
Not exactly. The `<refspec>` argument format is `<local src>:<remote dst>`, so `git push upstreamPdfarrangerTeam FixSelectOptions:master`. if `<local src>:` is omitted it's supposed to be the same as `<remote dst>` so a common practice is to  put the commits you want to send to the team master into your own master branch and just do `git push team master`.

pdfarranger/pdfarranger.py
```diff
@@ -1127,13 +1127,13 @@ def on_action_select(self, _action, option, _unknown):
         elif selectoption == 'DESELECT':
             self.iconview.unselect_all()
         elif selectoption == 'ODD':
-            for page_number, row in enumerate(model):
+            for page_number, row in enumerate(model, start=1):
```
dreua, line 1130: Very pythonic imo, I like it :+1:

Issue #262, opened 2020-08-15
show the current selection in the top bar


jeromerobert:
I'm not sure users will understand what this action is for and remember to use it when they need it.
How about a page number display of the last selected page instead, like in evince? (Showing page x of y in the top left instead of just y.) If we want to make it extra fancy, it could even show selections of multiple pages ("1, 5, 99 of 100", "1-100 of 100" etc.) and double it as an input (insert 2-22) to select those pages... Well these are just some ideas, not sure if it is worth our time really. But just showing the last focused page (ignoring selections of multiple pages for now) could be helpful imo, what do you think?

_Originally posted by @dreua in https://github.com/jeromerobert/pdfarranger/issues/261#issuecomment-673599744_

We could show the current selection in the top bar (next too the total page number), but that's for an other issue.

_Originally posted by @jeromerobert in https://github.com/jeromerobert/pdfarranger/issues/261#issuecomment-674410273_

PR #268, opened 2020-08-24
Display current selection

angsch:
Addresses #262 

In my opinion there is very little space in the top bar, so I opted for adding a status bar where the current selection is displayed. Perhaps the status bar will turn out to be useful also for other things.

image

### DIFF
--- a/data/pdfarranger.ui
+++ b/data/pdfarranger.ui
@@ -306,6 +306,36 @@ along with PDF-Arranger.  If not, see <http://www.gnu.org/licenses/>.
             <property name="position">3</property>
           </packing>
         </child>
+        <child>
+          <object class="GtkSeparator">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+          </object>
+          <packing>
+            <property name="expand">False</property>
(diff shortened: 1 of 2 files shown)


kbengs:
The status bar looks good imo. It could be that this is a bit overkill... to show all the ranges. But I don't know. To save same space in top bar rotate left & right could be removed imo.

Yet another possibility would be to show "absolute" page numbers in the description under the thumbnails. The word "page" could be removed and instead it could show for example: "original page number" (absolute page number) Any comments on that?

kbengs:
The status bar looks good imo. It could be that this is a bit overkill... to show all the ranges. But I don't know. To save same space in top bar rotate left & right could be removed imo.

Yet another possibility would be to show "absolute" page numbers in the description under the thumbnails. The word "page" could be removed and instead it could show for example: "original page number" (absolute page number) Any comments on that?

jeromerobert:
I agree with kbengs, a status bar takes a lot of place. But that PR code looks perfect and the feature is implemented, so I merged it as 03534ef7980fac07a5bf1ae5946fc5ba53a6684d. Time will tell.

Issue #51, opened 2019-01-30
Option to zoom to full page


madduck:
Upon starting the programme, it would be nice if I had the option to make it zoom so as to display as much as possible of the first page, not just thumbnails.

jeromerobert:
I won't work on that issue. I think that now that the zoom level is saved, it's not that useful. Anyway I leave the bug open in case that someone want to do it.

Nokia808:
@jeromerobert 
Hi.
This issue is related to this closed issue https://github.com/jeromerobert/pdfarranger/issues/235
I will post a comment about this in issue https://github.com/jeromerobert/pdfarranger/issues/235 kindly, to see it.

kbengs:
Evince use key f to zoom to full page, maybe we could use it also..

kbengs:
I have now made something that zooms into full page and scrolls to the page when double-clicking on a page.
When double-clicking again it zooms back to set zoom level. 
Key f zooms to the page where cursor is. Pressing key f again zooms back to set zoom level. 

Does anybody argue against this behavior?

jeromerobert:
Done in 613350f8b9027ab8963e23a5dd40f8bd25ac0b86. Thank you @kbengs .

PR #269, opened 2020-08-26
Scroll the view with up/down & page up/down keys

kbengs:
* Scroll a small step up or down with arrow keys. 
* Scroll to previous/next page row with page up/down keys
* Disable the iconview built in "move cursor" function

I wasn't sure if it should scroll just a bit or a whole page on keyboard arrows. What do you others think about this?

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -479,6 +479,7 @@ class PdfArranger(Gtk.Application):
         self.iconview.connect('motion_notify_event', self.iv_motion)
         self.iconview.connect('button_release_event', self.iv_button_release_event)
         self.iconview.connect('selection_changed', self.iv_selection_changed_event)
+        self.iconview.connect('key_press_event', self.iv_key_press_event)
 
         self.sw.add_with_viewport(self.iconview)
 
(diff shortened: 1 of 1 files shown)


dreua:
I personally would like to keep the possibility to move the cursor by keyboard, it works well if you just want to select a single page. I'm quite sure I used it a few times on the laptop because I don't like touch pads. I'm open for discussion though, maybe I'm missing a point? Maybe Shift+Arrows could be used for scrolling as an alternative.


(OT: If we hadn't bound page rotation to Ctrl+Left/Right one could even select multiple pages by Ctrl+Arrows+Space. (You can try this now, but only with Up/Down, so only in one column which is not really useful. Should we switch to Alt+Left/Right for rotation to fix this?)

kbengs:
My (maybe not so good) arguments for removing it was that it isn't working perfectly now. (because it doesn't scroll if cursor goes of screen) And therefore I guess that not many use the feature. This commit would give a possibility to scroll with keyboard. 

Stepping one page at a time up/down would make it work better as a pdf viewer. (as next step I was planning to try to find a way to do #51)

Would it be best if:
* The built in move cursor is not removed. Instead it would start to scroll when cursor otherwise would go of screen? (as for example nautilus works)
* Page up/down would step one page row at a time, as it already does in this commit?

Shift + arrows already works for selecting multiple pages.

dreua:
Yes, that would be ideal imo. I usually have small PDFs which fit in one screen, therefore I wasn't aware of this problem.

> Shift + arrows already works for selecting multiple pages.

You're right, my bad.

kbengs:
OK, a new try.

jeromerobert (approved):
Approving.

PR #272, opened 2020-09-04
GtkListStore refactoring

jeromerobert:
The current core data structure of pdfarranger is currently a GtkListStore. It's difficult to read because of the number
of page attributes. With this PR GtkListStore (PdfArranger.model) now only have 2 columns: one for the descriptions, and one with the Page object which contains all other attributes.

This should help adding and manage new attributes to pages.

### DIFF
--- a/pdfarranger/croputils.py
+++ b/pdfarranger/croputils.py
@@ -36,9 +36,8 @@ def dialog(model, selection, window):
 
     crop = [0.0, 0.0, 0.0, 0.0]
     if selection:
-        path = selection[0]
-        pos = model.get_iter(path)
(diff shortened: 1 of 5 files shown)


jeromerobert:
I do the PR from this repo and not my own personnal repo, because if I fork pdfarranger/pdfarrager it will break the redirection of jeromerobert/pdfarranger to this repo.

jeromerobert:
Setting as draft, as I need to do more tests.

kbengs:
I think its good, but:
Zoom in/out does not work
First comment in Page class has a extra "is"
Some comments don't start with capital letter ;-)

I may add more comments here if I find something

kbengs (approved):

Issue #199, opened 2020-05-09
Support for page scaling


yodatak:
It can be interesting to resize a pdf from A4 to A5 or A3 to A4 or something ^^

jeromerobert:
From what I read, this is currently [not possible with qpdf](https://github.com/qpdf/qpdf/issues/371), so it's not with pikepdf, so it's not with pdfarranger. I keep that bug open but I suggest you add a comment on the qpdf bug to express your need.

jberkenbilt:
It can certainly be done with qpdf, not just through the CLI. A client of libqpdf could modify the bounding boxes and, if desired, surround the page content by a suitable transformation matrix. However, there is no libqpdf call that does this for you. I may add one, and if so, it will be connected with the referenced issue.

merwok:
Does resizing here mean to fit two original pages on one page?
I have installed a bunch of texlive packages in order to get `pdfjam` and try this (not easy to find the right combination of angle and nup) before I found out my CUPS driver gave me options for this.
It would be great to see it in PDFArranger.

jeromerobert:
@merwok No it's not. You are talking about copying content of one page to an other existing page. It's a different issue. Here it's about scaling.

jberkenbilt:
For what it's worth, I'll mention that libqpdf has methods for overlaying one page on another by treating the page as a form XObject. qpdf CLI doesn't currently support 2-up type functionality, but cups filters uses the qpdf library to do that exact thing, and the library has enough calls to make it pretty easy for a client developer. If you'd like to see that in the qpdf CLI, create an issue [there](https://github.com/qpdf/issues). But if this is something pdfarranger might want to do sometime, be aware that the supporting functionality exists in qpdf. The qpdf cli's overlay/underlay feature would require only slight tweaking to do n-up. If I add support for page scaling, I might also add support for n-up.

But this is about pdfarranger, not qpdf, so I will refrain any further discussion about qpdf here except as relevant to this issue.

jeromerobert:
Thank you @jberkenbilt I missed that libqpdf was that powerful. The next step is to dig in pikepdf:
* https://pikepdf.readthedocs.io/en/latest/topics/content_streams.html#editing-a-content-stream seems to be what we need for scaling
* From https://github.com/pikepdf/pikepdf/issues/42 it's already possible to do overlaying (and not that easy).

dreua:
> RFE: With increasing import capabilities there is also increasing need to scale pages to a common size.
[...]

PR #277, opened 2020-09-10
Add page scaling

jeromerobert:
This is my first draft for #199. It generally works but I'm not yet satisfied because:
- I think crop and scale should be merged in the same dialog/action
- We should support absolute scaling (setting page size in mm for all selected pages) and not only relative scaling (applying a constant scale factor). We should support absolute crop also but that will be for an other branch/PR.
- the pikepdf part fail with some PDF files. For example images may disappear. I've not been able to reproduce this on PDF files I can share.

### DIFF
--- a/data/menu.ui
+++ b/data/menu.ui
@@ -80,8 +80,8 @@ along with PDF-Arranger.  If not, see <http://www.gnu.org/licenses/>.
           <attribute name="target" type="i">90</attribute>
         </item>
         <item>
-          <attribute name="label" translatable="yes">Cro_p</attribute>
-          <attribute name="action">win.crop</attribute>
+          <attribute name="label" translatable="yes">_Page Format</attribute>
+          <attribute name="action">win.page-format</attribute>
         </item>
         <item>
           <attribute name="label" translatable="yes">Crop White Borders</attribute>
@@ -288,8 +288,8 @@ along with PDF-Arranger.  If not, see <http://www.gnu.org/licenses/>.
         <attribute name="target" type="i">90</attribute>
       </item>
       <item>
-        <attribute name="label" translatable="yes">Cro_p</attribute>
-        <attribute name="action">win.crop</attribute>
+        <attribute name="label" translatable="yes">_Page Format</attribute>
+        <attribute name="action">win.page-format</attribute>
       </item>
       <item>
         <attribute name="label" translatable="yes">Crop White Borders</attribute>
--- a/pdfarranger/croputils.py
+++ b/pdfarranger/croputils.py
@@ -17,9 +17,138 @@
 from gi.repository import Gtk
 import gettext
 import cairo
-
+import locale
 
 _ = gettext.gettext
+
+
+def scale(model, selection, factor):
+    """Set the scale factor of a selection of pages."""
+    changed = False
+    try:
+        width, height = factor
+    except TypeError:
+        width, height = None, None
+    for path in selection:
+        it = model.get_iter(path)
+        page = model.get_value(it, 0)
+        if width is None:
+            f = factor
+        else:
+            # TODO: allow to change aspect ratio
+            f = max(width / page.size[0], height / page.size[1])
+        # Page size must be in [72, 14400] (PDF standard requirement)
+        f = max(f, 72 / page.size[0], 72 / page.size[1])
+        f = min(f, 14400 / page.size[0], 14400 / page.size[1])
+        if page.scale != f:
+            changed = True
+        page.scale = f
+        model.set_value(it, 0, page)
+    return changed
+
+
+class _LinkedSpinButton(Gtk.SpinButton):
+    """ A spin button which can be bound to an other button """
+
+    def __init__(self, minval, maxval, step, page=None):
+        Gtk.SpinButton.__init__(self)
+        self.set_digits(20)
+        self.set_width_chars(9)
+        self.connect("output", self.__output)
+        self.set_range(minval, maxval)
+        self.set_increments(step, step if page is None else page)
+        self.changing_from_brother = False
+
+    def __output(self, _user_data):
+        """ output signal handler to remove unneeded 0 """
+        s = locale.str(self.get_adjustment().get_value())
+        self.get_buffer().set_text(s, len(s))
+        return True
+
+    def connect_brother(self, brother, func):
+        """ Ensure a broher spinbutton have func(get_value()) value """
+        self.brother_func = func
+        self.connect("value-changed", self.__change_brother, brother)
+
+    @staticmethod
+    def __change_brother(self, brother):
+        if not self.changing_from_brother:
+            brother.changing_from_brother = True
+            brother.set_value(self.brother_func(self.get_value()))
+            brother.changing_from_brother = False
+
+
+class _RadioStackSwitcher(Gtk.VBox):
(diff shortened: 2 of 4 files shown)


kbengs:
Could we have radio buttons here (also) so we could select A3, A4 or A5 sizes?

jeromerobert:
Yes, I was thinking of adding a combobox with A3, A4... in the "Absolute" panel. But this will be the subject of another pull request. Here I will just merge crop and scale actions. I also need to fix undo/redo.

kbengs (commented):
:thumbsup: This is a good feature. What I notice:
* Absolute scaling does not always work when trying to make a page smaller
* Imo the word "page" would be better than "pages" in the dialog
* Zoom to full page needs a fix. I can fix it later also if you don't want to
* "= power" looks complicated. Is it needed?
* When aspect ratio is locked both width and height should ideally change at the same time
* When a width or height value is written and enter is pressed, it doesn't execute and close dialog
* Docstring for _AbsoluteScalingWidget has text relative
* If we do a relative scaling with 100% I would expect nothing to happen, and normally nothing does happen. But if that is done after a page is cropped then the cropping seems to be removed (if many pages are selected including the cropped page)

jeromerobert:
> "= power" looks complicated. Is it needed?
> When aspect ratio is locked both width and height should ideally change at the same time

The problem is that pages may have different sizes and different ratio.

The *2 power x* field allow to switch from An to An+1 (ex: A3 to A4) even if pages have different size. I admit this is not intuitive but I did not found anything better.

As I don't know the ratio of pages I should probably only allow to set either the width or the height, but not both. I don't plan to support change of ratio in this PR.

kbengs:
Okay I see. 
>As I don't know the ratio of pages I should probably only allow to set either the width or the height, but not both.

Maybe you could have both but write "---" or something to the one that is not in focus

angsch:
@jeromerobert This feature is awesome, well done!

Adding to kbengt's list:

* The display of the absolute measures and the relative scaling are not perfectly consistent. I import an A4 page, set the relative scaling to 2^-4. The true (and correct) size is 13mm x 19mm. The absolute measures display 25.4mm x 25.4mm, which is the minimum value the text field accepts.
* I import an A4 page, set the scaling to the maximum possible value 2^13, press ok. After opening the crop dialog again, the display is 2^12.2288.

> The 2 power x field allow to switch from An to An+1 (ex: A3 to A4) even if pages have different size. I admit this is not intuitive but I did not found anything better.

To the best of my knowledge, only the A format series (A3, A4, ...) has this property. An analogous scaling of a North America paper size (US letter etc.) yields a non-standard paper size. This makes me think that this feature is primarily useful when the input document is a valid A format. One possible solution could (later) be through the "absolute" dialog, explicitly offering A4, A3, US letter... as possible target page sizes. That way, it would also be possible to convert A4 to US letter, which would actually be huge. In fact, this could fix the dilemma with inconsistent page sizes in the selection. Instead of displaying any of the current page sizes, the target page size (the default value could be unrelated to the sizes of the selected pages) could be displayed.

jeromerobert:
>  The absolute measures display 25.4mm x 25.4mm, which is the minimum value the text field accepts.

This is the minimum PDF page size. Because of different page size and page ratio there is no way to make all field consistant. You request for a page size, but this may be truncated by the PDF page size limit. Maximum page size is 200x200 inch and minimum 1x1 inch.

> I import an A4 page, set the scaling to the maximum possible value 2^13, press ok. After opening the crop dialog again, the display is 2^12.2288.

Ok, I should limit that field to ±log(200)/log(2)≈7.6, not ±13.

> To the best of my knowledge, only the A format series (A3, A4, ...) has this property.

B and C format also have that property.

>  That way, it would also be possible to convert A4 to US letter,

That one is a bit more complex as it's a mix of scaling and margin change. If we do it only with scaling this will change the aspect ratio of the page.

@kbengs @angsch I don't feel satisfied with the relative scaling feature and it seems that you don't either :-). May be I should simply drop it. Documents with different page size are not that common. Rescaling a selection with different page size will be even less common. It will still be possible to rescale pages of different size one by one. I would just allow to set the page width of the selection or (exclusive or) the height of the page. What do you think ?

angsch:
I find the relative scaling interesting, but I am not sure how useful it is to a user. For absolute scaling I can think of a few use cases and it is certainly covers the relative scaling case.

> Rescaling a selection with different page size will be even less common.

I agree. The only thing that I see regularly and is somewhat in that direction is a PDF with portrait pages and a few pages with wide tables that are rotated in landscape format. Technically, it is the same page size, but for those cases it should be fair to assume that the user can scale the pages separately.

kbengs:
Yes I also would say you could drop it. I think absolute scaling is what most users want to have. And dialog will be simpler without relative scaling.

Nokia808:
Just I would like to give the following note. If we considered the scanned PDF files:

1) we will get frequently PDF files that need to rescale certain of it's pages to specific size which is size of remaining of majority of pages. 

2) Also, there are many scanned PDF books that each page has it's own size that differ from size of remaining pages. This happen when scanner software set at "automatic cropping" during scanning of such books, & this will result in pages of different sizes due to different contents in each of them which will cause each page to have it's specific size .....

I'm not understanding what is absolute scaling & relative scaling, but I like to bring your attention that selection of MULTIPLE pages within a pdf or ALL it's pages AT ONCE to rescale them to specific UNIFIED size in the aim to applied it for ALL SELECTED pages AT ONCE, is needed from time to time & not uncommon so it should be covered by this feature .....

jeromerobert:
> If we do a relative scaling with 100% I would expect nothing to happen, and normally nothing does happen. But if that is done after a page is cropped then the cropping seems to be removed (if many pages are selected including the cropped page)

I think I fixed all issues but that one. That one would require some deep changes in the crop dialog and I would prefer to do it later. Moreover cropping multiple pages which have already been cropped with different values is already a bit broken.

We now have 3 exclusives way to resize:
- Relative factor (without the too complicated 2^n field)
- Set width
- Set height

angsch (commented):
Here are a few comments:

### Functionality
1.  The +/-/scale spin buttons are disabled when the global minimum/maximum width/height/scale is reached. When one out of width/height/scale reaches the extremum, the other two should be disabled. To reproduce: (1) Load a portrait A4 page, (2) set width to minimum value 25,4 mm. The height is 36,659mm and can be reduced further, but changing does not have any effect. I think that this is something that can be fixed later.
2.  The displayed page size refers to the original page size before applying cropping. Scaling to the minimum width 25,4 mm is possible, in the exported PDF the real width is 13 mm. To reproduce: (1) Load a portrait A4 page, (2) split, (3) the width in the crop dialog is the A4 width, not as the cropped width, (4) set width to minimum value (5) export. 

It could be worth dropping the goal of producing ISO-compliant PDFs. Many PDF readers can handle non-compliant PDFs. If a user manipulates the width/height, this may be the size the user wants (e.g., to produce name tags, advertisement posters, ... ) even if the size is not ISO-compliant.

### Code
1.   `brother_func` [line 71] is not defined in `__init__()` [line 53]. To be on the safe side, consider adding `self.brother_func = None` to `__init__()`
2.   Typo: line 70 broher -> brother

grahamperrin:
I have `1.7.0rc1` but (sorry) I can't see how to perform scaling …

jeromerobert:
@grahamperrin See *Page Format*

PR #282, opened 2020-09-23
Disable reverse order after selection-splitting actions

angsch:
'Reverse order' works on a contiguous selection. Actions that lead to a non-contiguous selection (duplicate, split, paste interleave odd/even) should disable 'reverse order' in the right-click menu.


jeromerobert:
It would better to call `iv_selection_changed_event` (or even better find why it's not automatically called even if the selection changed). Note that if you call `iv_selection_changed_event` you can remove calls to `__update_statusbar`.

PR #286, opened 2020-09-25
Refactor the croputils.dialog function to 2 classes

jeromerobert:
Separate the crop frame and the dialog code. So it will be easier to add a new
frame for the scaling feature.


kbengs (commented):
It looks good to me.

Issue #292, opened 2020-10-05
Change meaning of file filter 'all files' to 'all supported files'


angsch:
Hi,

pdfarranger supports PDF files and, if img2pdf is available, a variety of image formats. In the import dialog, the default file filter is set to PDF files. Alternative file filters are 'all files' and 'supported image files'. I suggest to change the meaning of 'all files' from literally all files to all supported files and make this the default filter. I wonder how many users are actually aware that images are supported when the default filter is PDF files.

Other considerations:
* For software on Ubuntu, the proposed change appears rather standard. For example, evince uses 'all documents' as default to list all supported file formats. 'all files' is also available and lists all files.
* The file opener dialog of the default PDF viewer on Mac lists all files, but grays out all non-supported files. I find this solution neat, but I could not find any Gtk feature that would enable this.
* It is not clear to me what the default behavior on Windows is.

jeromerobert:
> I suggest to change the meaning of 'all files' from literally all files to all supported files

I would prefer the filter "all supported files" to be added in addition to the "all files" filter. This is what evince do.

> and make this the default filter.

+1

> It is not clear to me what the default behavior on Windows is.

It should go well

Issue #289, opened 2020-09-28
Add Ctrl+o as accelerator of import action


angsch:
Hi,

Currently, the only accelerator for importing files is the insert key. What do you say about adding Ctrl+o and having two accelerators (the insert key and Ctrl+o) for the import action? Even after using pdfarranger for quite some time, I still try Ctrl+o, realize that this is the wrong accelerator and then use the import button (rather than the proper accelerator key). I browsed over the currently defined accelerators and Ctrl+o appears to not be in use.

jeromerobert:
Indeed `insert` is unusual for such action. Gimp and Inkscape use Ctrl+I so I would prefer that we replace `insert` with `ctrl+I`.

angsch:
Ctrl+I sounds intuitive to me as well. Would you keep the old accelerator (and have two accelerators in total) or would you replace it?

kbengs:
Applications seems to have varying action for Ctrl+I. (indent, invert, information, import, filter, properties..) Could Ctrl+O still be better as it is well known?

PR #294, opened 2020-10-06
Add filter 'all supported files' (#292); change accelerator (#289)

angsch:
A new file filter 'all supported files' is introduced and made default. The accelerator of the import action is changed to Ctrl + o.


kbengs:
I think this is good. Only thing I could comment on is that the export dialog maybe don't need to have "all supported files" as we can only export to pdf ?

jeromerobert:
I merged it to 2 different commits (55181eb and 12a69c0) and changed the commit message according to:
https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

PR #299, opened 2020-10-11
Update translations

milotype:
- Update for Croatian and German translations with current .pot file
- New Croatian translation for .desktop file


jeromerobert:
Merged in 65c10c8. German changes were automatic changes so I dropped them.

Issue #296, opened 2020-10-10
Selection using navigation keys


angsch:
Hi,

I want to extract the last few pages from a document. 1) Select the start page through a mouse click, 2) hold Shift 3) press end key. The selection encloses all pages in the rectangle defined by the start page and the last page. If the last page is not in the bottom right corner the selection is view-dependent.

Example:  The start page is 1, depending on the window geometry, Shift + End gives different selections.

![image](https://user-images.githubusercontent.com/17718454/95654551-fdbf2480-0b00-11eb-9ff2-6d657c196ec3.png)

Scrolling to the last page and clicking on it, gives a contiguous selection. This is what I assumed would happen when using the end key. The home key produces the same issue.

I suggest do change the behavior for the home/end key to either of the two:

- The navigation keys (home, end, page up, page down) are a replacement for scrolling. Then it would make sense that the navigation keys only focus the first (home) or last (end) page, but _not_ select it. Page up/down works like this already, the keys change the focus, but do not select anything.
- The home/end keys are a replacement for scrolling _and_ clicking. Then the selection should be contiguous.

Personally, I prefer the first option.

kbengs:
Nautilus works as the second option.. I am not sure what I would prefer, I never use home/end, but as Page up/down works as it works maybe it makes more sense that home/end would work as the first option also.

kbengs:
What do you think about shift + keyboard arrows selection behavior, should it be as it is now or should it do a continuous selection like for example Nautilus do? (select all pages between first selected page and page where cursor is)

angsch:
Being consistent with Nautilus makes sense, both for the navigation keys and the arrow keys.

We could add a "go to page <page number>" option later to ease navigation and support the semantics of the second option. Simple navigation makes sense if the information is scattered, but the position is known. For example, to extract all figures whose page numbers are known from a table of figures. Not sure though how many users need this.

kbengs:
>Being consistent with Nautilus makes sense, both for the navigation keys and the arrow keys.

Do you want to do this issue @angsch ?

angsch:
@kbengs I am not really familiar with the iconview and also a bit short on time until the next weekend. If you have time and want to take over, I would be very glad.

jeromerobert:
3bd8a23

PR #303, opened 2020-10-22
Change behavior when navigating with keyboard keys

kbengs:
Draft for #296
* When navigating in iconview with left/right arrow, continue on
  previous/next line when end of a line is reached
* When selecting with keyboard navigation keys, select all pages
  between cursor and page where cursor was when shift was pressed

The reason for the complicated unselect approach is that plain unselect_all() is too slow for example when shift + down arrow is pressed and the key is auto repeated.

There is a unsolved issue here: if a page is selected with mouse click and you try to continue to move cursor with arrow keys, the cursor will be invisible (the dashed rectangle is not visible).
Any ideas or comments so far?

### DIFF
--- a/pdfarranger/iconview.py
+++ b/pdfarranger/iconview.py
@@ -15,6 +15,7 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 from gi.repository import Gtk
+from gi.repository import Gdk
 from math import pi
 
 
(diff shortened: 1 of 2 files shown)


jeromerobert:
I think it's good to be merged.

This is nothing new with this pull request, but we are triggering too many `selection_changed` events (`iv_selection_changed_event`). `Ctrl+A` trigger few events, but selecting the first page then doing `shift+end` trigger at least one event by page. The reason for this is that `Gtk.IconView` only allow programmatic selection of one page at a time (`select_path`). Each `select_path`do a lot of things (https://github.com/GNOME/gtk/blob/3.24.23/gtk/gtkiconview.c#L3365). We should disable our `selection_changed` while doing large selection changes. This could look like this:

```python
with iconview.edit_selection(): #__enter__ disable event handler
    # do all selection changes
    iconview.unselect_all()
    for x in foobar:
        iconview.select_path(x)
    #__exit__ restore the event handler
```     
Unfortunnatly there is nothing we can do about `gtk_icon_view_item_selected_changed` and `gtk_icon_view_queue_draw_item` calls, they cannot be disabled.

Anyway this problem is not perceptible and so by far not critical.

kbengs:
> but we are triggering too many selection_changed events 

Yes I noticed that a auto-repeated shift + down arrow will cause my older laptop to lag. If I block selection_changed event it works fine. So I think we have to block it as long as shift is pressed and unblock when key is released. I will try to make a fix for that and also I would like to find a solution for the invisible cursor.

kbengs:
It should be okay now.

jeromerobert:
You're a GTK guru :-)

PR #308, opened 2020-10-27
Allow editing the creation and modification date

### LINKED ISSUE #283: Allow to edit creation and modification date
Hi.

As title saying:  it is possible to edit date of creation & date of modification of PDF so that to make them empty ?

I think that at least date of modification is possible to make it some thing like this 00:00:00:00:00

Best.

angsch:
Closes #283 

- The user input of a date is parsed into the required format.
- Invalid date inputs are erased to have a smooth export.
- Two error checks are necessary. 1) In the standard case, when the user presses enter or changes rows, the `edited` signal is emitted and the input is validated. 2) It is possible to bypass the `edited` signal by pressing OK *during* the editing process. To capture this case, the implemented solution does an additional validation when the metadata is dumped.

### DIFF
--- a/pdfarranger/metadata.py
+++ b/pdfarranger/metadata.py
@@ -21,6 +21,8 @@ import gettext
 import re
 import json
 import traceback
+from datetime import datetime
+from dateutil import parser
 from gi.repository import Gtk
 _ = gettext.gettext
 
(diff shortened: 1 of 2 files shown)


jeromerobert:
It looks good. The only thing missing is the dependency on `dateutil` in `setup.py`.

angsch:
Good spot, now fixed.

PR #312, opened 2020-11-01
Allow to customize keyboard shortcut from configuration file

### LINKED ISSUE #276: Allow to customize keyboard shortcut from configuration file
Might be Ctrl-Shift-S.

jeromerobert:
Close #276

### DIFF
--- a/pdfarranger/config.py
+++ b/pdfarranger/config.py
@@ -111,10 +111,10 @@ class Config(object):
         self.data.set('window', 'maximized', str(maximized))
 
     def zoom_level(self):
-        return self.data.getint('DEFAULT', 'zoom-level', fallback=0)
+        return self.data.getint('window', 'zoom-level', fallback=0)
 
     def set_zoom_level(self, level):
-        self.data.set('DEFAULT', 'zoom-level', str(level))
+        self.data.set('window', 'zoom-level', str(level))
 
     def save(self):
         conffile = Config._config_file(self.domain)
--- a/pdfarranger/config.py
+++ b/pdfarranger/config.py
@@ -1,0 +1,72 @@
+# Copyright (C) 2020 pdfarranger contributors
+#
+# pdfarranger is free software; you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


pdfarranger/config.py
```diff
@@ -0,0 +1,146 @@
...
+    ('select(0)', '<Ctrl>a'),
+    ('select(1)', '<Ctrl><Shift>a'),
+    ('main-menu', 'F10'),
+]
+
+class Config(object):
+    """ Wrap a ConfigParser object for PDFArranger """
+
+    @staticmethod
+    def __get_action_list(m, r=[]):
```
dreua, line 51: [Codacy actually spotted an error here](https://app.codacy.com/manual/jeromerobert/pdfarranger/file/49786423434/issues/source?bid=13283750&fileBranchId=20839044#l51): The empty list is only initialized once the class is loaded so for subsequent calls r already contains items from previous runs. Made this mistake once and it was a nightmare to figure it out. I think the best solution is `r = None` and `if not r: r = []`

jeromerobert: Thanks. Fixed. I just removed the default value.

angsch:
The GTK way of writing [platform-independent accelerators](https://mail.gnome.org/archives/commits-list/2011-September/msg06655.html) is `<Primary>` instead of `<Ctrl>`.

jeromerobert:
@angsch Ok. This is not directly related to this PR so I'll do the change in an other commit.

kbengs:
I realize accelerators can be set in config.py, but can it be set in a .ini file also? If so should a explanation/example be included?

kbengs:
I saw "Unable to parse accelerator '-10': ignored request to install 1 accelerators" in terminal when I started it.

zoom-level = -10 was found under accelerators and also under DEFAULT in config.ini
After removing -10 I see no warning any more.

jeromerobert:
Thank you for the review. Both items should now be fixed. I'll merge by the end of the week.

PR #314, opened 2020-11-02
Add option to select all pages with the same page format

angsch:
A selection based on the page format facilitates the use of the absolute page scaling feature.

The most useful case is probably that a single page is selected and then the selection is extended to all pages of the same size. The implementation allows for multi-selection. I see some use cases when absolute scaling is applied incrementally or in conjunction with '`invert selection`.

### DIFF
--- a/data/menu.ui
+++ b/data/menu.ui
@@ -156,16 +156,21 @@ along with PDF-Arranger.  If not, see <http://www.gnu.org/licenses/>.
           <attribute name="action">win.select</attribute>
           <attribute name="target" type="i">3</attribute>
         </item>
-        <item>
-          <attribute name="label" translatable="yes">All From _Same File</attribute>
-          <attribute name="action">win.select-same-file</attribute>
-          <attribute name="target" type="i">4</attribute>
-        </item>
-        <item>
-          <attribute name="label" translatable="yes">_Invert Selection</attribute>
-          <attribute name="action">win.select</attribute>
-          <attribute name="target" type="i">5</attribute>
-        </item>
+          <item>
+            <attribute name="label" translatable="yes">All From _Same File</attribute>
+            <attribute name="action">win.select-same-file</attribute>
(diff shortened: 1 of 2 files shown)


kbengs:
Yes this should be useful. Think you forgot hamburger menu. :-) And if you put it there it should be greyed out if nothing is selected.

angsch:
Thanks, kbengs.

jeromerobert:
I missed it in #290 but I would have expected either (A):

```python
            ('select', self.on_action_select, 'i'),
```

or (B)

```python
            ('select-xxx', self.on_action_select_xxx),
            ('select-same-file', self.on_select_same_file),
            ('select-same-format', self.on_select_same_format),
```

but not (C)

```python
            ('select', self.on_action_select, 'i'),
            ('select-same-file', self.on_action_select, 'i'),
            ('select-same-format', self.on_action_select, 'i'),
```

To me, the benefit of A is that it's short to write (no need to write a function for each type of select).

The benefit of B is that it will make shortcut configuration file and the `.ui` files easier to read and maintain. If we chose A or C we won't be allowed to change the integer of `Invert Selection` from 5 to 6 as here.

In C the selection type is identified with both its name and its integer ID. I would say one is enough.

@kbengs what's your opinion ?

kbengs:
We need to grey out 'select-same-file' and 'select-same-format' when nothing is selected. Can the lookup_action function handle 'i' arguments? If it can be done somehow I think (A) would be nice. But if not I think the current solution is quite clean, all select types use a common function.

@angsch menu.ui has same extra indentation that cases unneeded changes.

jeromerobert:
> Can the lookup_action function handle 'i' arguments?

Ok I missed that point.

jeromerobert:
Removed extra indentation in menu.ui and merged as 0e8a9e545e289e3eba0012edf172aee4ea9000f6

PR #318, opened 2020-11-07
Select a continuous range when drag-selecting

kbengs:
The initial idea to use event_after and select the range of pages doesn't do what we want when selecting in top-right <-> left-bottom direction. We will then get some bonus pages selected. So here is another approach.

* Click between pages and drag mouse pointer to select all pages
  between click-start location and current mouse pointer location
* When ctrl is pressed, toggle selection for drag-selected items
* Cursor is changed to text-selecting cursor to give a hint that we are selecting like a text-editor

I think this is better than the current rubberband selecting. What do you think?


kbengs:
Fixed conflict and added shift + drag-select handling.

PR #319, opened 2020-11-07
menu: Add 'new' button to open new document

jadahl:
I was arranging PDFs, and was going to organize multiple files into
multiple PDFs. What I was missing was a way to create a new empty
document from withit PDF Arranger; this commits adds that.

It simply works by spawing a new instance of PDF Arranger using the app
launch context from Gdk, using the .desktop file of PDF Arranger itself,
which is what I was in fact was doing to create new documents, but via
the launch mechanism in the desktop environment.

### DIFF
--- a/data/menu.ui
+++ b/data/menu.ui
@@ -27,6 +27,12 @@ along with PDF-Arranger.  If not, see <http://www.gnu.org/licenses/>.
     </section>
     <section>
       <item>
+        <attribute name="label" translatable="yes">_New Window</attribute>
+        <attribute name="action">win.new</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
         <attribute name="label" translatable="yes">_Save</attribute>
         <attribute name="action">win.save</attribute>
         <attribute name="verb-icon">document-save-symbolic</attribute>
--- a/pdfarranger/config.py
+++ b/pdfarranger/config.py
@@ -30,6 +30,7 @@ _DEFAULT_ACCELS=[
     ('export-selection(2)', '<Primary>e'),
     ('export-all', '<Primary><Shift>e'),
     ('quit', '<Primary>q'),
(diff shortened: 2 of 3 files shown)


jeromerobert:
Unfortunately, relying on `DesktopAppInfo` is currently not very portable. It fails under Windows and when pdfarranger is installed with `pip`:

```
  File "/path/to/pdfarranger/pdfarranger.py", line 777, in on_action_new
    app_info = Gio.DesktopAppInfo.new(desktop_file)
TypeError: constructor returned NULL
```

I can think of 3 possible solutions:
* Ensure that the `.desktop` is found when installing with `pip` and `setup_win32.py`. I don't know if it's possible and what it entails.
* Use `sys.executable` and `subprocess` in place of `DesktopAppInfo`. It should be easy.
* Disable and hide the `New` action when it does not work. Really not my favourite.

jadahl:
> Use sys.executable and subprocess in place of DesktopAppInfo. It should be easy.

This sounds somewhat doable, but the problem I see with it is that it doesn't handle everything related to startup notifications on Linux. We get that for free by using app info and the launch context from gtk. Maybe the `.desktop` can be used when it's found, and then fall back on `sys.executable` if that fails?

jeromerobert:
> Maybe the .desktop can be used when it's found, and then fall back on sys.executable if that fails?

Yes it will be perfect.

kbengs:
Many applications use name "New Window" I would suggest we use that.

jadahl:
> Many applications use name "New Window" I would suggest we use that.

Changed to "New Window".

> Yes it will be perfect.

I now have a fallback path that practically runs

```python
sys.executable -mpdfarranger
```

Is there any better way than passing `-mpdfarranger` to `sys.executable`?

jadahl:
Also fixed conflicts with master.

PR #322, opened 2020-11-10
2 fixes for accelerators

jeromerobert:


angsch:
Looks in general good to me. I found only one small problem related to conflicting accelerators.

It is possible to create a conflict if the default accelerator map does not define an accelerator. To reproduce:
~~~
export-selection(2) = <Primary>e
export-selection(3) = <Primary>e   # added - the default does not define an accelerator for this action
~~~
For this example, the accelerator shortcut triggers `export-selection(2)`.

PR #323, opened 2020-11-11
Change behavior when selecting with shift + click

kbengs:
This is consistent with Nautilus shift-click-select

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -1393,35 +1393,41 @@ class PdfArranger(Gtk.Application):
                 self.zoom_to_full_page()
             return True
 
-        x = int(event.x)
-        y = int(event.y)
-        self.click_path = iconview.get_path_at_pos(x, y)
+        click_path_old = self.click_path
+        self.click_path = iconview.get_path_at_pos(event.x, event.y)
 
         # Go into drag-select mode if clicked between items
         if event.button == 1 and not self.click_path:
             self.iv_drag_select.click(event)
+            if event.state & Gdk.ModifierType.SHIFT_MASK:
+                return 1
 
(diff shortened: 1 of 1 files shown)


angsch:
- The return value is sometimes `True`, `1` or missing in some branches.

- I found a case that behaved differently from what I expected.
    1. Load a PDF with 4 pages. 
    2. Select all pages through clicking on page 1, holding shift and clicking on page 4.
    3. Remove page 2 from selection by holding Control and clicking on page 2.
    4. Hold shift and click on page 3.

    All pages are selected. I expected that only the pages 1--3 are selected (which is what Nautilus selects). Other than that new selection works really smoothly, nice work.

kbengs:
Thank you for testing.

> The return value is sometimes True, 1 or missing in some branches.

Yes I know I haven't been consistent with the return value. Documentation I have found say that True should be returned to stop further processing. But 1 also works. I don't know if 1 is a old recommendation.
We still use some of the iconviews built in functions, therefore True or 1 is not always returned. (For example when a page is clicked it is selected and cursor is set there.)
> Hold shift and click on page 3

It is behaving like Dolphin in this aspect. When shift is held, what is already selected does not get deselected. Imo this is better. The new drag-select also work the same way: when shift is pressed it does not deselect, but instead it allows you to select more pages.
(The result after step 4 on Nautilus on my machine is that only page 3 is selected)

angsch:
> It is behaving like Dolphin in this aspect. [...]
(The result after step 4 on Nautilus on my machine is that only page 3 is selected)
The behavior of Nautilus on your system surprises me. I wonder if it is triggered by custom settings or the theme. I had check with Dolphin and in a fresh install I get the behavior you describe. So please ignore my comment.

jeromerobert:
It looks good and ready for merge.

PR #325, opened 2020-11-14
Warn when TOC/forms are lost on export

### LINKED ISSUE #265: Warn about loss of attributes like table of contents and links
Using 1.4.2 from Debian. Please ignore if this is not relevant to latest versions.

Recently I tried to crop and rearrange a few pages in an ebook. When I saved it, the table of contents and links were gone in the resulting file. Thankfully, you don't automatically overwrite the original file, so I was able to go back to it.

I realize that these more complex PDF features are not within the scope of this project. But I suggest that you warn users if these exist in the original file and will be lost when saving. Some people may not check this before they delete the original file or send the new one to others.

angsch:
Proposal to close #265

Fillable forms and outlines are lost on export. This commits screen for forms and outlines and displays a warning if any imported document contains such. The warning can be disabled and the choice is stored in the config file. I consider this a compromise that is not too annoying.

There are two things I am not happy with and ask you for your opinion.
- In the config file, the newly added `disable-warnings` feels misplaced in the category `[window]`. Opening a separate category for a single item does not appear reasonable either. Are more customizations expected so that a category `[preferences]` is justifiable?
- The warning appears if any of the documents has forms or outlines, even if the pages selected for export are not affected. The reason is that I cannot find an easy way to get the page on which a form is located.

### DIFF
--- a/pdfarranger/config.py
+++ b/pdfarranger/config.py
@@ -88,6 +88,8 @@ class Config(object):
         self.data = configparser.ConfigParser()
         self.data.add_section('window')
         self.data.read(Config._config_file(domain))
+        if 'preferences' not in self.data:
+            self.data.add_section('preferences')
         if 'accelerators' not in self.data:
             self.data.add_section('accelerators')
         a = self.data['accelerators']
@@ -118,6 +120,12 @@ class Config(object):
 
     def set_zoom_level(self, level):
         self.data.set('window', 'zoom-level', str(level))
+
+    def content_loss_warning(self):
+        return self.data.getboolean('preferences', 'content-loss-warning', fallback=True)
+
+    def set_content_loss_warning(self, enabled):
+        self.data.set('preferences', 'content-loss-warning', str(enabled))
 
     def save(self):
         conffile = Config._config_file(self.domain)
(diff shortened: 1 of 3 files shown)


jeromerobert:
It looks good ! May be you should rename the `warning` configuration key to something like `content-loss-warning` as the first one is very generic.

jeromerobert:
Yes `[preferences]` a section would be good. And the existing `zoom-level` key should be moved there.

> The warning appears if any of the documents has forms or outlines, even if the pages
> selected for export are not affected. The reason is that I cannot find an easy way to get the
> page on which a form is located.

It does not matter, even with that your PR is an improvement.

kbengs:
Looks good to me too. I only notice a tiny thing: When the warning opens and you hit enter it will select "Do not..". I would expect it to execute (hit enter). But I might be wrong. And I don't know how hard it is to change that.

angsch:
Thank you for your suggestions. The relocation of `zoom-level` to `[preferences]` should be a separate commit, do you want to do this, Jérôme?

jeromerobert:
> The relocation of zoom-level to [preferences] should be a separate commit, do you want to do this, Jérôme?

Yes. done: 6164b7a

PR #328, opened 2020-11-19
Zoom to full page faster

kbengs:
With the current implementation of "zoom to full page" the zoomed and rendered page will be visible after all pages have been completely rendered. With a pdf with many pages it can take a long time. These commits will make it zoom to full page much faster.

All of this is based on that we set a fixed cell height. ;-) But It is only enabled when we zoom to full page, so I think it is a good compromise.


jeromerobert:
This is indeed much better !

PR #337, opened 2020-11-28
Make New Window action work under Windows

kbengs:


kbengs:
Now codacy is happy but tests and coverage is not. I guess this is a problem with the test.

jeromerobert:
> Now codacy is happy but tests and coverage is not. I guess this is a problem with the test.

Yes, some timeout need to be extended.

PR #342, opened 2020-12-05
Update README.md

sahwar:
Add note on the open-source PDFsam app which is a more heavyweight alternative to pdfarranger.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ interactive and intuitive graphical interface. It is a frontend for
 [Sourceforge](http://sourceforge.net/projects/pdfshuffler)).
 It’s a humble attempt to make the project a bit more active.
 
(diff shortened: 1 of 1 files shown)


jeromerobert:
What about pdfmixertool, pdfchain and pdftk ? A blog post dedicated to open source PDF GUI would be the right place for this, or may be just https://alternativeto.net/software/pdf-arranger/.

sahwar:
> What about pdfmixertool, pdfchain and pdftk ? A blog post dedicated to open source PDF GUI would be the right place for this, or may be just https://alternativeto.net/software/pdf-arranger/.

Great idea! Please also mention these tools in the README.md next to PDFsam! It is good to list the 'competition' apps! 👍 As far as I have tested, my statement that pdfranger (mostly written in Python3.x) is more lightweight than PDFsam (mostly written in Java) - stands correct!

So I propose that we include all these apps in a list in a new subsection of `pdfranger`'s README.md where pdfranger is compared to the other apps in a Markdown comparison table! 👍

Go ahead and commit that change IF you agree with that!

DISCLAIMER: I like pdfranger and I prefer it over the more heavyweight but more full-featured PDFsam.

PR #348, opened 2020-12-17
Fix MediaBox check

angsch:
The typo causes all exports to be on US letter.


jeromerobert:
Oops, thank you so much.

PR #351, opened 2020-12-19
Move some classes from pdfarrager to core

jeromerobert:
* To make the pdfarranger module less bloated
* Before more meaningful changes in PDFDoc to support PDF encryption.


kbengs:
Looks good imo. I saw no issues when testing it either.

PR #354, opened 2020-12-24
Allow import of encrypted PDF files

jeromerobert:


angsch:
Looks fine to me and I have only two minor comments.
- Image that you import several password-protected PDFs at once. In what order to you have to enter the passwords? In particular, it is difficult to identify the right PDF when you mistyped the password once. Can you add the name of the PDF prompting the password to the dialog?
- I get a Gtk warning for the password dialog `GtkDialog mapped without a transient parent. This is discouraged.` I can think of two fixes, namely changing to a Window or propagating the parent to PDFDoc and from there to the dialog. Since neither is elegant, I am ok with leaving the warning for now.

jeromerobert:
@angsch Good point. I fixed both issues. Thank you.

Issue #341, opened 2020-12-04
Memory is not always released


kbengs:
Here is two possible ways to see this problem:

Way 1:
* Open a pdf and zoom to full page (to get high memory usage)
* Select all + delete all -> memory usage stays high
* Load another pdf -> memory usage is increasing

Way 2:
* Open a pdf and zoom to full page
* Select all + rotate 90 degrees
* Zoom back to small thumbnails -> memory usage is still high

jeromerobert:
Tagged 1.7.0 because should be at least investigated before releasing 1.7.0

kbengs:
It is a lot related to undo manager. Undo manager should probably not store the thumbnails, but instead render them when undoing.
But this is not the only explanation. A lot of memory is still not released when doing for example Select all + delete.

jeromerobert:
Clearing milestone has undo manager is fixed. Second part of the issue will be managed later.

kbengs:
Found this discussion:
https://stackoverflow.com/questions/51938963/python-memory-not-being-released-on-linux

Last post suggest one should call glibc:s malloc_trim to release memory. Tried to call it after pages are deleted. And yes memory is then released !
I will investigate a bit more and maybe come with a PR later.

```
import ctypes

def malloc_trim():
    ctypes.CDLL('libc.so.6').malloc_trim(0)

malloc_trim()
```

dreua:
Sounds great! Will it be portable to Windows?

kbengs:
Tested Linux vs Windows memory usage with patch applied in Linux and not in Windows. I did remove pages, rotate pages, zoom in/out and noticed how memory usage changed. The conclusion is that Windows don't need the patch, it works equally well without it. (Except that memory usage is 30-40 Mb more on Windows all the time.) So we could just: ``if os.name == 'posix'``

kbengs:
After select all + delete and malloc_trim memory usage goes down to 40-50Mb. But with one particular file it goes down to around 160Mb. So it seems like we have a third case also.
The file is 10Mb, 461 pages, in A3 format. Memory usage before "select all" + delete around: 1Gb. After: 160Mb.

If I scale the pages to A4, save and then try the same thing with this file, then memory will go down to 40-50Mb.

dreua:
[...]

PR #355, opened 2020-12-25
Use less ram by not storing thumbnails in undomanager

kbengs:
Ram can be saved by not storing thumbnails in undomanager. In a side by side comparison with before and after this commit, (zooming in, doing changes and zooming out..) the latter usually use much less ram.
Ram is not always released when I expect it should, but zooming in/out and doing changes, ram can suddenly be released.

To my understanding "r.thumbnail = None" should be enough, but "del r.thumbnail" seems to make it more likely that ram is released.

This commit is still not (usually) successfull at the select all + delete case.

Related to #341

### DIFF
--- a/pdfarranger/core.py
+++ b/pdfarranger/core.py
@@ -103,10 +103,13 @@ class Page:
         ts = [self.filename, self.npage, self.angle, self.scale] + list(self.crop)
         return "\n".join([str(v) for v in ts])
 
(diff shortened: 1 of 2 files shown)


kbengs:
In Windows 7 & 10 select all + delete seems to free most of memory (after this commit). Overall memory freeing seems to work much better in Windows.

Issue #306, opened 2020-10-25
Have "Split Page" support n x m split instead of just 1 x 2


ole-tange:
It is already possible to make a poor-man's poster with the program:

Edit > Split pages, followed by Rotate, followed by Edit > Split pages, followed by Rotate.

This way you get a 2x2 poster.

It is, however, not ideal: The overlap is minimal (around 1 mm on my printer).

But with very few changes this could be improved.

I hereby propose a menu for generating a poster. In the initial version it is fine, if you just do 2x2, 3x3 and 4x4. Those should be very easy to implement given the above tools. The poster should add a bit of overlap (say 3-5%), so it is easier to cut off a margin and make it fit. Therefore the pages should be scaled 1.9 (not 2), 2.85 (not 3), 3.8 (not 4) or similar.

For people using ISO-paper sizes (A4, A3, A2) it should also be easy to make a poster to 1x2 (Rotate, scale 1.4), 2x4 (Rotate, scale 2.8).

boythere:
Very tricky!

After splitting, the size of pages can be re-cropped. 

For example,
L-R-T-B: 48% 48% 0% 0%

angsch:
@ole-tange I don't think that pdfarranger will ever be the tool of choice for creating high end posters for printing. Fixing the margins for print for all sorts of formats and page tessellations is not entirely trivial. For the tasks that you give as examples, you may want to have a look at a specialized tool such as [pdfposter](https://pdfposter.readthedocs.io/en/stable/index.html).

boythere:
@angsch 

1. I'll be glad to see that the pdfposter-like feature would be delivered in new pdfarranger someday.

2. Besides a poster, we always need a booklet as the output type so that lots of people just know clicking `Print` only.

XanderLeaDaren:
By booklet, do you mean:

n, 1, 2, n-1,
n-2, 3, 4, n-3,
n-4, 5, 6, n-5,
…

so that you may fold all pages and staple them?

If so, I would support such an idea and we should create a feature issue for this. 😃

boythere:
@XanderLeaDaren Yes, we like it.

ole-tange:
[...]

PR #356, opened 2020-12-25
Add grid split

angsch:
Generalization of the page split feature as discussed in #306


jeromerobert:
Nice work. Some bugs:

* The *Cancel* button has the same behavior as the *Ok* button
* Both frame contains *Columns* and *Vertical splits*
* Undo leave an additionnal duplicate page

angsch:
I fixed the Cancel button and the erroneous labels. The problem with undo I cannot reproduce. Do you recall what you did to get the issue?

jeromerobert:
> The problem with undo I cannot reproduce.

Indeed, I was probably wrong, I cannot reproduce it.

PR #357, opened 2020-12-25
Add the "Insert Blank Page" action

### LINKED ISSUE #244: Add "Insert blank (empty) page" option
Hi.

This feature is highly needed by many users, for example to add notes within empty page.

I suggest the following, which suitable to current design of pdfarranger:
User when select a page (any page), then right click this selected page, the right click list appear & containing - in addition to current options - 2 new options: "add empty page before" & "add empty page after". The size of new empty page should be the same of size of selected page.

jeromerobert:
Close #244

### DIFF
--- a/data/menu.ui
+++ b/data/menu.ui
@@ -137,6 +137,10 @@ along with PDF-Arranger.  If not, see <http://www.gnu.org/licenses/>.
           <attribute name="action">win.paste</attribute>
           <attribute name="target" type="i">3</attribute>
         </item>
+        <item>
+          <attribute name="label" translatable="yes">Insert Blan_k Page</attribute>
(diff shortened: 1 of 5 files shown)


angsch:
Nice work. I like how you solved the issue with the page size.

When you create a blank page, the default directory in the file chooser dialog is modified. The next time you import a document, you start in `/tmp`. This also applies to the exporter. If it is not too difficult to fix, I would prefer to not alter the default directory set by the user.

kbengs:
A few more comments:
* The action is not in right-click menu, is that by purpose?
* Mnemonics accelerator is missing
* No newline at end of croputils.py

jeromerobert:
Issues fixed. Thank you.

PR #372, opened 2021-01-01
Disable fuzzy matching in translations

jeromerobert:
Fuzzy matching sometimes leads to nonsense. I removed some of them in 719455ce5060b8d593619f99c0318b1d5aa8ed13 using google translate. On the other hand, fuzzy matching sometimes lead to a not so bad translation, yet I think it's not safe enough to be used by default. What's your opinion ?

### DIFF
--- a/po/cleanuppo.sh
+++ b/po/cleanuppo.sh
@@ -23,6 +23,6 @@
 cd "$(dirname $0)/.."
 for po in po/*.po
 do
-  # Remove #~ lines
-  msgattrib --no-obsolete -o "$po" "$po"
+  # Remove #~ lines and fuzzy
+  msgattrib --no-fuzzy --no-obsolete -o "$po" "$po"
 done
--- a/po/updatepo.sh
+++ b/po/updatepo.sh
@@ -24,5 +24,5 @@
 cd "$(dirname $0)/.."
 for po in po/*.po
 do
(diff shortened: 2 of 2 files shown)


kbengs:
Yes I think you are right.

dreua:
I thought that fuzzy matching just provides suggestions to make the translator's live easier, did I get something wrong?
I assume:

- Fuzzy matched strings won't be displayed until a someone removes the fuzzy comment, until then the untranslated string is used.
- Fuzzy comments are only removed by a translator who verified the correct translation. (In that case it saves work and may help with consistency.)

(I've seen those wrong matches like Right=Height and worse. Totally agree that they must not be used without verification. On the other hand it would be great if fuzzy matched strings could be used in cases where the original string was just slightly modified without changing it's meaning, like fixing a typo. That however would require the tools to have some understanding of the language and I don't suppose the regular/free tooling has that.)

jeromerobert:
@dreua I just checked and you are right ! There is a `-f` flag in `msgfmt` but we (actually distutils-extra) are not using it. So this PR make no sens, I close it.
@angsch When you updated the `es.po` file, you left some fuzzy tags. I guess that was not intentional.

dreua:
Yeah those fuzzy tags are easy to forget, was wondering why my translations weren't showing up in app during the last translation session :see_no_evil:

Issue #291, opened 2020-10-01
Problems with editing if not saving after each operation


msdobrescu:
Hello, first, it is a great tool! It saves my time! a lot, I use it every day for my boy at school! It's brilliant! A must have!

I edit files scanned by my copier that is able to generate them directly, but I must rotate, delete, interleave pages etc..
During these operations I have noticed that can't be copied properly from one instance to another if the source was not saved first. Regardless the state, if pages are marked and copied from a source file that was edited, they are sent to clipboard in the state they are in the file on the disk, not in memory (as being edited). As effect, pasting them to a new document will bring only the pages existing in the source, in their state in the file on disk. This is misleading. Please copy the pages as they are in memory, in the edited source file or clarify the behaviour.
Same for cutting.

I use version 1.6.2 on Sabayon Linux.

Operating System: Gentoo Linux
KDE Plasma Version: 5.19.5
KDE Frameworks Version: 5.74.0
Qt Version: 5.15.0
Kernel Version: 5.7.0-sabayon
OS Type: 64-bit
Processors: 8 × Intel® Core™ i7 CPU 920 @ 2.67GHz
Memory: 23,5 GiB of RAM
Graphics Processor: GeForce GTX 1060 6GB/PCIe/SSE2

Thank you.

kbengs:
That was interesting. It is something that I can't reproduce. Does drag and drop to another instance work as it should? Would you be able to test run the application from a git clone:

git clone https://github.com/pdfarranger/pdfarranger.git
cd pdfarranger
pip install -e .
./setup.py build
python3 -m pdfarranger

Then we could add some debug printing to the code.

kbengs:
I installed Sabayon and did a git clone. But I couldn't see any problem when running pdfarranger this way. Also if I go back to the stage when 1.6.2 was released it works fine: if I rotate a page and copy-paste it to another instance it will be rotated also on the other instance.
@msdobrescu How did you install pdfarranger?

msdobrescu:
Hi, sorry for the delay, I was busy, I won't let you down, but I need time.
I was installing it through the regular package manager.
I'll test some more, maybe there is something related to specific PDFs.

kbengs:
No problem. I was also curious about what Sabayon is, so I tested it. :-)

msdobrescu:
I love Sabayon, I prefer it over all distros. I love Mageia too. Ubuntu is kind of messy in my experience.
[...]

PR #376, opened 2021-01-03
Hide the file name of blank pages

jeromerobert:
Then blank page file name is always a temporary file which is meaningless to the user. We set it to None so it's not displayed.

This currently break copy/paste across instances. The problem is related to #291 so let's fix it before concidering this PR.


jeromerobert:
Now much simpler thanks to #377

kbengs:
Should be ok

PR #377, opened 2021-01-05
Copy pages from temp files when using copy-paste & dnd

kbengs:
It should be working now.


kbengs:
And as usual I still find something.. The filename in password dialog is from the tempfile. So don't look too much on this yet.

kbengs:
Okay ready for review.

PR #379, opened 2021-01-08
Fix slow paste-interleave-odd/even

kbengs:
If I paste-interleave-odd a 17MB pdf with 435 pages it takes over 100 seconds before rendering begins. 
After this commit rendering begins after 1 - 2 seconds.


jeromerobert:
Nice

PR #392, opened 2021-01-16
Fix combined rotation and scaling

### LINKED ISSUE #387: Scaling, saving: loss of orientation in some cases
### Steps

1. Take a PDF with a landscape page
2. scale the page to 50%
3. observe the scaled page, still landscape
4. save as …

### Expected

5. the page, landscape

### Actual result

5. the page, portrait

jeromerobert:
Fix #387


angsch (commented):
Looks good to me

Issue #386, opened 2021-01-14
Behavior of crop white borders on rotated pages


angsch:
For my landscape [test.pdf](https://github.com/pdfarranger/pdfarranger/files/5815996/test.pdf):

- Rotate to the right and crop white borders. Page format dialog shows `left: 0, right: 0.1, top: 0, bottom: 0.2`
- Rotate to the left. Page format dialog shows the correctly translated crops `left: 0, right: 0.2, top: 0.1, bottom: 0.0`
- Crop white borders again. `left: 0, right: 0.1, top 0, bottom: 0.2`

Note that the right crop and the top crop has shrunk. The crop-white-borders feature apparently resets the crops and computes fresh ones. Is this intended? I expected that the second cropping is applied to the already cropped page. In other words, crops can only increase, but not decrease.

kbengs:
Crop white borders after page rotate for sure has some bug. It can be seen clearly on this pdf.
[pdf-sample.pdf](https://github.com/pdfarranger/pdfarranger/files/5818241/pdf-sample.pdf)

[result.pdf](https://github.com/pdfarranger/pdfarranger/files/5818256/result.pdf)

dreua:
I see two issues here:
1) Crop white borders is not aware of page rotation. It calculates the values based on the not rotated page and applies them on the rotated page which does not make sense. (The Crop percentage values are always the same, no matter how you rotated the page. Also note that applying the percent values to a 90/270° rotated page results in total nonsense because of the changed width/height.)
2) Crop white borders ignores previous crop and starts on the whole page.

I'd say 1 is absolutely a bug that should be fixed while 2 is a niche use case that must not necessarily be fixed and would also need another file to reproduce after 1 is fixed. (Maybe cut marks or a dark border that you crop off by hand and then apply automatic cropping to the remaining page?)

General note to keep in mind when working on rotation or page dimensions, not sure if it is important in this issue or not: PDF pages can be already saved with a rotation in the PDF, e.g. when you rotate a page in pdfarranger or another application, save it and then load that file. It would be advisable to test fixes with a PDF that contains both, previously rotated and unrotated pages to make sure we fix this once and for all.

jeromerobert:
I would also not concider (2) as a bug. (1) is similar to #387 but I guess both will need their own fix.

jeromerobert:
[...]

PR #393, opened 2021-01-17
Fix "crop white border" on rotated pages

jeromerobert:
* Related to #386
* Only fix the issue on rotated pages
* Cropped pages are still concidered are not cropped. This will be fixed later.


kbengs (commented):
Looks good and seems to do the job.

PR #394, opened 2021-01-20
Update German and Croatian translations

milotype:


po/de.po
```diff
@@ -424,92 +422,88 @@ msgstr "Rei_henfolge umkehren"
...
+#: data/menu.ui.h:23
 msgid "Paste _Before"
-msgstr "Einfügen _vor Auswahl"
+msgstr "Einfügen _vor"
 
-#: data/menu.ui.h:25
+#: data/menu.ui.h:24
 msgid "Paste Interleave _Odd"
-msgstr "Abwechselnd einfügen _ungerade"
+msgstr "Leere _ungerade Seite einfügen"
```
dreua, line 445: "Abwechselnd einfügen" is not great, but I think you misunderstood the meaning here: We are not inserting empty pages but pasting from clipboard in a way that alternates an existing page and a pasted page just like the teeth in a zipper.

dreua (changes requested):
Hey, thank you for the improvements and fixes to the German translations! (Most of those were by me and I realize now, that being mother tongue doesn't imply good translations.)

The only point where I disagree is "paste interleaved", would you have another idea for that by any chance?

Can't comment on Croatian translations but I'm sure they are good. PR should be merged as soon as "paste interleaved" in German is fixed.

kbengs:
For the swedish translation I have used "Paste as odd pages" and "Paste as even pages" maybe that could be a idea. I don't know if "Paste Interleave Odd/Even" is good English either.

milotype:
Sorry, if I misinterprerated ... consulting merriam-webster, this was my assumption ...
```
interleaf noun - in·​ter·​leaf
Definition of interleaf (Entry 2 of 2)
1 : a usually blank leaf inserted or fastened between two leaves of a book (as for written notes or for protecting a color plate)

```

If you agree with @kbengs, I could change german and croatian translations. Would it be plural or singular? Are you pasting from clipboard (a single page?) multiple times? Sorry, but I'm a bit confused here.

PR #395, opened 2021-01-22
Update Czech translation

pavelzahradnik:


po/cs.po
```diff
@@ -93,7 +93,7 @@ msgstr "Oříznout okraje"
 #: pdfarranger/croputils.py:157
 msgid ""
 "Cropping does not remove any content from the PDF file, it only hides it."
-msgstr ""
+msgstr "Oříznutí neodstraní žádný obsah z PDF dokumentu, pouze jej skryje."
 
 #: pdfarranger/croputils.py:166
 #, fuzzy, python-format
```
dreua, line 99: Translations with "fuzzy" comment will not be used, is this intentional?

pavelzahradnik: No, sorry, it should be deleted, all occurrences.
```suggestion

```

dreua:
Thank you!

Issue #390, opened 2021-01-15
MIME: allow PDF Arranger to be recognised as a handler for all types of file that it can handle


grahamperrin:
Building upon https://github.com/pdfarranger/pdfarranger/pull/202

### Enhancement request

Please make the application recognisable for the purpose of opening PNG, JPEG and other types of file that it can handle. 

### Workarounds

With KDE Plasma 5 (on FreeBSD-CURRENT) I can remember an association, for example: 

![2021-01-15 05:14:46](https://user-images.githubusercontent.com/192271/104686242-b2667480-56f4-11eb-98b1-21956ff9ec20.png)

– however (unless I'm missing something) the remembrance can **not** be made without also setting the app as a default. 

So each such remembered association must be followed by demoting PDF Arranger, or (as shown below) promoting what was previously my preferred application: 

![2021-01-15 05:19:03](https://user-images.githubusercontent.com/192271/104686331-e5106d00-56f4-11eb-8788-3c2f843e5fb7.png)

dreua:
I previously thought about this and wasn't sure we should do it since pdfarranger is not really a good program to open images in general, it's useful only for the special task of creating PDFs from images. On the other hand I don't see a big argument against it and since it is now requested we can just do it. Any thoughts from the other maintainers?

The implementation should be easily done by adding the correct mime-types in the desktop file. (PNG, JPEG/JPG, anything else?)

@grahamperrin Just to double-check, can you quickly confirm that this is only about the images and pdfarranger appears for opening PDFs by default (like it should)?

grahamperrin:
The application is _not_ recognised as a recommended application by Firefox on (Tier-3) FreeBSD-CURRENT: 

![image](https://user-images.githubusercontent.com/192271/104797009-cd97b980-57b2-11eb-873a-9274cb19c280.png)

– however I guess that a fix for this (if any) might be out of scope for https://github.com/pdfarranger/pdfarranger

It will be more meaningful for me to check with a cleaner installation of a desktop environment, with (say) Dolphin instead of Firefox. I'll do so as soon as I can.

dreua:
@grahamperrin Let's try to stay on topic and with reproducible issues here, I'll hide your last comments since it's not adding value to this and the linked issue in my opinion.

dreua:
For reference on MIME types here are excerpts from desktop files on my system:

Gimp:
[...]

PR #413, opened 2021-02-07
Add image formats as supported mime types

dreua:
Closes: https://github.com/pdfarranger/pdfarranger/issues/390


dreua:
I quickly tested with png and jpg, both work without issues.
You can use this flatpak build for testing the open with, but it is also built without --filesystem=host so don't be surprised if drag and drop and clipboard don't work.

Issue #14, opened 2018-09-29
Tag a new release?


paride:
The port to gtk3/python3 is a huge improvement for this great tool, it is really worth a release. Many distributions (e.g. Debian) normally package only "official" releases, so the new features won't make it in the next release if a version is not tagged.

jeromerobert:
See #9

I just e-mailed the original to known what are is plans

jeromerobert:
Duplicate of #9 (which I just reopened)

PR #416, opened 2021-02-11
Upload RPMs

dreua:


dreua:
Worked on the first try. :tada:  You can now download the RPM build results as ZIP. (And yes, unfortunately only as ZIP, see: [Display files in browser where appropriate. · Issue #14 · actions/upload-artifact](https://github.com/actions/upload-artifact/issues/14))

I just wanted to test this and I'm not sure whether this should be merged at all, because the RPM works only on Fedora 33, so that is not even include me since I'm still on 32 currently. On the other hand it's free and may come in handy for testing once I upgrade :shrug: 

@jeromerobert Should we give fully automated Windows builds another try? I think an exe build per PR would be useful for testing.

jeromerobert:
> Should we give fully automated Windows builds another try? I think an exe build per PR would be useful for testing.

Yes ! Well I would prefere to have it for `main` (only or also ?). This way we could have a link in the wiki or even in the `README.md`. I think very few people will test binaries from PR but a bit more would test `main`.

PR #421, opened 2021-02-14
Normalize spelling: PDF Arranger

dreua:
We had at least three ways of spelling "pdfarranger", "PDF Arranger" or "PDF-Arranger".

This PR tries to normalize that in just two forms: PDF Arranger as the official name in documentation and all UI strings and pdfarranger as package name, filename, in various parts of the source code and source code comments.

This is just a suggestion based on the current usage (PDF Arranger is used in the most important places like the window title and in the desktop file.) I can easily replace PDF Arranger by PDF-Arranger or go pdfarranger all the way if that would be preferred.


dreua:
I'd wait for with this for 1.8 just in case it breaks anything.

kbengs:
I think that is good.

dreua:
Just rebased this on main and resolved a merge conflict I introduced myself. No need to re-review, I'm pretty sure I got this right.

Is the change in general appreciated or would you rather leave it like it is or change it all to another spelling (@jeromerobert )?

jeromerobert:
Looks good to me. You can merge.

Issue #327, opened 2020-11-15
Scaling and moving for poster


ole-tange:
In #306:

> IMHO the second one is currently not clear nor well specified and will need the maturation of the existing scaling/croping feature.

I do not think you can separate the process into two parts without making it overly complicated.

Let me try to specify the process of making an n x n poster. I will focus on on the x axis, but y axis will be similar.

```
// A border of 1cm will work for all printers I have used in my life - thus this does not need to be configurable
border = 1cm
// Make a 1cm border used for overlap (and non-printable areas) in both sides
page_center_width = page_width - 2*border
// generate n x n poster
// the resulting poster (after printing and putting it together with overlaps) will measure poster_width
poster_width = border + page_center_width*n + border
// the original page should be scaled by this much
scale_factor = page_center_width * n / page_width 
poster = original_page * scale_factor
// move the poster so that the original x=0 is moved to start of the page_center area
page 1 = move(poster,border - page_center_width*0)
// move the poster so that start of the page_center area will be the end of the page_center area of page 1
page 2 = move(poster,border - page_center_width*1)
// move the poster so that start of the page_center area will be the end of the page_center area of page n-1
page n = move(poster,border - page_center_width*(n-1))
```
The corresponding calculations should be done for the y values.

This should be enough for converting a single page into an n x n poster - no matter the aspect ratio of the paper.

To make a n x 2n poster (portrait to landscape) is similar (assuming page_height = sqrt(2)*page_width as it is for ISO 216 paper sizes):

```
// A border of 1cm will work for all printers I have used in my life - thus this can be fixed
border = 1cm
// Make a 1cm border used for overlap (and non-printable areas) in both sides
page_center_width = page_height - 2*border
// generate n x 2n poster
// the resulting poster (after printing and putting it together with overlaps) will measure poster_width
poster_width = border + page_center_width*n + border
// the original page should be scaled by this much
scale_factor = page_center_width * n / page_height
poster = original_page * scale_factor
// move the poster so that the original x=0 is moved to start of the page_center
page 1 = rotate(90,move(poster,border - page_center_width*0))
// move the poster so that start of the page_center will be the end of the page_center of page 1
page 2 = rotate(90,move(poster,border - page_center_width*1))
// move the poster so that start of the page_center will be the end of the page_center of page n-1
page n = rotate(90,move(poster,border - page_center_width*(n-1)))
```

Similar calculations for y values. The widest direction will not go to page n, but to page 2n.

[...]

PR #422, opened 2021-02-14
Make 'Split page' support posters

ole-tange:
Closes: https://github.com/pdfarranger/pdfarranger/issues/327


angsch (commented):
On testing, I found a few problematic cases.

### Zero-dimensional pages

Pages where either the column width or the row height is zero will create a thumbnail that can be exported and yields an illegally sized page.

Reproduce
- Split a page
- Untick equal column width
- Increment vertical splits

Please add a check so that we do not get empty pages.

### Uneven splits

If you set the sizes manually, the splits are evenly sized, regardless of whether or not the checkbox equal column width/row height is ticked.

Uneven splits, say 33% + 67%, are still supported, but it is more difficult. The following works:
- Untick equal column width
- Increment vertical splits
- Tick equal column width
- Untick equal column width
- Decrement vertical splits

I am ok with not supporting uneven splits, but then I suggest to remove the checkbox equal column width/row height and keep equal sizes all the time.

### Content loss possible

The targeted content overlap implicitly assumes that the sizes add up to at least 100 %. If this is not the case, the center parts of the page are lost.

Reproduce
- Split a page
- Overwrite col 1 width with 30%. This will automatically adjust the width of the second column to 30%.

The result is two pages showing the leftmost 30% and the rightmost 30%. The center part is lost. I suggest to add a check that `crop_sum` is at least 100.

### Other comments
Did you possibly forget to adjust the comment in line 144? The calculation for the size of the i-th slice is `crops[2*i+1] - crops[2*i]`.

dreua:
I just realized that "dreua added this to the 1.8.0 milestone" could be misunderstood: All I actually wanted to say is, that this should not go into 1.7.x, but that's quite obvious. It's a nice to have feature but not a must have for 1.8 therefore I'll remove the milestone again.

angsch:
That conflict is ... unexpected. Does someone know how to exclude that file from the PR so that we do not get a merge commit?

jeromerobert:
@angsch `git rebase -i`, edit the culprit commit, unstage the `.pot` file, `push -f` the branch again.

jeromerobert:
@angsch I did it and pushed https://github.com/pdfarranger/pdfarranger/tree/poster

angsch:
@jeromerobert  Thank you!

The changes adjust ole-tange's version such that overlap is only possible if the splits have equal size. The overlap is then evenly distributed. I do not see a way to generalise overlap to uneven splits.

It was quite a bit of rewrite and if any of you have time to test, that would be great.

PR #425, opened 2021-02-22
Release memory after rendering & deleting pages

kbengs:
As mentioned in #341 there are at least one more "memory leak" case. So I'd say the issue should be left open even if this is merged.

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -26,7 +26,9 @@ import locale  # for multilanguage support
 import gettext
 import gc
 import subprocess
+import ctypes
 from urllib.request import url2pathname
+from functools import lru_cache
 
 
 sharedir = os.path.join(sys.prefix, 'share')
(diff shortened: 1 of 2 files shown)


pdfarranger/pdfarranger.py
```diff
@@ -27,6 +27,8 @@
 import gc
 import subprocess
 from urllib.request import url2pathname
+if os.name == 'posix':
+    from ctypes import CDLL as libloader
```
jeromerobert, line 31: I would just import ctypes here (i.e not test os.name and not rename CDLL to something else)

pdfarranger/pdfarranger.py
```diff
@@ -163,6 +165,12 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def malloc_trim():
+    """Release free memory from the heap."""
+    if os.name == 'posix':
```
jeromerobert, line 170: This may fail on posix system like BSD. It would be more portable to try and catch `AttributeError` and `OSError`. Also may be it would be better to try only once and cache the `malloc_trim` function on success.

jeromerobert (changes requested):

pdfarranger/pdfarranger.py
dreua: Sound like a good use for singleton pattern, just return a no-op function on error.

kbengs:
Okay, I changed according to Jerome's suggestions.

Issue #427, opened 2021-02-26
[FIXED] TypeError: only pages can be inserted with libqpdf >= 10.2.0 **AND**  pikepdf < 2.7.0


sillysimon:
*Moderation Edit:*
in most cases the answer is `pip install --upgrade pikepdf`
Update: 1.7.1 includes a workaround. Upgrade what ever is more convenient for you.
/@dreua

---------

Hi there,
I have a problem with the arch software package version of the software. Every time I try to save a modified document it does not do so but instead throws the following error:
```python
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/pdfarranger/pdfarranger.py", line 695, in choose_export_pdf_name
    self.save(mode, file_out)
  File "/usr/lib/python3.9/site-packages/pdfarranger/pdfarranger.py", line 157, in wrapper
    func(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/pdfarranger/pdfarranger.py", line 769, in save
    exporter.export(self.pdfqueue, to_export, file_out, mode, m)
  File "/usr/lib/python3.9/site-packages/pdfarranger/exporter.py", line 168, in export
    pdf_output.pages.append(new_page)
TypeError: only pages can be inserted
```
if I use the flatpak version it works fine. Although I would really prefer to use the native arch package

jeromerobert:
Could apply the following changes to help finding the bug:
```patch
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -97,6 +97,7 @@ def _scale(doc, page, factor):
         Resources={'/XObject': {'/p{}'.format(page_id): xobject}},
         Rotate=rotate,
     )
+    print("end of _scale", type(new_page), new_page)
     return new_page
 
 def check_content(parent, pdf_list):
@@ -150,6 +151,7 @@ def export(input_files, pages, file_out, mode, mdata):
         try:
             new_page = copy.copy(new_page)
         except TypeError:
+            traceback.print_exc()
             if _report_pikepdf_err:
                 _report_pikepdf_err = False
                 traceback.print_exc()
@@ -157,6 +159,7 @@ def export(input_files, pages, file_out, mode, mdata):
                       "1.7.0 or greater. Continuing but pdfarranger will not "
                       "work properly.".format(pikepdf.__version__),
                       file=sys.stderr)
+        print("new_page copy", type(new_page), new_page)
         if angle != 0:
             new_page.Rotate = angle + angle0
[...]

PR #435, opened 2021-03-08
Add workaround for pikepdf < 2.6.0

dreua:
See: https://github.com/pikepdf/pikepdf/issues/174
See: https://github.com/pdfarranger/pdfarranger/issues/427

(Also removed the walrus operator for compatibility with older python versions.)

How do you feel about pushing this out as a bugfix release in the next days?

### DIFF
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -165,12 +165,20 @@ def export(input_files, pages, file_out, mode, mdata):
         if cropped:
             new_page.MediaBox = cropped
         new_page = _scale(pdf_output, new_page, row.scale)
+
+        # Workraround for pikepdf < 2.7.0
+        # https://github.com/pikepdf/pikepdf/issues/174
+        new_page = pdf_output.make_indirect(new_page)
+
         pdf_output.pages.append(new_page)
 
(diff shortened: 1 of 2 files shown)


dreua:
Tested save and export with: 
```
$ python3 -c 'import pikepdf; print("pikepdf:", pikepdf.__version__); print("libqpdf:", pikepdf.__libqpdf_version__)'
pikepdf: 1.19.4
libqpdf: 10.2.0
```

dreua:
Also works with pip install which pulled `pikepdf-2.8.0.post1`
@pdfarranger can someone have a look? I'd like to push this workaround to our users, so that qpdf updates are no longer a problem.

vivi90 (approved):

vivi90 (approved):
@dreua Good idea

vivi90:
@jeromerobert Yes, would be great.

kbengs:
I see no issue with it. Tested with pikepdf 2.6 + libqpdf 10.2.0 and pikepdf 2.8.0.post1+ libqpdf 10.2.0.
Malloc trim also work for me.
Commit message should be < 2.7.0 ..

dreua:
> Commit message should be < 2.7.0 ..

Dang :/

I think now it's too late. I could revert and add a new commit, but is that worth it?

PR #436, opened 2021-03-14
Render more sharp thumbnails (in some cases)

kbengs:
* Set cell size correctly so that thumbnail is not (in some cases)
  scaled.
* With GObject.TYPE_FLOAT the resample value emitted from rendering
  thread will have changed a bit when it is received in update_
  thumbnail. With GObject.TYPE_PYOBJECT it is transferred unaltered.
* The issue that cells are (sometimes) moving up or down while
  rendering is also eliminated here.

Here are two examples. Before commit to the left and after to the right. I see same result with many pdfs.

unsharp-sharp

picture2

### DIFF
--- a/pdfarranger/core.py
+++ b/pdfarranger/core.py
@@ -441,11 +441,10 @@ class PDFRenderer(threading.Thread, GObject.GObject):
         w, h = page.get_size()
         scale = p.scale / self.resample
         thumbnail = cairo.ImageSurface(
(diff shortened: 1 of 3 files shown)


angsch:
The code looks good. For some PDFs the improvement is remarkable.

kbengs:
Thanks for having a look. I see that do_get_size() could use Page.width_/height_in_pixel() instead. I'll fix that.

kbengs:
Now thumbnails should be sharp for any page size, scale, rotation and crop

Issue #246, opened 2020-07-24
Allow to do page tiling


Nokia808:
Hi. This option is already possible by QPDF !

User should be able to enter the following values:

- raw number,
- column number,
- spacing (in cm/mm ...)

If you accept to add this feature, then, kindly do not forget to add sub-option to support right to left arrangement which needed for Arabic & Hebrew PDF when selecting a raw number higher than 1 ....

jeromerobert:
> this option is already possible by QPDF

How ?

Nokia808:
@jeromerobert 
Hi dear. There is an application called "PDF Mix Tool" already do this. Currently "change page layout" + "page alignment" + "margins" are all in one window. But in future developer of "PDF Mix Tool" will separate them, which is better. "PDF Mix Tool" is here:
https://gitlab.com/scarpetta/pdfmixtool

Currently, developer of "PDF Mix Tool" is working very hard to bring version 0.7 which may take few months from now to be completed ... I asked him to add features & he, kindly, added them & will add. I discover your nice application & see it very promising, so asked to add features that make it a more complete tool ...

Regarding "PDF Mix Tool", go to single file mode, then go to menu, &  try to edit any profile for "Multipages" & you will see these options - see screenshots:

![1](https://user-images.githubusercontent.com/25375330/88449852-ec3b7a00-ce39-11ea-8729-ce45512f95b0.png)
![2](https://user-images.githubusercontent.com/25375330/88449855-f2315b00-ce39-11ea-91cc-e5aa6788292e.png)
![3](https://user-images.githubusercontent.com/25375330/88449860-f6f60f00-ce39-11ea-9b6c-af1acccd5cd3.png)

jeromerobert:
Ok so this is about porting a pdfmixtool feature to pdfarranger. Really not an emergency.

jeromerobert:
The QPDF code for this feature is here:
https://gitlab.com/scarpetta/pdfmixtool/-/blob/b8a9bf3c/src/pdf_edit_lib/pdf_writer.cpp#L223-393

angsch:
Dear fellow developers,

I have the tiling working with pikepdf now and next is the integration into pdfarranger. I see two options.
* We go for an in-memory rendering solution similar to the page scaling feature. Only on export, the pages are physically glued together. This requires the class  `Page` to support multiple source documents.
* We go for a temporary file similar to the blank page feature. Programmatically this should be the easier solution.

What are your thoughts?

[...]

PR #440, opened 2021-03-23
Tile two pages with rotations

angsch:
**First proposal towards page tiling #246.**

- Exactly two pages can be glued together vertically.
- Pages can have any rotation.
- The functionality aims at being a basic building block to support more complex tiling operations like for instance, horizontal tiling, m x n posters, booklets,... .

**Justification of the design decision:**
We now have to rotate the content (not just the page orientation). To compensate, we have to shift the content back into the first quadrant of the Cartesian coordinate system. Rotations work recursively. Hence, it is a challenge to tile more than two arbitrarily rotated pages in a single step. Therefore the functionality is limited to two pages and the (hopefully correct) offsets are hardcoded. I consider this easier to maintain than solving equations to get the shifts right. This is not a limitation because it is possible to apply tiling repeatedly.

**Known issues:**
- Crop is not supported. Crop for the outside frame of a glued page can be added. The problematic part is the center. Example: We split A -> [A1, A2] and B -> [B1, B2] and want to glue together [A1, B2]. Since cropping only hides the content, we have to find a way to properly chop off A2 and B1.
- Scaling is not supported.
- Error handling is missing. It is possible to tile pages whose height does not match.

**Input needed:**
- I am not sure how to best handle crops. The ugly solution is to generate a temporary cropped file and use that. It would be nice to have tiling work as an "undo split" operation.
- What should happen when you apply tiling to a general selection? Tile every second page ("undo split")? Glue everything together (what if not all pages have the same height?)

### DIFF
--- a/data/menu.ui
+++ b/data/menu.ui
@@ -110,6 +110,10 @@ along with PDF Arranger.  If not, see <http://www.gnu.org/licenses/>.
           <attribute name="action">win.split</attribute>
         </item>
         <item>
+          <attribute name="label" translatable="yes">_Stitch Pages</attribute>
+          <attribute name="action">win.stitch</attribute>
+        </item>
+        <item>
           <attribute name="label" translatable="yes">Cu_t</attribute>
           <attribute name="action">win.cut</attribute>
         </item>
@@ -326,6 +330,10 @@ along with PDF Arranger.  If not, see <http://www.gnu.org/licenses/>.
       <item>
         <attribute name="label" translatable="yes">_Split Pages</attribute>
         <attribute name="action">win.split</attribute>
+      </item>
+      <item>
+        <attribute name="label" translatable="yes">_Stitch Pages</attribute>
+        <attribute name="action">win.stitch</attribute>
       </item>
       <item>
         <attribute name="label" translatable="yes">Insert Blan_k Page</attribute>
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -67,6 +67,157 @@ def _mediabox(page, crop):
     return [x1_new, y1_new, x2_new, y2_new]
 
 
+def _rshift(lrotate, rrotate, lsize, rsize):
+    wl, hl = lsize
+    wr, hr = rsize
+    shift = {
+    # lrotate : rrotate : offset of right page
+        0: { 0 : [wl, 0],
+             1 : [wl, hr],
+             2 : [wl + wr, hr],
+             3 : [wl + wr, 0]},
+        1: { 0 : [wr, wl],
+             1 : [0, wl],
+             2 : [0, hr + wl],
+             3 : [hl, wl + wr]},
+        2: { 0 : [0, hr],
+             1 : [0, -wl + hr],
+             2 : [-wl, 0],
+             3 : [-wl + wr, hr]},
+        3: { 0 : [0, -hl + wr],
+             1 : [hr, 0],
+             2 : [hl, -wl],
+             3 : [0, -wl]}
+    }
+    return shift[lrotate][rrotate]
+
+def _lshift(lrotate, lsize):
+    w, h = lsize
+    shift = {
+        0 : [0, 0],
+        1 : [0, h],
+        2 : [w, h],
+        3 : [w, 0]
+    }
+    return shift[lrotate]
+
+def _create_tmp_page(tmpdir, page, angle, crop, scale):
+    rotation_matrices = {0 : [1, 0, 0, 1],      # 0 degrees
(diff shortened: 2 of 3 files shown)


angsch:
It is about putting the content of two pages onto a single page.

dreua:
> It is about putting the content of two pages onto a single page.

I see, if you think about it this way, tiling makes sense. Maybe "joining" leaves less room for misunderstandings but if I'm the only one confused here we may es well continue using "tiling". Thanks for the clarification!

Unfortunately I don't have time to dive into this new feature, I still have Windows CI and the Annotations bug on my todo list, but it sounds like a great addition to PDF Arranger :+1:

kbengs:
Looks like a good start. Using temporary files is not a problem imo. I have no good ideas how to handle crops. (rendering pages to images and tile the images, but that is probably not what we want)
> What should happen when you apply tiling to a general selection?

Maybe it could check that an even number of pages are selected and then tile them two by two. Or then it could be like it is now.

angsch:
- Feature is renamed to stitching
- Only pages with exactly the same height can be stitched
- Two pages with any crop, scaling or rotation can be stitched, i.e., placed next to each other onto one page.

image

>>   What should happen when you apply tiling to a general selection?

> Maybe it could check that an even number of pages are selected and then tile them two by two. Or then it could be like it is now.

Let's make that decision later. The current version only stitches 2 pages.

Sometimes pages that logically should be stitchable are not. 

1. Crop A and save as B.
2. Load B.
3. Load A and apply crop.

A and B should have exactly the same dimensions - unfortunately in practice they are not guaranteed to have this. I assume that we have to relax the criterion when pages are stitchable to allow for some rounding error. I consider this a minor point.

kbengs (commented):
I found one case that does not work as expected. To reproduce:
* Open a pdf and rotate all pages 90 degrees
* Export
* Open the exported file
* Stick two pages -> it does not look right

Some things I would still do:
* Delete the "source" pages
* Insert the new page after last selected page
* Maybe do a scroll_to_selection() to center the stitched page in window. (because cell size probably change after the  operation and the page could be lost out of sight otherwise)

angsch:
- Fixed the issue when the loaded document has rotated pages
- The stitched page replaces the selected two pages at the right spot

I am not happy with the integration into the undo manager. To restore the original two pages, you have to undo twice. Both the deletion of the original pages and the pageadder are a single action that is tracked. I don't know how exactly the undo manager works and currently do not see an easy way out, do you?

dreua:
@jeromerobert wrote the undo manager. I haven't looked into this specific problem but it sounds doable to me. Maybe understanding and modifying the undo manager is necessary.

angsch:
Should have rebased earlier - would not have encountered my problem with the undo manager. Great work again, kbengs

kbengs:
Nice. Found a few more things:
* If two pages that are not adjacent are stitched together the resulted page will not be selected
* Scroll_to_selection does not work for me. It does work with GObject.timeout_add(50, self.scroll_to_selection) (the function would need to be improved a bit so that it would work without the timeout)
* Annotations seems to be removed. That is maybe ok, but are they removed or just moved to a side, and could become visible sometimes?
* If the Linux Format page in # 443 is stitched the invisible margins will become visible
* I tried to do the stitch action repeatedly. 4 pages worked good, but with 4+4 pages it did not look right any more. (see pictures)

4 pages stitched is ok:
before1

But not 4+4:
after2

Davexx:
In case anyone is interested, the Fujitsu Scan Snap Organizer Viewer has a stitch feature they call Merge, with "Merge Vertically" and "Merge Horizontally" as the two options.  One has to select the two pages and right-click to see it in the context menu.
If the pages don't match, an error window pops up and says the paper size or page orientation do not match, prompting the user to check the size or orientation and try again.
If this feature does get worked on, I would suggest a warning message like the one mentioned above, but then ask the user if they would like to 1.) scale or crop the left page to match the right page,  or 2.) scale or crop the right page to match the left page.  I think this would make it way more convenient a feature.

I also support the idea of some sort of bulk stitching method, where a user could, say, select a contiguous group of even amount of pages and pdfarranger would execute the stitches on each pair from left to right.

I realize this is not easy to code, but it sure would make for a powerful feature and a nice compliment to the "Split Pages" feature.
Thanks for considering.

DegrangeM:
Hi, any progress on this ? I would love this feature <3

Being able to put two A4 paper into one A4 paper (by scaling and rotating) would be really usefull.
I saw the PR changed to stitching which make what I need a little harder to do, but stitching is usefull too.

kbengs:
This was implemented in #847

Issue #439, opened 2021-03-21
Enhancement to "About" dialog


m-holger:
Would it be possible to add something along the lines of

(PDF Arranger uses pikepdf vers. ??? and  libqpdf vers. ???)

to the "about" dialog? This information is relevant when reporting issues, and determining the versions (especially for libqpdf) is not all that intuitive.

Happy to have a stab at it myself if it helps.

dreua:
Yes! I've thought about this a few times but never came around to do it, please go ahead! I'm happy to help, just create a PR or tag me in a comment on your fork's commit and I'll be there.

The code to get pikepdf and libqpdf versions is quite simple: `python3 -c 'import pikepdf; print("pikepdf:", pikepdf.__version__); print("libqpdf:", pikepdf.__libqpdf_version__)'`

Just those two versions would be a great start, I'd say that's our [MVP](https://en.wikipedia.org/wiki/Minimum_viable_product).

If you are bored by those two versions, I already have Ideas for some additions on my mind:

- [ ] Include pdfarranger version
- [ ] Easy to copy and paste / copy button / create prefilled GH issue button (VS Code does this, probably just some URL parameters. I can have look if you want)
- [ ] Python file path:  `__file__` (Should tell us: Win/Linux, installed as admin or user, maybe even flatpak or not?)

m-holger:
#445 implements the basic requirement, i.e. report libqdf and pikepdf versions.

With regard to the enhancements suggested by @dreua, I would suggest that the priority should be 
to provide additional useful information, leaving the niceties such as pre-filled GH issue pages for later.
Assuming this is accepted, I require some guidance on the what and how.

# What further information do we want to display?

@dreua identified the Python file path: `__file__ `(Should tell us: Win/Linux, installed as admin or user, maybe even flatpak or not?)

I would assume that part of `sys.implementation` would also be useful: `_multiarch, name, version`?

``` python
>>> sys.implementation
namespace(_multiarch='x86_64-linux-gnu', cache_tag='cpython-38', hexversion=50857456, name='cpython', version=sys.version_info(major=3, minor=8, micro=5, releaselevel='final', serial=0))
```


Is there anything else?

# How do we want to display the information?

I see 3 main options:

## 1. Repurpose the credits panel

Change the button label from "Credits" to, say, "More Info", and add an additional section (after the creator and contributor sections)
to hold the additional diagnostic info and perhaps (a link to) some instructions on how to report issues.

Main advantages:

[...]

PR #441, opened 2021-03-23
Add libqpdf and pikepdf version numbers to About dialog

m-holger:
Minimal implementation of #439 

Adds the line 
```
(PDF Arranger uses libqpdf x.x.x and pikepdf x.x.x)
```
to the About Dialog . Also changes PDF Arranger website to a link.

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -27,6 +27,7 @@ import gettext
 import gc
 import subprocess
 import ctypes
+import pikepdf
 from urllib.request import url2pathname
 from functools import lru_cache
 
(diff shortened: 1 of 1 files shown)


dreua:
Thanks for the PR!

Please link to the corresponding GH issue so everyone involved can easily navigate between the two. (Ideally add the full link to the commit message for future reference. Use Closes: <link> to automatically close the issue once the commit is merged.)

Two minor nitpicks: The two commits should be squashed into one. You may use the force (push -f) on your feature branch, it should automatically update this PR. And commit messages should be present tense (Add instead of Added).

dreua:
You can fix those nitpicks as a git exercise or we can squash those before merging and leave the tense as it is, not that important imo. Your choice.

m-holger:
@dreua Sorry, I was trying to  make a push request to my own repository and therefore didn't provide a comment until after I realised that this was a "live" push request.

Will update as per your suggestions tomorrow.   

Reminder - needs checking that string change does not cause runtime errors on non-English versions

kbengs:
The code looks good to me. The commit message break the guidelines a bit: Subject line should not end with a dot. There should be a empty line between subject and body.

m-holger:
This modification unnecessarily breaks the translation of the existing AboutDialog message

dreua:
@m-holger What is the plan here?

m-holger:
@dreua I am testing a modified implementation that preserves existing translations and will issue a new PR when done.

PR #446, opened 2021-03-27
Fix thumbnail size during scale transition

kbengs:
When page format is changed the thumbnail will not have the right size
until the thumbnail is re-rendered. This commit changes resample so
that thumbnail is scaled until it is re-rendered.

### DIFF
--- a/pdfarranger/croputils.py
+++ b/pdfarranger/croputils.py
@@ -43,6 +43,7 @@ def scale(model, selection, factor):
         f = min(f, 14400 / page.size[0], 14400 / page.size[1])
         if page.scale != f:
             changed = True
+        page.resample = page.resample * f / page.scale
         page.scale = f
         model.set_value(it, 0, page)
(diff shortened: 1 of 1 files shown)


dreua:
I'm not sure I understand the issue here, can you give me the steps to reproduce? The thumbnails look fine to me after cropping or scaling.

kbengs:
Yep, the steps are: (in main)
* Open a pdf where it takes several seconds to render all pages
* Change scale factor for last page to 200%
* Notice how the tumbnail looks like while you wait for the thumbnail to be re-rendered (it's too small)

kbengs:
The problem with previous code was that if you change scale factor and while you wait for re-rendering of the thumbnail, you change scale factor again. Then the thumbnail would not have the right size.
But that case is fixed now also.

dreua:
Alright, I'll try to have another look today!

PR #450, opened 2021-03-28
Remember window position on screen

JATIN2111999:
Created function for Window position restore PR

### DIFF
--- a/pdfarranger/config.py
+++ b/pdfarranger/config.py
@@ -111,6 +111,13 @@ class Config(object):
         self.data.set('window', 'width', str(size[0]))
         self.data.set('window', 'height', str(size[1]))
 
+    def set_position(self,position):
+        self.data.set('window', 'root_x', str(position[0]))
+        self.data.set('window', 'root_y', str(position[1]))
+
+    def position(self):
+        return self.data.getint('window', 'root_x', fallback=10), self.data.getint('window', 'root_y', fallback=10)
+
     def maximized(self):
         return self.data.getboolean('window', 'maximized', fallback=False)
 
--- a/pdfarranger/config.py
+++ b/pdfarranger/config.py
@@ -111,7 +111,7 @@ class Config(object):
         self.data.set('window', 'width', str(size[0]))
         self.data.set('window', 'height', str(size[1]))
 
-    def set_position(self,position):
+    def set_position(self, position):
         self.data.set('window', 'root_x', str(position[0]))
         self.data.set('window', 'root_y', str(position[1]))
 
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -406,6 +406,7 @@ class PdfArranger(Gtk.Application):
         if self.config.maximized():
             self.window.maximize()
         self.window.set_default_size(*self.config.window_size())
+        self.window.move(*self.config.position())
         self.window.connect('delete_event', self.on_quit)
         self.window.connect('focus_in_event', self.window_focus_in_out_event)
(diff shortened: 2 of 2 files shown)


dreua:
Err you understood this the wrong way around, the rest of your commit message title was great I just wanted the number / link in the commit message body.

JATIN2111999:
Now

JATIN2111999:
hey @dreua is my PR right or wrong i am still confuse can you give any conformation i resolve all changes in single commit and also added whitespaces after commas.

pdfarranger/config.py
```diff
@@ -111,6 +111,13 @@ def set_window_size(self, size):
         self.data.set('window', 'width', str(size[0]))
         self.data.set('window', 'height', str(size[1]))
 
+    def set_position(self,position):
```
dreua, line 114: Here is still a space miissing

dreua (changes requested):

dreua:
The commit message I had in mind was: 
```
Remember window position on screen

Closes: https://github.com/pdfarranger/pdfarranger/issues/397
```
(Including the empty line)

pdfarranger/config.py
JATIN2111999: oh damn sorry

JATIN2111999: done

kbengs:
There is one line over 100 characters. Codacy use to compalin about that. Don't know why it doesn't now.
Maybe a bit late to comment but is this something we should do? Should the application itself decide where it will open up? Or is that something that the desktop environment should handle..? On KDE I see no other application that will remember it's position.

dreua:
Most applications I just tested on Gnome remember their position: Firefox, Signal, Telegram Spotify
The Gnome applications like Terminal and Nautilus however, don't. Seems like it's up to the application.
Actually I don't care but I think it doesn't hurt and we can close this old issue with this PR. But we could of course also close it without this PR.

JATIN2111999:
Yes you are right about that but it is necessary in some use cases you want your application to take that some place where you closed it so it can be helpful to navigate your mouse cursor and that's why @dreua created this issue.

dreua:
> @dreua created this issue.

I did not do such a thing.

JATIN2111999:
> > @dreua created this issue.
> 
> I did not do such a thing.

Oh yeah right sorry

dreua:
> There is one line over 100 characters.

Yeah that looked long to me, too. Wasn't sure whether we break at 80 or 100 characters or not at all. If we want to enforce this, we need this checked by CI and documented, I'll merge this for now. Thanks @JATIN2111999 !

dreua:
> it can be helpful to navigate your mouse cursor.

Wait, now I'm curious: Are you automating PDF Arranger with scripted mouse movement?

JATIN2111999:
> > it can be helpful to navigate your mouse cursor.
> 
> Wait, now I'm curious: Are you automating PDF Arranger with scripted mouse movement?

There is no mouse movement I am just storing the window co-ordinates and moving the window to the last terminated co-ordinate that's all.

dreua:
In this PR, yes I'm aware of that, I read the code. My question was about your use case

JATIN2111999:
> In this PR, yes I'm aware of that, I read the code. My question was about your use case

My OS don't remembered the position of window and also we are storing all properties like window size and acceleration this property is important to some extent.
So the use case will be like if we close the program by mistake we can open the application at the same place instead of finding the window in new place this may be very useful for big screen users as we are storing the window size we also giving them the exact position of window where it was terminated.

PR #451, opened 2021-03-28
Github Actions: create Windows packages on each push

### LINKED ISSUE #417: CI: Windows builds
> Should we give fully automated Windows builds another try? I think an exe build per PR would be useful for testing.

Yes ! Well I would prefere to have it for `main` (only or also ?). This way we could have a link in the wiki or even in the `README.md`. I think very few people will test binaries from PR but a bit more would test `main`.

_Originally posted by @jeromerobert in https://github.com/pdfarranger/pdfarranger/issues/416#issuecomment-777795410_

jeromerobert:
Close #417

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -19,6 +19,20 @@ jobs:
       run: python3 -X tracemalloc -u -m unittest discover -v -f -s tests
     - name: Codecov
       run: curl -s https://codecov.io/bash | bash
+  build-win32:
+    runs-on: ubuntu-latest
+    container:
+      image: jeromerobert/wine-mingw64
+      options: --user root
+    steps:
+    - uses: actions/checkout@v2
+    - run: ./setup.py build
+    - run: HOME=/root wine python setup_win32.py bdist_msi
+    - run: HOME=/root wine python setup_win32.py bdist_zip
+    - uses: actions/upload-artifact@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/install.yml
```diff
@@ -19,6 +19,20 @@ jobs:
...
+      options: --user root
+    steps:
+    - uses: actions/checkout@v2
+    - run: ./setup.py build
+    - run: HOME=/root wine python setup_win32.py bdist_msi
+    - run: HOME=/root wine python setup_win32.py bdist_zip
+    - uses: actions/upload-artifact@v2
+      with:
+        name: Windows installer
+        path: dist/*
```
dreua, line 35: This creates one zip which is 68 MB in size, containing the MSI with 37MB and the ZIP with 35MB. Most users won't need both, what do you think about putting those in separate zips? (This would be a workaround to Github's current limitations.)

dreua (approved):
Just one idea, the rest looks great to me, thank you so much for taking care of this!!

.github/workflows/install.yml
dreua: Of course this can be merged as is, we can always change it later.

jeromerobert:
> Most users won't need both, what do you think about putting those in separate zips?

Done in 6c68c47

PR #452, opened 2021-03-28
Use visible size when creating blank pages

angsch:
With this fix `create blank page` considers crops and scaling when suggesting a page size.

Behavior so far: The suggested size is the original page size, not the visible area.

image


kbengs:
Merged in 20370a4

Issue #437, opened 2021-03-19
Copy and paste of pages with annotations


m-holger:
When using PDFarranger to duplicate pages that contain annotations using copy and "paste after", or using "duplicate",
I observed the following:

- Editing the annotation (using Okular) on one page would modify the annotation on the other page (after save & reload)
- A new annotation created on one page would also appear on the other page  (after save & reload)

Even deleting all annotations (and saving/reloading) does not clear the problem. Any subsequent annotations are still shared by all copies of the page, including new copies created when the pages had no annotations. Even copying such a page into a new file does not clear the problem.  

Duplicating pages without annotations works as expected, i.e. any subsequent annotations only affect the page annotated, not any copies of the page.

kbengs:
This sounds like #291 What version of PDF Arranger do you have?

m-holger:
I run pdfarranger 1.7.0 installed from source on a Mint 20.1, with pikepdf 2.8.0. 

To me,  this does not sound like the same problem as # 291. Particularly, the problem does occur on files loaded from disk.

I know very little about pdf files. That said, my first impression is:

For each page, there is a container object that contains all the annotations for that page (i.e. all the highlights, boxes and comments) . This container appears to get created when the first annotation is placed on the page, and does not (at least on Okular) ever get deleted, even when all annotations on the page are deleted.  When a page is duplicated, the cloned page receives a reference to annotations on the original page rather than a deep copy, which means that the original as well as all cloned pages share all current and future annotations.

I don't know whether my impression is plausible, but that's the best I can come up with at the moment. Unfortunately I do not have time to delve into the code right now

kbengs:
You are right it's not bug 291. I am able to reproduce it. Annotations seems to be referenced when a page is copied.

m-holger:
The main problem (from a user  perspective) is not that the annotations themself are referenced, but that the container holding all annotations on a page is referenced. The only work-around that I have at the moment is to print the affected page to a pdf file and then replace the problem page and its clones with the printed version.

[...]

PR #453, opened 2021-03-29
Fix error with annotations on duplicated pages

m-holger:
Ensure duplicated pages receive a copy of the annotations, not a reference
Fixes https://github.com/pdfarranger/pdfarranger/issues/437

### DIFF
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -171,6 +171,12 @@ def export(input_files, pages, file_out, mode, mdata):
         new_page = pdf_output.make_indirect(new_page)
 
         pdf_output.pages.append(new_page)
+        # Ensure annotations are copied rather than referenced
+        # https://github.com/pdfarranger/pdfarranger/issues/437
+        if pikepdf.Name.Annots in current_page:
+            pdf_temp = pikepdf.Pdf.new()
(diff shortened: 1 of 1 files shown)


pdfarranger/exporter.py
```diff
@@ -171,6 +171,12 @@ def export(input_files, pages, file_out, mode, mdata):
         new_page = pdf_output.make_indirect(new_page)
 
         pdf_output.pages.append(new_page)
+        # Ensure annotations are copied rather than referenced
+        # https://github.com/pdfarranger/pdfarranger/issues/437
+        if pikepdf.Name.Annots in current_page:
+            pdf_temp = pikepdf.Pdf.new()
+            pdf_temp.pages.append(pdf_input[row.nfile - 1].pages[row.npage - 1])
```
dreua, line 178: Can we use current_page here? If no, why not? Would improve readability imo.

dreua (changes requested):

pdfarranger/exporter.py
m-holger: @dreua Done

kbengs:
Looks good to me. :+1:

dreua (approved):

dreua:
Thank you!

PR #454, opened 2021-03-29
Fix some codacy warnings

dreua:
Let's run the CI...

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -114,8 +114,7 @@ from .iconview import CellRendererImage
 from .iconview import IconviewCursor
 from .iconview import IconviewDragSelect
 from .config import Config
-from .core import img2pdf_supported_img
-from .core import *
+from .core import img2pdf_supported_img, PageAdder, PDFDocError, PDFRenderer
 GObject.type_register(CellRendererImage)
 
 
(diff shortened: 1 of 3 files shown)


dreua:
Should fix around 12 issus: https://app.codacy.com/gh/pdfarranger/pdfarranger/pullRequest?prid=7175081

pdfarranger/pdfarranger.py
```diff
@@ -114,7 +114,7 @@
 from .iconview import IconviewCursor
 from .iconview import IconviewDragSelect
 from .config import Config
-from .core import img2pdf_supported_img
+from .core import img2pdf_supported_img, PageAdder, PDFDocError, PDFRenderer
 from .core import *
```
dreua, line 118: TODO: Check if we can entirely remove the * import

angsch: I would play it safe and import everything listed in `__all__` in core.py

```suggestion
from .core import img2pdf_supported_img, Page, PageAdder, PDFDoc, PDFDocError, PDFRenderer
```

dreua: Good idea!

BTW: How do you suggest this change so that I can commit it right away? Have seen that every now and then but it seems to be well hidden every time I review code...

dreua: Well that caused some other warnings... Maybe we should stuck with the strand accept those?

angsch:
I guess that is the right moment to sigh. Let's go for your initial proposal
```diff
-from .core import img2pdf_supported_img
+from .core import img2pdf_supported_img, PageAdder, PDFDocError, PDFRenderer
-from .core import *
```
to get rid of this alert once and for all. I couldn't find anything that breaks.

angsch:
@dreua Can you create a new PR from your original commit? I think that this is easier than cherry-picking.

dreua:
Yeah, will do! Sorry I've got a lot of stuff going on lately...

dreua:
@angsch ready for review, sorry for taking so long

PR #456, opened 2021-03-30
Render on demand

kbengs:
I think this is working quite well now. At least I don't know of any issues with it. The improvements in the commit:
* Lower memory usage
* No need to wait long for thumbnails when scrolling
* App has better overall responsiveness while rendering
  (scrolling, dragging window)
* Only new or changed pages need to be (re)rendered
  (copy-pasting one page only need to render one page)
* App is more suitable for larger pdfs

The drawbacks:
* When scrolling in a larger pdf it will take some time for pages to be rendered

### DIFF
--- a/pdfarranger/core.py
+++ b/pdfarranger/core.py
@@ -72,7 +72,9 @@ class Page:
         self.size = list(size)
         self.angle = angle
         self.thumbnail = None
-        self.resample = 2
+        self.resample = -1
+        #: A low resolution thumbnail
+        self.preview = None
         self.scale = scale
         #: The name of the original file
         self.basename = basename
@@ -133,6 +135,8 @@ class Page:
         if incl_thumbnail == False:
             del r.thumbnail  # to save ram
             r.thumbnail = None
+            r.resample = -1
+            r.preview = None
         return r
 
     def set_size(self, size):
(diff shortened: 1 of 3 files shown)


angsch:
Much smoother, great work!

- The numbering of the thumbnails starts with 0. Should it start with 1 to be consistent with the page numbers?
- I loaded a JPG image and zoomed in further and further until I got
```
[...]/pdfarranger/core.py", line 490, in update
cairo.FORMAT_ARGB32, int(0.5 + w * scale), int(0.5 + h * scale)
cairo.Error: invalid value (typically too big) for the size of the input (surface, pattern, etc.)
```
The thumbnail is not rendered, but PDF Arranger keeps running without any problems. Not a priority in my opinion - PDF Arranger is not an image viewer.

kbengs:
> The numbering of the thumbnails starts with 0.

I changed so that it starts from 1. (info in statusbar)

> I loaded a JPG image and zoomed in further and further until I got

This problem exist also in main, so I did't do anything about that here.
Maybe we should limit max zoom level so that a page could be zoomed to a size of max twice the
width of window or similar.

I also changed preview size so that it is now about 4000 pixels regardless of page size.

angsch:
Looks good to me. I support merging this into main in spite of the codacy warnings - those will disappear with dreua's PR.

PR #459, opened 2021-04-04
Fix error with annotations on duplicated pages

m-holger:
Ensure duplicated pages receive a copy of the annotations, not a reference
Ensure annotations are stored as indirect objects
Fixes https://github.com/pdfarranger/pdfarranger/issues/437
Fixes https://github.com/pdfarranger/pdfarranger/issues/437#issuecomment-812852016

### DIFF
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -170,13 +170,17 @@ def export(input_files, pages, file_out, mode, mdata):
         # https://github.com/pikepdf/pikepdf/issues/174
         new_page = pdf_output.make_indirect(new_page)
 
-        pdf_output.pages.append(new_page)
         # Ensure annotations are copied rather than referenced
         # https://github.com/pdfarranger/pdfarranger/issues/437
-        if pikepdf.Name.Annots in current_page:
-            pdf_temp = pikepdf.Pdf.new()
-            pdf_temp.pages.append(current_page)
-            pdf_output.pages[-1].Annots = pdf_output.copy_foreign(pdf_temp.pages[0].Annots)
+        if pikepdf.Name.Annots in new_page:
+            new_page.Annots = pikepdf.Array(
+                [pdf_output.make_indirect(copy.copy(annot))
+                    for annot in new_page.Annots])
+        # Annotations must be indirect or they cannot be edited
+        # Using copy.copy(annot) instead of pikepdf.Dictionary(annot)) to
+        # make a new annotations dictionary for compatibility with pikepdf < 2.8.0
+
+        pdf_output.pages.append(new_page)
 
     if exportmode in ['ALL_TO_MULTIPLE', 'SELECTED_TO_MULTIPLE']:
         for n, page in enumerate(pdf_output.pages):


kbengs:
Hm, one of the original problems is back: If a "inline note" is edited (in okular) on one page it will be reflected on a copied page also.
You also would need to make the change on top of latest main, because you are now changing lines that exist in main.
I would: (in feature branch)
```
git fetch upstream
git rebase upstream/main
```

m-holger:
@kbengs Sorry  & thanks. Do my changes work?

kbengs:
No not yet. :smiley:  Now the problem is that annotations can't be modified or deleted for pdfs that are saved with pdfarranger.
You also have one extra empty line on 180. (two empty lines after each other) Don't give up and feel no hurry. ;-)

m-holger:
@kbengs Arghh! Tested it on the wrong output file.

kbengs:
Nice! I can't find anything that does not work now. Let's see if someone else has some comments.

angsch:
We may want to consider updating the dependency on the pikepdf version.

I get a crash with pikepdf 2.7.0
```
  File "[...]/pdfarranger/pdfarranger.py", line 722, in choose_export_pdf_name
    self.save(mode, file_out)
  File "[...]/pdfarranger/pdfarranger.py", line 159, in wrapper
    func(*args, **kwargs)
  File "[...]/pdfarranger/pdfarranger.py", line 796, in save
    exporter.export(self.pdfqueue, to_export, file_out, mode, m)
  File "[...]/pdfarranger/exporter.py", line 178, in export
    for annot in new_page.Annots])
  File "[...]/pdfarranger/exporter.py", line 178, in <listcomp>
    for annot in new_page.Annots])
  File "[...]/site-packages/pikepdf/objects.py", line 210, in __new__
    return _qpdf._new_dictionary(d)
TypeError: _new_dictionary(): incompatible function arguments. The following argument types are supported:
    1. (arg0: dict) -> pikepdf._qpdf.Object
```
After an upgrade to the latest pikepdf version your commit worked marvelously.

m-holger:
I will change

``` python
 [pdf_output.make_indirect(pikepdf.Dictionary(annot))]
```
to
``` python
 [pdf_output.make_indirect(copy.copy(annot))]
```
which is what pikepdf >= 2.8 is doing under the hood anyway.

I think requiring pikepdf >- 2.8.0 might be a good idea - it certainly would improve the readability of `exporter.export` if the workarounds could be removed.

m-holger:
@kbengs  Could you please recheck. Thanks

dreua:
> We may want to consider updating the dependency on the pikepdf version.


> I think requiring pikepdf >- 2.8.0 might be a good idea

Unfortunately, when packaged for Linux distros, requiring something which is too new is a problem because the pikepdf package may not be available in a sufficient version.

Repology seems to be a good source for more information on packaged versions: https://repology.org/project/python:pikepdf/versions

I once created this version Matrix with repology, maybe it can help with the decision-making - editing the URL should be quite self-explanatory:

https://repology.org/badge/versions-matrix.svg?projects=pdfarranger,python:pikepdf%3E=1.7,pdfshuffler&header=pdfarranger%20packaging%20status

I know that the pikepdf-Fedora-package-maintainers won't update to 2.x during a release because they consider this a major release (It is according to [Semantic Versioning](https://semver.org/lang/de/) because it goes from 1.x to 2.x, not sure if that is actually a breaking change in pikepdf that would prevent upgrading.) At any rate, the fallout should be carefully considered before raising the required versions. Probably in its own issue for better clarity and structure.

kbengs:
I couldn't find any issues with it, but.. I found another issue with annotations. If a page with annotations is rotated, the annotation is not rotated. But that's probably for a separate issue/commit.

This might be details, but the commit message should describe what the commit does, so maybe:
```
Commit title: Ensure annotations are stored as indirect objects

Commit body: Fixes https://github.com/pdfarranger/pdfarranger/issues/437#issuecomment-812852016
```

The three new comment lines after the code are good info for reviewers. But I'd say thouse three lines
could instead be here in this forum (or in commit message). Also I think in pdfarranger a comment is never after the code.

m-holger:
@kbengs 

> EDIT: I saw that issue with "inline note". Tested in Okular. Could maybe be a bug i Okular also.

Same behaviour in Mint Cinnamon XReader . 

I only found this behaviour with /FreeText annotations, and all of them had the NoZoom and NoRotate flags set. So this behaviour looks correct to me. In any case, the PDF 1.7 spec states that the annotations /Rect entry is not modified by rotation or scaling, so any issue would be a reader rather than a PDF Arranger issue.

m-holger:
@dreua 

> At any rate, the fallout should be carefully considered before raising the required versions.

Accepted, but if relatively old versions of pikepdf need to be supported than your wiki page should probably state that code should be developed in a venv with pikepdf==1.7.0 (with a quick guide on how to install that version)

@kbengs 

> commit message should describe what the commit does

>  a comment is never after the code

No issue with that, but again, this should be made clear to potential contributors.


> The three new comment lines after the code are good info for reviewers. But I'd say those three lines
> could instead be here in this forum (or in commit message).

Here I would disagree. If something has to be coded in a specific way for reasons that are not readily apparent than this should be documented in the code itself for the benefit of future readers,

kbengs:
I think this is not relevant any more since abb3f02. I can reopen if I was wrong.

Issue #460, opened 2021-04-10
About bumping pikepdf version requirement


jeromerobert:
> We may want to consider updating the dependency on the pikepdf version.

> I think requiring pikepdf >- 2.8.0 might be a good idea

Unfortunately, when packaged for Linux distros, requiring something which is too new is a problem because the pikepdf package may not be available in a sufficient version.

Repology seems to be a good source for more information on packaged versions: https://repology.org/project/python:pikepdf/versions

I once created this version Matrix with repology, maybe it can help with the decision-making - editing the URL should be quite self-explanatory:

https://repology.org/badge/versions-matrix.svg?projects=pdfarranger,python:pikepdf%3E=1.7,pdfshuffler&header=pdfarranger%20packaging%20status

I know that the pikepdf-Fedora-package-maintainers won't update to 2.x during a release because they consider this a major release (It is according to [Semantic Versioning](https://semver.org/lang/de/) because it goes from 1.x to 2.x, not sure if that is actually a breaking change in pikepdf that would prevent upgrading.) At any rate, the fallout should be carefully considered before raising the required versions. Probably in its own issue for better clarity and structure.

Edit: I could try discussing and fixing this for Fedora with the pikepdf maintainers. However, there are many distributions and I expect them to have very similar rules and problems.

_Originally posted by @dreua in https://github.com/pdfarranger/pdfarranger/pull/459#issuecomment-815672525_

jeromerobert:
From Repology I would says that `>=1.17` would be a good trade off.

dreua:
For comparison: Our current requirement in `setup.py` is `pikepdf>=1.7.0`

I think `>=1.17` should be no problem, `>=1.17.3`  should be fine, too, not sure whether it makes a difference for us.

m-holger:
Perhaps another point to consider

```
v2.10.0

-    Fixed a XML External Entity (XXE) processing vulnerability in PDF XMP metadata parsing. 
     (Reported by Eric Therond of  Sonarsource.) 
     All users should upgrade to get this security update. CVE-2021-29421 was assigned to this issue.
```

dreua:
Wrong audience: Please report this to the pikepdf maintainers of all the distros out there! We are not forcing anyone to use outdated pikepdf, PDF Arranger is, afaik, fully compatible with the latest pikepdf. (I'll update the flatpak asap, thanks for the reminder!)

m-holger:
@dreua

[...]

Issue #458, opened 2021-04-02
Unusual drag&drop behaviour


m-holger:
Drag&drop between two instances of PDF Arranger behaves like cut&paste. This feels wrong and appears inconsistent with other applications such as a file manager:

#### Case 1 - the two instances have different documents open

- This is akin to draging a file onto a different volume. The default behaviour for this is to copy rather than move.

- It is unlikely that the user intents to edit both file. The more likely intent is to copy pages into the destination document.

#### Case 2 - the two instances have the same documents open

There are two likely scenarios:

1. The user is trying to copy pages to a relatively distant part of the document.  

2. The user is trying to move pages to a relatively distant part of the document. Here the cut&paste behaviour would seem appropriate, except that this does not work as the two instances work on different copies of the document and therefore the re-saved document will either reflect the cut or the paste, but not both.

kbengs:
Windows 10 copy when you drag to another volume, but moves when you drag to another folder on same volume. Imo it should always move. :-)
In linux default for all filemanagers I tried is move.

@m-holger have you tried to hold ctrl while you drag?

m-holger:
@kbengs The default manager Nemo on Linux Mint Cinnamon certainly copies when dragging files onto a different volume. (Move is the default within a volume)

Dragging an image file onto a document does not delete the file.

Does your file manger really delete a file from your hard disk when you drag it onto a usb stick?

Imo drag&drop should only move without explicit user action when it does not pose a risk of data loss, and even then only when there is a reasonable expectation that this is what the user actually wants. Moving pages onto a document that may ultimately not be saved (either as result of user choice or of a crash) risks data loss. On top of that, I can only think of two relatively rare scenarios where moving makes sense in the context of PDF Arranger.

kbengs:
No you are right. It the target is a usb stick then it will copy. You might have a point here.

PR #462, opened 2021-04-11
Don't trash table of contents and links on export

m-holger:
Experimental rewrite of exporter.export
- preserve internal links of main file (i.e. pdfqueue[0])
- preserve table of content/bookmarks of main file
- preserve form fields of main file

Definitely requires further work if this looks promising

### DIFF
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -25,12 +25,16 @@ import io
 import gi
 import locale
 import packaging.version as version
+from typing import Any, Dict, List
+
 from . import metadata
 from gi.repository import Gtk
 gi.require_version("Poppler", "0.18")
 from gi.repository import Poppler
 import gettext
 _ = gettext.gettext
+
+from .core import Page
 
 # pikepdf.Page.add_overlay()/add_underlay() can't place a page exactly
 # if for example LC_NUMERIC=fi_FI
@@ -226,6 +230,21 @@ def _apply_geom_transform(pdf_output, new_page, row):
     return _scale(pdf_output, new_page, row.scale)
 
 
+def _apply_geom_transform_job(pdf_output:pikepdf.Pdf, new_page:pikepdf.Page, page:Page) -> None:
+    new_page.rotate(page.angle, relative=True)
+    new_page.MediaBox = _mediabox(new_page, page.crop)
+    # add_overlay() & add_underlay() will use TrimBox or CropBox if they exist
+    if '/TrimBox' in new_page:
+        del new_page.TrimBox
+    if '/CropBox' in new_page:
+        del new_page.CropBox
+    if page.scale != 1:
+        pdf_output.pages.append(new_page)
+        new_page.obj.emplace(_scale(pdf_output, pdf_output.pages[-1], page.scale))
+        del(pdf_output.pages[-1])
+
+
+
 def _remove_unreferenced_resources(pdfdoc):
     try:
         pdfdoc.remove_unreferenced_resources()
@@ -349,6 +368,53 @@ def _append_page(current_page, copied_pages, pdf_output, row):
         new_page.Annots = pdf_output.copy_foreign(indirect_annots)
 
 
+def _transform_job(pdf_output: pikepdf.Pdf, pages: List[Page], quit_flag = None) -> None:
+    """ Same as _copy_n_transform, except it doesn't copy. Requires pikepdf >= 8.0 """
+    # Fix missing MediaBoxes
+    for page in pdf_output.pages:
+        if page.mediabox is None:
+            page.mediabox = pikepdf.Array((0, 0, 612, 792))
+
+    # We don't need to call _append_page as the Job interface copies pages / annotations as necessary.
+    mediaboxes:List[pikepdf.Object] = []
+    i = 0
+    for page in pages:
+        if quit_flag is not None and quit_flag.is_set():
+            return
+        mediaboxes.append(pdf_output.pages[i].mediabox)
+        _apply_geom_transform_job(pdf_output, pdf_output.pages[i], page)
+        for lpage in page.layerpages:
+            i += 1
+            _apply_geom_transform_job(pdf_output, pdf_output.pages[i], lpage)
+        i += 1
+
+    # # Add overlays and underlays
+    for i, page in enumerate(pages):
+        # The dest page coordinates and size before geometrical transformations
+        dx1, dy1, dx2, dy2 = mediaboxes[i]
+        dw, dh = dx2 - dx1, dy2 - dy1
+
+        # Call to rotate in _apply_geom_transform_job ensures /Rotate exists
+        rotate_times = int(round((pdf_output.pages[i].Rotate % 360) / 90) % 4)
+        for lpage in page.layerpages:
+            # Rotate the offsets so they are relative to dest page
+            offset = lpage.rotate_array(lpage.offset, rotate_times)
+            offs_left, offs_right, offs_top, offs_bottom = offset
+            x1 = page.scale * (dx1 + dw * offs_left)
+            y1 = page.scale * (dy1 + dh * offs_bottom)
+            x2 = page.scale * (dx1 + dw * (1 - offs_right))
+            y2 = page.scale * (dy1 + dh * (1 - offs_top))
+            rect = pikepdf.Rectangle(x1, y1, x2, y2)
+
+            if lpage.laypos == 'OVERLAY':
+                pdf_output.pages[i].add_overlay(pdf_output.pages[i + 1], rect)
+            else:
+                pdf_output.pages[i].add_underlay(pdf_output.pages[i + 1], rect)
+            # Remove the temporary added page
+            del pdf_output.pages[i + 1]
+
+
+
 def export_doc(pdf_input, pages, mdata, files_out, quit_flag, test_mode=False):
     """Same as export() but with pikepdf.PDF objects instead of files"""
     pdf_output = pikepdf.Pdf.new()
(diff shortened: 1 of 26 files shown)


dreua:
### UX
On the one hand, this sound like a good idea, on the other I'm afraid it will be hard to explain to people what the "first file" is and why it only works on that. This could easily be perceived as an unreliable or unpredictable feature and therefore lead to a worse user experience (UX). Just a first thought for discussion, I'm not decided yet. Another thought might be a bit selfish: I'd rather have no feature than one that people constantly complain about.

### Technical
What happens when pages are deleted/reordered? Will the three features you mention still work and point to the correct page?

m-holger:
@dreua 

>  it will be hard to explain to people what the "first file" is and why it only works on that

I would describe the first file as the file opened for editing, and the remaining files as imported files. That the first file is the document being edited is partly recognised by the fact that it is the default destination for saving. Not explicitly making the distinction between the file being edited and the imports has other consequences, including:-
- on saving, warning that the edited document already exists (no surprise), but no warning that it has been modified on disk by another application/another instance of PDF Arranger (potential surprise, potential data loss)
- lack of protection against opening the same document multiple times in PDF Arranger for competing edits (potential data loss)
- inconsistent behaviour of cut&paste/drag&drop of the same document intra-instance vs multi-instance of PDF Arranger (potential data loss, potential dramatic increase in file size). [See also #458. The proposed change would provide for predictable objgens of pages, which would allow addressing the issues]
 

>  Will the three PDF features you mention still work and point to the correct page?

Links to existing pages will work, even if the pages have been reordered. Obviously links to deleted pages will be broken. While not ideal, in my view better than trashing all links. For example, what is the point of having a copy of the [pikepdf documentation](https://readthedocs.org/projects/pikepdf/downloads/pdf/latest/) without any of the links?

kbengs:
I think that looks promising. I really miss those things. To make it more clear what is the main pdf, maybe we could:
* Only show the name of the main pdf in headerbar (not any of the other imported documents)
* Have a open dialog for opening of the main document
* Have the import dialog for opening of other documents

m-holger:
@dreua 
> What would be the minimum pikepdf version for this feature?

My code runs with pikepdf  1.10.3 (It may work with earlier versions - I haven't tested it). At the moment I can't see any reason why the workarounds that I have commented out can't be fixed so they don't interfere with my code.

dreua:
> My code runs with pikepdf 1.10.3 (It may work with earlier versions - I haven't tested it). At the moment I can't see any reason why the workarounds that I have commented out can't be fixed so they don't interfere with my code.

1.10.3 would be fine according to #460 . I'm not sure what you mean by your second sentence: I think those workarounds are or will be fixed only in a new pikepdf version (>= 2.x.x) and that could be a problem as I've explained earlier.

dreua:
### Goals
I understand the drive behind this PR and the idea sounds good to me. However, I wonder what our long term goals are and whether this is a step in the right direction or rather a detour from the actual goals.

I guess the two of you know more about PDFs and the libraries than me: Would it be a reasonable and achievable long term goal to support proper merging of these PDF-features, i.e. no need to treat the first PDF as special but all PDFs as equal while merging and retaining all links, annotations etc.?

m-holger:
@dreua
> I wonder what our long term goals are

Me too. Some broad vision statement from the team and indication of a roadmap would be useful

> I guess the two of you know more about PDFs and the libraries than me

I assume you are referring to kbengs and lgtm-com. Most of my knowledge of PDFs (other than from a user perspective) and all of my knowledge of the libraries was aquired over the last 6 weeks.

> no need to treat the first PDF as special but all PDFs as equal

Other than for a simple concatenation of PDFs I don't see how this will be possible. Certainly for the way I use PDF Arranger there always is one file that is being edited, and (usually) others that supply imports.

> Would it be a reasonable and achievable long term goal to support proper merging of these PDF-features

I don't know enough to judge whether it is achievable, but it certainly would be my long-term goal - and I think a reasonable one - to import not only page content, but also everything related to the pages (such as bookmarks etc). However, my immediate priority  is to be able to edit a PDF without destroying its structure and transforming it into a simple concatenation of individual pages. In my personal view, this is critical if the vision for PDF Arranger is to become a successful tool with broad appeal.

m-holger:
@dreua

>  I'm not sure what you mean by your second sentence: I think those workarounds are or will be fixed only in a new pikepdf version (>= 2.x.x)

What I meant is that (as far as I can tell at the moment) the workarounds can be modified so that they can be integrated into my version of export. I understand that the workarounds, in some form, will be required for the foreseeable  future

dreua:
Alright, of course, now I get it!

m-holger:
My implementation of export coincidentally side-steps the pikepdf==2.6.0 issue; the  pikepdf<2.7 workaround is therefore not required.

I am ignoring the   pikepdf<=1.10.1 workaround for the moment in view of #460

m-holger:
I have pushed a updated version.  Main changes
- smaller output file (workaround for  https://github.com/pikepdf/pikepdf/issues/196)
- scaling a page does not break links

kbengs:
I can't give valuable feedback on this because my knowledge about pdfs and pikepdf is very limited. But I did a quick test and noticed problems with scaled pages. With one pdf links go to wrong pages. With another pdf the links don't seem to work at all.

Have you had any thoughts about the case when user deletes all pages and open a new pdf? The new file will not be pdfqueue[0]. Pdfqueue can't just be deleted either because undomanager need to have access to old files. How should it ideally work according to you (anyone) ?

dreua:
@kbengs thanks for your input, I wasn't aware of those issues you describe.

In general: I think the qpdf dev is working on features that will eventually allow us to fully support those pdf features without the "first PDF" Hack. I personally am fine with waiting for that.

m-holger:
@kbengs Thanks for your feedback

> I can't give valuable feedback on this

Any report of problems is valuable and helpful

> noticed problems with scaled pages

Could you share the pdf / instructions to duplicate the problem

> Have you had any thoughts about the case when user deletes all pages and open a new pdf?

This goes slightly outside of what I am trying to do into the wider question of  UI design. At the moment I assume that pdfqueue[0] is the file being edited, but for the purposes of the export function it would be just as easy to pass in the file id of the main file as a parameter.

As to the specific scenario you raise - the problem at the moment is that there is no "Open" dialogue. If there was an "Open" dialogue (and corresponding "New" for the case where the user wants to start with an empty document) the situation would be pretty clear cut. In the absence of an "Open" dialogue we could assume that the next file imported after all pages have been deleted becomes the file being edited. However, I would question whether this would be clear to the user or whether it would appear as random and unpredictable. in my view, it would be better to be explicit about what is being edited and what is being imported.

m-holger:
@dreua

> In general: I think the qpdf dev is working on features that will eventually allow us to fully support those pdf features without the "first PDF" Hack

QPDF currently preserves outlines and links for the first file, and employs the same "hack"

I would not consider distinguishing between the main document and imported documents a hack - I believe this is what normally happens and I don't think that QPDF is going to move away from this model even as it improves support for importing outlines and named destinations.

kbengs:
It's with this pdf https://files.danfoss.com/download/Drives/Vacon-20-Complete-Manual-DPD00716F1-UK.pdf
* Select all and scale to 50%
* Export
* Test the links on page 3

The pdf that seems to break links I can't share.

Regarding the question whether include this or not, I can at least say I have found this useful on the battlefield. :-)

m-holger:
@kbengs Thanks. Looks to me like the links work fine - they are just not where you would expect them. Something to sort out.

Could you try your other file with a scale factor of 99.5% to see whether its the same problem or something else.

kbengs:
Yes with 99,5% scale the other pdf works fine also.

m-holger:
@kbengs Links and other annotations should now be in the right place after scaling

kbengs:
If we would merge this PR I think we also at least need to have a open dialog to open the main document.
@jeromerobert @angsch do you have any opinions about this PR?

@m-holger If a document is removed and a new one is opened it is currently not possible to export from the second document.
(pikepdf._qpdf.PdfError: .....invalid Type /QPDFFakeName in page tree)

m-holger:
@kbengs I have put a fix in for this case, although I can not think of a situation where deleting all pages from the main document makes much sense. If you had a genuine reason for doing it please let me know what it is and I'll try to figure out how best to handle it.

pdfarranger/exporter.py
```diff
@@ -75,41 +113,63 @@ def _set_meta(mdata, pdf_input, pdf_output):
...
-        Type=pikepdf.Name.Page,
-        MediaBox=newmediabox,
-        Contents=doc.make_stream(content.encode()),
-        Resources={'/XObject': {'/p{}'.format(page_id): xobject}},
-        Rotate=rotate,
-    )
-    return new_page
+    orig_page.MediaBox = newmediabox
+    #Question - if we scale a cropped page, do we loose content?
+    if '/CropBox' in orig_page.keys(): del(orig_page['/CropBox'])
```
angsch, line 145: Scaling is relative with respect to the origin. Hence, no matter if scaling is active or not, the relative clipped content is the same.

Why is it necessary to delete the CropBox here? If it is necessary, should it moved to `_mediabox()`?

angsch:
Wow, a lot of work as gone into this.

> #Question - what else do we need to scale?

I cannot think of any other drawables other than the ones you already have.

Annotations become mislocated on differently scaled duplicated pages.

1. Load a page with an annotation and duplicate it.
2. Scale the second page to 200.
3. Export.

The annotation on the first page is no longer visible.

The code contains a few comments on potential issues/rewrites addressing for instance the multi-page export. For the sake of simplicity, can we restrict this PR to annotations? Your concerns are valid and please regard this request as encouragement to create issues and/or further PRs.

mara004:
This thread is really interesting because I am currently dealing with the same problem: trying to preserve as much "special" parts of the original documents as possible.

I am somewhat surprised you take one pikepdf document as base and then delete pages. As you know, this is kind of troublesome with pikepdf/qpdf and requires workarounds, though it might be fixed at some point.
However, I think a cleaner solution would be to initialise a new document and append the desired pages. Then you can simply copy over relevant parts of the /Root objects from the input documents to preserve other content, such as form fileds (/AcroForms) and layers (/OCProperties). This would make the strange FixedPageList workaround unnecessray and also ensure no (potentially troublesome) residues of that one base document are left. (I'm not certain whether `remove_unreferenced_resources()` is capable of cleaning it all up.)

I think preserving the table of contents from one input document is not a good idea either. I know little about ToC internals in PDFs, though I believe too much can go wrong, as page references might become invalid or lead to the wrong place. I'm really not sure, but at least it should be verified and tested for different document structures to make sure there are no problems/regressions.

More elegant would be to read the outline items of all source documents and then add them to the new document accordingly, so that "bookmarks" could be preserved for all input PDFs.
This is hard, though, since resolving bookmark destinations is fairly complicated, and pikepdf does not (yet) provide means to do it for you. Threre are three types of references from an outline item to a page: Direct destinations (easy to resolve), named destinations (possible, but more complicated), and actions (very difficult). For details, see https://github.com/pikepdf/pikepdf/issues/149 and https://github.com/pikepdf/pikepdf/issues/173.
(The PyMuPDF library, in contrast, does have a method returning the table of contents with level, description, and page number for each outline entry.)

Another question is whether AcroForms could be preserved for more than one document. PDFsam/sejda is capable of doing so, but it's non-trivial since form widget names need to be deduplicated. I recently opened an issue about this: https://github.com/pikepdf/pikepdf/issues/207 (it's rather caothic, sorry about that).

@angsch's review comment concerning the CropBox is correct. It is not necessary, nor desirable to remove an existing CropBox, since this inadvertently changes page size and might reveal hidden (unwanted) content for some documents. You can simply resize the box by multiplying all its coordinates with your scale factor, preferably in a list comprehension. And, hypothetically, it would also be better to set CropBox to MediaBox rather than using `del()` I think.

This comment got rather long, but I wanted to share these points for consideration/discussion.
PS: Thank you for maintaining pdfarranger ;)

mara004:
> I don't know enough to judge whether it is achievable, but it certainly would be my long-term goal - and I think a reasonable one - to import not only page content, but also everything related to the pages (such as bookmarks etc).

I agree. This would also be my goal. It seems very hard to achieve, though, considering the complex and manifold structure of PDFs. I think it would be best if methods to merge content attached to `/Root` could be added to pikepdf/qpdf directly.

> If we would merge this PR I think we also at least need to have a open dialog to open the main document.

Alternatively, you could make a dropdown menu enabling the user to select which document to use as base, perhaps?

mara004:
> [...] could be added to pikepdf/qpdf directly.

Actually the [todo list of qpdf](https://github.com/qpdf/qpdf/blob/e0775238b8b011755b9682555a8449b8a71f33eb/TODO#L186) already discusses the topic of preserving bookmarks and forms, interestingly.

m-holger:
I may revisit this once pdfarranger bumps the minimum version requirement for pikepdf to at least  6.0 (i.e. QPDF 11.0+), at which point this will become relatively trivial to implement.

ovari:
> I may revisit this once pdfarranger bumps the minimum version requirement for pikepdf to at least 6.0 (i.e. QPDF 11.0+), at which point this will become relatively trivial to implement.

With [Linux Mint 21.1 Cinnamon](https://linuxmint.com/) and [PDF Arranger from Flathub](https://flathub.org/apps/com.github.jeromerobert.pdfarranger), the version of pikepdf is 7.2.0 as shown in the image below.

image

Thank you

m-holger:
However, according to setup.py

```install_requires=['pikepdf>=1.17.0','python-dateutil>=2.4.0'],```

I am working on the assumption that the intention is that PDF Arranger should behave identically whatever supported version of pikepdf and qpdflib is used.

ovari:
> However, according to setup.py
> 
> `install_requires=['pikepdf>=1.17.0','python-dateutil>=2.4.0'],`
> 
> I am working on the assumption that the intention is that PDF Arranger should behave identically whatever supported version of pikepdf and qpdflib is used.

@jeromerobert can you please help? Thank you

jeromerobert:
@ovari I would prefer to keep 1.17 compatibility for PDF Arranger 1.10. 1.11 will be pikepdf >= 6. Anyway I would not merge such possible tricky PR in 1.10 because people would not have time to deeply test it.

PDF Arranger will remain somehow stable until september for fix releases (1.10.x) so you can start to work on current master with pikepdf >=6 in mind and no fear that it change a lot.

jeromerobert:
@m-holger

> I am working on the assumption that the intention is that PDF Arranger should behave identically whatever supported version of pikepdf and qpdflib is used.

I don't think so. Having feature available with new version of pikepdf and not available with old version would be IMHO acceptable.

m-holger:
@jeromerobert 

Noted.

Can you confirm the minimum version of Python supported by PDF Arranger. (Might be useful to add to the "for developers" section of the readme file).

jeromerobert:
> Can you confirm the minimum version of Python supported by PDF Arranger. (Might be useful to add to the "for developers" section of the readme file).

Difficult question. We have a test with Python 3.8 with PikePDF 1.19 in Github Action. But most recent version of PikePDF or Python GTK may not support a so old Python. So it depends on how you install PDF Arranger.

m-holger:
At the moment, PDF Arranger creates Metadata when editing and saving files that do not contain Metadata. Is that a design choice or happenstance? In other words, do you want me to replicate this behavior?

ovari:
> At the moment, PDF Arranger creates Metadata when editing and saving files that do not contain Metadata.

If the source has no metadata, should the target also have no metadata?

m-holger:
> If the source has no metadata, should the target also have no metadata?

(unless the user explicitly sets metadata)

That in a nut shell is the question.

Has the question been decided? If not, what is the decision?

m-holger:
For pikepdf >= 8.0, the current commit preserves outlines for pages from the first file, provided the pages are not converted into xobjects (i.e are not scaled or converted into over/underlays).

I have ignored the cases of splitting output into individual pages or booklet creation as preservation of outlines does not make much sense in those circumstances.

Coverage fails because the code is pikepdf >-8.0 specific, which is not currently tested during CI.

I have not been able to investigate the failures for pikepdf < 6.0 as I have not managed to install PDF Arranger for Python 3.8.

kbengs:
I tested it on a few pdfs and could see no issues. Nice work!

> provided the pages are not converted into xobjects (i.e are not scaled or converted into over/underlays).

@m-holger are you aware of any way to have links working on scaled pages?

m-holger:
> are you aware of any way to have links working on scaled pages?

This can be done by prepending the appropriate instructions to the content stream and fixing a fair number of items in the pdf file. The tricky and time consuming bit is to identify all the bits that need to be fixed. I've written some code a couple of years back which dealt with the common cases, which I will dust off if it doesn't look like a scale methods will make it into qpdf 11.6 or 11.7. 

At the moment I am looking at preserving links (not outlines) in imported files, which looks more straightforward but will need to be fixed in qpdf.

jeromerobert:
I would like to fix the coverage before merging.

jeromerobert:
Ok it's because we do not test with pikepdf >= 8.

m-holger:
@jeromerobert 

> Ok it's because we do not test with pikepdf >= 8.

It might be worth adding pikepdf 8 to CI and coverage considering the amount of code that is specific to 8

jeromerobert:
Yes ! It's in progress.

m-holger:
> At the moment I am looking at preserving links (not outlines) in imported files, which looks more straightforward but will need to be fixed in qpdf.

This will be in 11.6.0

m-holger:
There also will be a repair for files with missing media boxes.

PR #463, opened 2021-04-14
Fix issues with very small & very large thumbnails

kbengs:
To reproduce the issues fixed:
* Open a large pdf
* Zoom in so thumbnail is wider than window
* While it still renders, try to scroll -> it is slow
* Remove the pdf
* Zoom out to smallest possible
* Open a less heavy pdf
* Select all + change scale factor to 1% (smallest possible)
* Zoom in slowly -> at some point thumbnails will be cropped at top and bottom. When cell width has grown wider than 50px the problem is gone.
* Zoom in to max
* Select all + change scale factor to 5000% (largest possible) -> "cairo.Error: invalid value (typically too big)"

Iconview also has issues when iw height goes over 2^23 pixels. Cells in beginning and in the end of iw will not be rendered then. 2^23 pixels will roughly be reached with a 3000 pages pdf when zoomed to max.
This issue is not fixed here.

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -1123,6 +1123,7 @@ class PdfArranger(Gtk.Application):
             self.paste_pages_interleave(data, before, ref_to)
             GObject.idle_add(self.retitle)
             self.iv_selection_changed_event()
+            self.zoom_set(self.zoom_level)
             self.silent_render()
 
     def read_from_clipboard(self):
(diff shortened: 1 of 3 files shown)


pdfarranger/pdfarranger.py
```diff
@@ -1669,18 +1665,31 @@ def sw_scroll_event(self, _scrolledwindow, event):
...
         """Sets the zoom level"""
         level = min(max(level, -10), 40)
-        if level == self.zoom_level:
+        zoom_level_old = self.zoom_level
+        self.zoom_level = level
+        self.zoom_scale = 0.2 * (1.1 ** level)
+        # Limit max zoom level so that thumbnail is max 23Mb
+        if len(self.model) > 0:
+            max_limit = 6000000  # 6000000 pixels * 4 byte/pixel -> 23Mb
+            max_page_size = max(p.size[0] * p.size[1] * p.scale ** 2 for p, _ in self.model)
```
angsch, line 1674: `p.size` is the original size without crops. Do we need `p.size_in_points()` here?

kbengs: The thumbnails are always rendered in full size and crops are applied later. Goal here is to limit memory usage, so I would say this is how it need to be for now. May be that we should change how the rendering work so it would only render the cropped page. (I did a test now where I opened a pdf with many pages, added crop 99% to left and top, and zoomed all out. Memory usage was then 12Gb..) But I'd say that is for a new issue. The 6000000 limit could be higher. Do you think it should be?

dreua: I think it's perfectly fine to optimize for the general/expected use case. If someone really needs 99% crop and this becomes a problem I'd ask them nicely to use a different tool...

I haven't gone into the details of what you are doing here but from what I understand it sounds like great work, thanks! 👍

angsch: Thanks for the explanation. I see the primary use case of first cropping and then rendering in the support of negative crops, but that is probably quite a bit into the future.

angsch:
`self.model[0][0] = self.model[0][0]  # Let iconview refresh itself`
I see that this has an effect and works, but I don't understand why this assignment is sufficient. Could you give an explanation?

kbengs:
Iconview probably just understand that if one cell is updated, sizes and positions of other cells also likely need to be updated. So it will simply always update all cells size and position and re-render the widget.

PR #464, opened 2021-04-15
Add tolerance to detect same page format

angsch:
Accuracy can be lost in the computation of the page format when crops are flipped

- Load an A4 page
- Set the top crop to 6.7 and the bottom crop to 55.3
- Duplicate the page
- Rotate the second page by 180 degrees
- Select one of the two pages
- Select same page format

Only one page is selected although both pages have the same page format. Strange looking crops as above could be the result of the crop-white-borders features. This fix prunes some digits to relax the condition on what is considered the same page format.

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -1230,9 +1230,12 @@ class PdfArranger(Gtk.Application):
         elif selectoption == 'SAME_FORMAT':
             selection = self.iconview.get_selected_items()
             formats = set(model[row][0].size_in_points() for row in selection)
+            # Chop digits to detect same page format on rotated cropped pages
+            formats = [(round(w, 8), round(h, 8)) for (w, h) in formats]
             for row in model:
                 page = model[row.path][0]
-                if page.size_in_points() in formats:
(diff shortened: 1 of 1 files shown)


kbengs:
merged in a4a56bb

dreua:
Sorry for the late review: I see a possible hard to find bug if the values differ in the 9th decimal place and the value there is around 5, so one could be rounded up or down creating a difference in the 8th place. This is not a problem if the difference only appears at a decimal place of lower significance but that might actually be worth a source code comment imo. Feel free to commit that directly without PR if you agree.

In general, floats should be checked for equality by something like `abs(f1 - f2) < allowed_delta`, but I do see the usefulness of rounding here.

angsch:
> In general, floats should be checked for equality by something like abs(f1 - f2) < allowed_delta, but I do see the usefulness of rounding here.

If you find this more readable, please go ahead and commit it. I opted for the truncation to be closer to pikepdf, which operates on Decimals.

dreua:
No, rounding is fine with me. I was just wondering whether the magic number 8 is the correct number of decimals for what you are trying to achieve.

Issue #141, opened 2020-02-11
Wrong Crop / Rotation on duplicated pages


dreua:
When saving pdfs with duplicated pages and the instances of those have different rotations or crop settings, the rotation / cropping of the first instance of the duplicated page is applied to all the following instances.

Steps to reproduce: 
- Load PDF
- Duplicate a page (either Ctrl+Drag or via menu)
- Change rotation or crop of one of the pages
- Save PDF

Expected results:
Saved PDF looks the same as preview

Actual result:
Saved PDF looks different than preview. All instances of a duplicated page have the same crop / rotation applied.

[Originally reported by Fedora user norbertj, thanks for testing!](https://bodhi.fedoraproject.org/updates/FEDORA-2020-8e8c2b9725)

dreua:
Also affects the "Split Pages" feature

jeromerobert:
I try changing `MediaBox` and `Rotate` on the result of `copy_foreign`  instead of on `current_page`. That does not help :disappointed:.

dreua:
Yeah I tried the same, should have written it here to save you the work :/ Looks like a bug in pikepdf then, doesn't it? I'm also looking into it right now...

jeromerobert:
A bug in PikePDF or in QPDF. `copy_foreign` (and `_add_pages` which was used internally before you introduced `copy_foreign`) should be a deep copy, but it's not. So when we update the `MediaBox` or `Rotate` of a page in update all it's duplicates.

dreua:
Yeah that totally makes sense. They probably never hat that use case before but it's useful especially for the "split pages" function. Are you reporting this upstream or would you want me to do that?

dreua:
The qpdf command can handle it though: `qpdf in.pdf --pages . 1,1,1,1 -- --rotate=180:1,3 out.pdf` (This duplicates the first page of in.pdf 4 times and flips the first and third page.)

dreua:
There is still an issue with cropping, I will have to look into that later.

dreua:
Correction: We are good with the `copy` workaround so the fix can be merged but there is still a bug in pikepdf when setting the mediabox while the rotation is fixed in the latest 1.10.2. I'm going to report that since they probably want to fix that too.

dreua:
[...]

PR #472, opened 2021-04-25
Set minimum pikepdf version to 1.17.0

### LINKED ISSUE #460: About bumping pikepdf version requirement

jeromerobert:
* All major distro now have it
* Drop a workaround for https://github.com/pdfarranger/pdfarranger/issues/141 /
  https://github.com/pikepdf/pikepdf/issues/80
* Close #460

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ lines.
 
 ## Install from source
 
-*PDF Arranger* requires [pikepdf](https://github.com/pikepdf/pikepdf) >= 1.7.0, but pikepdf >= 1.15.1 is highly recommended.
+*PDF Arranger* requires [pikepdf](https://github.com/pikepdf/pikepdf) >= 1.17.0.
 pip will automatically install the latest pikepdf if there is no pikepdf installed on the system.
 
 **On Debian based distributions**
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -15,10 +15,7 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 
-import copy
 import pikepdf
-import traceback
-import sys
 import os
 import tempfile
 from . import metadata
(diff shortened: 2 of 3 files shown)


dreua:
I think your part of the CI is failing, no idea why: `/usr/lib/python3.8/distutils/dist.py:274: UserWarning: Unknown distribution option: 'cmd_class'`

dreua (approved):
Changes look good to me. Feel free to merge, but we might have a closer look at the failing CI first.

dreua:
Found the reason for our CI-fail: https://github.com/pikepdf/pikepdf/commit/19b9b0391e8abc00905cea2df4ce0f1c6c0bb9c2#

jeromerobert:
Changing to draft, until I fix `pip install pikepdf`.

jeromerobert:
I think I'm just going to upgrade the CI docker image to Alpine 3.13.5

dreua:
btw: there is a "Rerun Jobs" button at the top right of [the action pages](https://github.com/pdfarranger/pdfarranger/actions/runs/783552216). Looks like you were not aware of that.

Issue #447, opened 2021-03-28
Bug reporting: Create issue templates


dreua:
@pdfarranger how do you feel about issue templates? I think we could have one for feature request and one for bug reporting, each providing some basic structure and information for the issue creator. I see those often around GitHub and I think it could be useful for reporters and improve the initial reports so that we need less back and forth.
Templates are just suggestions, nobody needs to follow them if they don't want to.

I'm looking for feedback from the regular contributers here. If some positive feedback comes in I'll start drafting those templates 

Related: #439 already improved access to debug information within PDF Arranger. I think we can extend that as needed.

angsch:
I have seen the issue templates mostly in bigger projects, but I don't see any reason why PDF Arranger should not have one.

dreua:
> I have seen the issue templates mostly in bigger projects [...]

OT: I think [we are growing](https://dreua.github.io/1616624000.html) :wink:

kbengs:
Sounds like a good idea.

jeromerobert:
Done for bug report: 16fba788aca408a39d5725f88ce038e446ed699f

I don't feel like making one for feature requests as it would be almost free form. I also wonder if the Github Issue Tracker is the best entry point for users who want to submit feature requests. May be we need a Gitter/Discord/Slack/Mattermost or whatever.

dreua:
Thank you! I think think we can close this issue for now and see how that works in practice. I'm not sure about additional channels, keeping it all in one place has its advantages imo. I've never used Gitter though, maybe I'd need to look into that first.

PR #473, opened 2021-04-25
First issue template for PDF Arranger

jeromerobert:


dreua:
In #447 I said I'd take care of that, but didn't come around to it yet. Thanks for taking the initiative!

Already looking good to me, I think we can merge this in any state and refine it as needed.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
```
dreua, line 10: After finding the "rich diff" button: I think there needs to be a second line break after each `**heading**`. You could also consider using `### heading` (or some other level) instead.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+A clear and concise description of what you expected to happen.
+
+**PDF files**
+Check if the bug appear only with some PDF. If the bug is specific to some PDF files please attach one to the issue.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**System and Versions**
+ - PDF-Arranger version and where you get it (distro package, flatpak, pip, git, download, ...)
```
dreua, line 30: I'd prefer "PDF Arranger" to be consistent. "got" could be used as past tense of "get" but the message is clear anyway.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+ - pikepdf version and where you get it (distro package, pip, git, ...)
+ - OS name and version (ex: Windows 10, Fedora 33, ...)
```
dreua, line 32: Are you using Fedora, too? Did I convince you with my constant advertising? :laughing:

dreua: I'd prefer `e.g.` over `ex:` but it's both alright with me. Picking details here, sorry.

jeromerobert: > Are you using Fedora, too? Did I convince you with my constant advertising?

Oh no, when I'm buried I'll still use Debian, but I knew you'd like that ;-)

angsch:
> **PDF files**
Check if the bug appear only with some PDF. If the bug is specific to some PDF files please attach one to the issue.

Having img2pdf in mind, what about generalizing _PDF files_ to _documents_?

PR #475, opened 2021-04-26
Fixes #449

### LINKED ISSUE #449: Page crop can exceed 100%
The check that crops of opposite sides sum to at most 99% can be bypassed. The export yields an unexpected PDF.

1. Open page format dialog, set left crop to 80, press OK.
2. Open page format dialog again, _type_ 80 into the right crop field, press OK.

The crop is accepted:
image

The exported PDF looks:
image

angsch:


kbengs:
Should it limit the value that you currently try to set too high instead? Now it will trim down the left value if you set the right value too high.

angsch:
> Should it limit the value that you currently try to set too high instead? Now it will trim down the left value if you set the right value too high.

When a user types in a value, I assume that this is really the value the user wants there.

kbengs:
Yeah it can be seen that way too. Can be merged imo.

jeromerobert:
Merged as 9f4fca0047a182fc0f35ed6f82bf25938fd90ccd (just added a commit title)

Issue #471, opened 2021-04-23
Suggestion: covering edges - instead of cropping them


untimished:
PDF Arranger is a real treat for cropping and rotating scans.
Covering the edges with white rectangles as an alternative to cropping the pages would be a great addition.
This would make it possible to cover unwanted black patches on single or multiple scans, while retaining the document size.
And cropping could be applied in the next step.

pikepdf should provide the necessary functions to draw the rectangles.
(I'm only having some issues with scaling and positioning)

Would this be a welcome feature?

dreua:
It's nothing that I personally would need or work on but I don't see a problem to merge a good PR. (No promises, I don't decide this on my own.)

Would you be interested to work on such a feature?

angsch:
This is a special case of the page extensions as proposed in #267. First crop a page, then add white margins. To support both positive crop (hide content) and negative crop (additional margins) on the same page, we probably need to introduce a CropBox.

untimished:
I tried using a smaller CropBox, but I get the same result like using a MediaBox.
This:
```
new_page.MediaBox = [0, 0, 500, 500]
new_page.CropBox = [100, 100, 400, 400]
```
produces the same result as this:
```
new_page.MediaBox = [100, 100, 400, 400]
```
(And using a TrimBox doesn't result in any visible changes.)

Positive and negative cropping the entire page works great, but I fail at cropping just the page content.
Am I missing something?
If covering up the content behind rectangles is the only working solution, that's what I'd try to implement in a PR.

dreua:
> I tried using a smaller CropBox, but I get the same result like using a MediaBox.

Is this true for both, the PDF Arranger preview (rendered by poppler) and viewing the resulting PDF in other viewers?

angsch:
Sorry, you are right. It should be a TrimBox, not a CropBox. It would be awesome if you could prepare a PR. Here is a working example that uses the TrimBox - perhaps it helps you find the issue.

```
import pikepdf

in_pdf = pikepdf.open('test.pdf')
page = in_pdf.pages[0]
mediabox = page.MediaBox
x0, y0, x1, y1 = [float(x) for x in mediabox]
width = x1 - x0
height = y1 - y0

out_pdf = pikepdf.Pdf.new()
new_page = out_pdf.copy_foreign(page)

# Crop left 20% and right 50%
new_page.TrimBox = [x0 + 0.2 * width, y0, x1 - 0.5 * width, y1]

[...]

PR #477, opened 2021-04-27
[For discussion / feedback] option to crop only content

untimished:
Option to clip the content instead of cropping the entire page.
The poppler preview displays the regular cropping method, but saving as pdf retains the page, clipping only the content.

Issue: #471

dialogue

Rotating pages causes issues and needs to be fixed.
This should be a draft. Excluding changing the internal preview means that this option can be implemented with very few changes.


angsch:
Looks promising. I don't know how to deal with the rendering issue and hope that someone else can comment on it.

### Crops are not rotated on rotated pages
The crops have to be rotated on export. 
1. Load a page and rotate it to the right
2. Set right crop to 20% and tick the checkbox
3. Export

The exported PDF has the wrong format.
image

### crop white borders
The current code for crop white borders is not compatible.
1. Load a page
2. Crop white borders

I get the error message:
`TypeError: crop() missing 1 required positional argument: 'clip'`
Perhaps we can make the clip argument optional?

### Select Same Page Format
This selection is based on the rendered result. Hence, real crop and the new feature are considered to have the same page format.

1. Load 2 pages.
2. Apply some crop to the first page
3. Apply the same crop to the second page and tick the checkbox.
4. Select Same Page Format

Both pages are selected although they do not have the same page format.

kbengs:
Yes, looks promising.
In is_data_valid the test len(tmp) == 9 need to be 10 now.

I did a test with the rendering. It's a start but there are more things to sort out.
```
--- a/pdfarranger/iconview.py
+++ b/pdfarranger/iconview.py
@@ -48,6 +48,11 @@ class CellRendererImage(Gtk.CellRenderer):
             return
 
         w0, h0, w1, h1, w2, h2, rotation = self.get_geometry()
+        if self.page.clip:
+            if w1 == w0:
+                w2, h2 = w0, h0
+            else:
+                w2, h2 = w1, h1
         th = int(2 * self.th1 + self.th2)
         w = w2 + th
         h = h2 + th
@@ -69,17 +74,37 @@ class CellRendererImage(Gtk.CellRenderer):
         window.rectangle(0, 0, w2 + 2 * self.th1, h2 + 2 * self.th1)
         window.fill()
 
-        # image
+        # white paper
         window.set_source_rgb(1, 1, 1)
         window.rectangle(self.th1, self.th1, w2, h2)
+        window.fill()
+
+        # image
+        window.set_source_rgb(1, 1, 1)
+        if self.page.clip:
+            window.translate(self.th1, self.th1)
+            a = self.page.crop[0] * w1
+            b = self.page.crop[2] * h1
+            c = w1 - self.page.crop[0] * w1 - self.page.crop[1] * w1
+            d = h1 - self.page.crop[2] * h1 - self.page.crop[3] * h1
+        else:
+            a = self.th1
+            b = self.th1
+            c = w2
+            d = h2
+        window.rectangle(a, b, c, d)
         window.fill_preserve()
         window.clip()
 
-        window.translate(self.th1, self.th1)
+        if not self.page.clip:
+            window.translate(self.th1, self.th1)
         scale = self.page.resample * self.page.zoom
         window.scale(scale, scale)
         x = int(self.page.crop[0] * w1)
         y = int(self.page.crop[2] * h1)
+        if self.page.clip:
+            x = 0
+            y = 0
         window.translate(-x, -y)
         if rotation > 0:
             window.translate(w1 / 2, h1 / 2)
```

pdfarranger/core.py
```diff
@@ -319,8 +321,9 @@ def move(self, treerowref, before):
         self.before = before
         self.treerowref = treerowref
 
-    def addpages(self, filename, page=-1, basename=None, angle=0, scale=1.0, crop=None):
+    def addpages(self, filename, page=-1, basename=None, angle=0, scale=1.0, crop=None, clip=None):
```
kbengs, line 324: Could you have clip=False here and skip the next line?

untimished:
Really appreciate the help, thank you for your comments and proposals.
@kbengs I will **focus on exporting first** and then use your diff for the **preview display later on**.

There is a **big remaining issue**, with **copying cropped/clipped files** - these get **cropped repeatedly or shrink**.

pdfarranger/exporter.py
```diff
@@ -147,13 +147,12 @@ def export(input_files, pages, file_out, mode, mdata):
         angle = row.angle
         angle0 = current_page.Rotate if '/Rotate' in current_page else 0
         new_page = pdf_output.copy_foreign(current_page)
-        #new_page.Rotate = 0
+        new_page.Rotate = 0
```
untimished, line 150: Setting new_page.Rotate = 0 and adding an optional angle argument to  _mediabox() in exporter.py is changing quite a lot and a bit of a hack.

kbengs:
> And copy_foreign Annotations throws Error:

This should be fixed with either of m-holgers PRs.

mara004:
As a design consideration, I think it might be better to make trimming pages a separate feature, for one might want to apply both crop and trim.

pdfarranger/exporter.py
```diff
@@ -146,9 +144,25 @@ def export(input_files, pages, file_out, mode, mdata):
...
-            new_page.Rotate = angle + angle0
-        new_page.MediaBox = _mediabox(new_page, row.crop)
+        new_page.Rotate = 0
+        mediabox = _mediabox(new_page, row.crop, angle + angle0)
+        if not row.clip:
+            new_page.MediaBox = mediabox
+        # clip content instead of cropping the page
+        else:
+            print(row.crop)
+            new_page.CropBox = _mediabox(new_page, row.crop, angle + angle0)
```
mara004, line 154: You could alternatively use `TrimBox`

kbengs:
"Hide margins" was implemented in https://github.com/pdfarranger/pdfarranger/pull/954

PR #480, opened 2021-04-30
Add symbolic icon

foreverxml:
[Issue regarding this](https://github.com/pdfarranger/pdfarranger/issues/)


dreua:
I will need to fix the rpm packaging but that is nothing you need to worry about.

PR #482, opened 2021-05-05
Fix scroll to selection &  Hide iconview when changing window width

kbengs:


jeromerobert:
Hiding the iconview make resizing smoother but for some reason, the iconview takes some time to be back. Sometime the iconview remains grey until I move the mouse on the window.

kbengs:
> Sometime the iconview remains grey until I move the mouse on the window.

I haven't been able to reproduce that. But I have done some small changes that should make iconview visible a bit faster. The mouse button release event and enter notify event are not emitted always, but for theese cases the timeout should make iconview visible again. @jeromerobert can you still see the problem?

It does take some time for iconview to get visible again after changing window width, but at least for me iconview is ready for interaction faster with this commit than before. For example on a 3670 page pdf, when resizing by double-clicking on headerbar iconview is ready for interaction in about 4 seconds, where it before took about 8 seconds to get ready. And resizing window by dragging window edge is almost impossible in main for such large pdf.

jeromerobert:
> It does take some time for iconview to get visible again after changing window width

As far as I can see, it's not "changing", but only increasing. Reducing the width is always fast. When I increase the width I often have a 4s delay before the iconview gets back. It does not depends on the size of the PDF file as I can reproduce it with a 10 pages PDF files.

I will take a closer look (i.e. add some trace) to understand why this is happening.

jeromerobert:
`enter_notify_event` is always triggered when decreasing width. `enter_notify_event` is sometime, but not always, triggered while increasing width.

jeromerobert:
I don't know if that can help but, as far as I can see, the only event which is always called with resizing end is `GDK_FOCUS_CHANGE` (i.e. `window_focus_in_out_event`).

jeromerobert:
Adding `self.set_iconview_visible(timeout=False)` to `window_focus_in_out_event` fix the issue. I wonder if we need to keep `set_iconview_visible` in `window_enter_notify_event` and `window_button_release_event`.

kbengs:
Okay thanks. On my machine ```window_focus_in_out_event``` is not triggered when changing width so it seems we need to have all three events.

jeromerobert:
When starting with the window already maximized the iconview remains grey. `self.iconview.get_columns()` is then -1 so it leads to an infinit loop in `get_visible_range2()`. `render()` should exit if `self.iconview.get_columns()<=0`.

When switching from maximize to unmaximize the iconview remains grey for a while. I'm trying to find the right additional events where to call `set_iconview_visible`.

kbengs:
> When starting with the window already maximized the iconview remains grey

I think a ```len(self.model) > 0``` addition will fix this case: 
```    
def window_configure_event(self, _window, event):
    """Handle window size and position changes."""
    if self.window_width_old not in [0, event.width] and len(self.model) > 0:
```

jeromerobert:
Calling `set_iconview_visible`on `window_state_event` fix the grey iconview effect when unmaximizing the window. I cannot find how to fix it when maximizing.

kbengs:
I can see that it sometimes takes 1,5s longer before iconview gets visible again (when timeout make it visible). This is for example when pressing maximizing and unmaximizing buttons on headerbar. I can't find a good way to fix it. What desktop environment do you see the longer delays on @jeromerobert ? Do you think this delay is acceptable?

jeromerobert:
It's with XFWM4/XFCE4. Well yes, it's acceptable like this.

PR #488, opened 2021-05-14
Fix alpha channel removal for LA images

mara004:
LA images were wrongly not converted to RGBA since they do not contain the "transparency" information (for whatever reason). Nevertheless, they have an alpha channel that needs to be removed prior to forwarding the file to img2pdf.
This patch fixes the problem by explicitly checking for LA mode.


jeromerobert:
Merged as 0a14f24a9d75d974fb3b6370688d9d92d04b382b

PR #493, opened 2021-05-26
Add Catalan translation

juxuanu:


dreua:
Testing with `LANG=ca_ES pdfarranger` because only ca does not work (only the document title is translated in that case and a Warning gets printed). Looks good to me, can be merged once my previous question is addressed.

juxuanu:
Hey! Catalan is the language, Valencian is a variation. The same way there are variations in Spanish for Mexico and Spain.

You can read more about it here: https://en.wikipedia.org/wiki/Valencian_language

dreua:
Thanks for the quick reply! The Valencian variation seems to be pretty outdated (lots of untranslated strings): Any recommendation how to deal with it? Should we remove it in favor of your translation, i.e. can it be expected that people speaking "ca@valencia" are better served with your "ca" translation that with their current incomplete one?

juxuanu:
They would understand it, yes, although the differences in some forms are very different. I'd guess there is nothing in my translation that any Valencian speaker would not understand.

dreua:
Alright, I'll leave this for now, ideally the valencian translation would be updated, maybe we can find someone to do that.

@jeromerobert  do you have a list of translators per language somehwere? You should add @juxuanu and maybe ping someone for ca@valencia.

Thanks again for the translation and the PR :+1:

Issue #381, opened 2021-01-09
Installation instructions for FreeBSD


grahamperrin:
Hi, do the lines at https://github.com/pdfarranger/pdfarranger#for-developers need something additional? 

I have PDF Arranger installed on my main computer with FreeBSD-CURRENT but testing on a clean machine (helloSystem 0C155) there's this: 

```text
FreeBSD% pwd
/home/liveuser
FreeBSD% ls
Desktop     xorg.conf.d
FreeBSD% mkdir Documents
FreeBSD% cd Documents 
FreeBSD% git clone https://github.com/pdfarranger/pdfarranger.git

Cloning into 'pdfarranger'...
remote: Enumerating objects: 29, done.
remote: Counting objects: 100% (29/29), done.
remote: Compressing objects: 100% (22/22), done.
remote: Total 2963 (delta 15), reused 11 (delta 7), pack-reused 2934
Receiving objects: 100% (2963/2963), 1.14 MiB | 1.05 MiB/s, done.
Resolving deltas: 100% (2065/2065), done.
FreeBSD% cd pdfarranger

FreeBSD% pip install -e .

Defaulting to user installation because normal site-packages is not writeable
Obtaining file:///usr/home/liveuser/Documents/pdfarranger
    ERROR: Command errored out with exit status 1:
     command: /usr/local/bin/python3.7 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/usr/home/liveuser/Documents/pdfarranger/setup.py'"'"'; __file__='"'"'/usr/home/liveuser/Documents/pdfarranger/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-zctcsy20            
         cwd: /usr/home/liveuser/Documents/pdfarranger/                                                        
    Complete output (5 lines):                                                                                 
    Traceback (most recent call last):                                                                         
      File "<string>", line 1, in <module>                                                                     
      File "/usr/home/liveuser/Documents/pdfarranger/setup.py", line 24, in <module>                           
        from DistUtilsExtra.command import (                                                                   
    ModuleNotFoundError: No module named 'DistUtilsExtra'                                                      
    ----------------------------------------                                                                   
[...]

PR #496, opened 2021-05-28
README.md: FreeBSD: alternative approach to source

grahamperrin:
Preparing for intended addition of PDF Arranger to the ports collection for FreeBSD. 

If all goes well: 

1. I'll remove PDF Arranger from <https://wiki.freebsd.org/WantedPorts>
2. await appearance of packages under <https://www.freshports.org/print/pdfarranger/#packages>
3. we can request a listing at <https://github.com/pdfarranger/pdfarranger/wiki/Binary-packages>.

### Background

* https://github.com/pdfarranger/pdfarranger/issues/381
* the port to OpenBSD, https://github.com/openbsd/ports/tree/master/print/pdfarranger
* https://forums.FreeBSD.org/threads/80375/post-514535

> … submit the port tonight.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,13 @@ sudo dnf install poppler-glib python3-distutils-extra python3-pip python3-gobjec
 
 **On FreeBSD**
 
+`cd /usr/ports/print/pdfarranger/ && sudo make install clean`
+
+* see https://www.freshports.org/print/pdfarranger/#add
+* no need for subsequent use of `pip3`
+
+– or: 
+
 ```
 sudo pkg install devel/gettext devel/py-gobject3 devel/py-pip devel/py-python-distutils-extra graphics/poppler-glib textproc/intltool textproc/py-pikepdf x11-toolkits/gtk30
 ```


jeromerobert:
Thank you, but this change might be neither necessary nor desirable. This section of the README is for installation using `pip` and as a user, not using system package manager.

When your package will reach freshports it will automatically appear in [Repology](https://repology.org/project/pdfarranger/packages) and thus in the [*Linux packages* section of the README](https://github.com/pdfarranger/pdfarranger#linux-packages) and in https://github.com/pdfarranger/pdfarranger/wiki/Binary-packages.

Well, now that we have FreeBSD and OpenBSD packages, the *Linux packages* section should probably be renamed to *Linux and BSD packages*.

grahamperrin:
Thanks for the clarification. 

A slightly grey area, in that `make install` (from source) is distinct from [pkg-install(8)](https://man.freebsd.org/pkg-install(8)), however the proposed edit to `README.md` _is_ a bit of a misfit, so let's close this. 

Expect the `make install` hint to appear at https://www.freshports.org/print/pdfarranger/#add in due course.

grahamperrin:
@dreua @jeromerobert FYI https://forums.FreeBSD.org/threads/80375/post-515904

PR #497, opened 2021-05-29
Use client side decorations in Windows

### LINKED ISSUE #297: Snap assist doesn't work in Windows 10
This is a gtk issue that we can't do anything about but I put the info here if somebody search for it.

Windows 10 has features that:
 * Resize windows when they are dragged to edges
 * Suggest companion windows when a window is dragged to left or right side
 * Minimize other windows when a window is shaken
 * Move windows with Win + Left or Right
 
This does not work for pdfarranger. The issue came with 1.5.0 when the application changed to GtkHeaderBar.
 
It is a open issue on gitlab:
https://gitlab.gnome.org/GNOME/gtk/-/issues/2785

Don't know if this issue should be left open or be closed.

kbengs:
Titlebar in dialogs will get a gnome look. The change is done because
it will enable window moving with Win + left/right.
Close #297 (the rest need to be fixed in gtk)

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -73,10 +73,12 @@ APPNAME = 'PDF Arranger'
 VERSION = '1.7.1'
 WEBSITE = 'https://github.com/pdfarranger/pdfarranger'
 
-# Add support for dnd to other instance and insert file at drop location in Windows
 if os.name == 'nt':
+    # Add support for dnd to other instance and insert file at drop location in Windows
     import keyboard  # to get control key state when drag to other instance
     os.environ['GDK_WIN32_USE_EXPERIMENTAL_OLE2_DND'] = 'true'
+    # Use client side decorations. Will also enable window moving with Win + left/right
+    os.environ['GTK_CSD'] = '1'
 
 import gi
 


angsch:
```diff
+# Use client side decorations. Will also enable window moving with Win + left/right
+os.environ['GTK_CSD'] = '1'
```

I am surprised that there is an effect. According to https://developer.gnome.org/gtk3/stable/gtk-running.html, the default value is 1

> GTK_CSD.  The default value of this environment variable is 1. If changed to 0, this disables the default use of client-side decorations on GTK+ windows [...].

and the proposal to make 0 the default is still open: https://gitlab.gnome.org/GNOME/gtk/-/issues/760. However, I don't have access to Windows right now to try the patch out.

jeromerobert:
I'll have access to a Windows box and time to test this PR during July.

jeromerobert:
@angsch You are right:

https://gitlab.gnome.org/GNOME/gtk/-/blob/d536fa5a/gtk/gtkwindow.c#L6129

but:

https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-gtk3/0003-gtkwindow-Don-t-force-enable-CSD-under-Windows.patch

Issue #438, opened 2021-03-21
Create Booklet from multiple pages and vice versa


javadevelopment4:
It would be great if someone builds or provide the following options

1) Convert A4 PDF to Booklet PDF
2) Convert Booklet PDF to A4 PDF 

Booklet to A4 PDF Example - https://www.sejda.com/split-pdf-down-the-middle
PDF to Booklet - https://www.bookletcreator.com/

dreua:
- We are not building such specific features as "A4 only". It has to be valuable to all users, not just you, therefore it must support different page sizes.
- Splitting is already supported, so all that is left would be an automated reordering of the booklet pages, correct?
- Joining pages is not supported at yet. If we are going to do it, this would be the first step and should be tracked as a first separate issue imo.
- Let's give this 3 weeks. If nothing productive comes up it should be closed and tagged `stale feature`.

dreua:
I'm not sure I understand those issues correctly, is this actually a duplicate of #246 ?

angsch:
> I'm not sure I understand those issues correctly, is this actually a duplicate of #246 ?

No, it is mostly about page reordering. You pile a bunch of landscape pages, fold them in the middle and want to read them as a book. I see two ways of realizing the booklet.
- Reorder the pages, export the PDF and use the printer options to print 2 pages onto 1.
- Reorder the pages, glue pairs of pages together and then print.

I assume that the second option is easier in practice due to duplex printing and sorting of the printed pages.

A dedicated feature for the second option is in my opinion a bit too specific, but I think a page reordering functionality could be useful. If you want to create a booklet using the second option, you would have to do two steps.

dreua:
That really sounds like something that would be more convenient in the printing dialog because the booklet PDF will be very useless in a normal PDF viewer. (Hard to read or check without printing it.)

dreua:
A general page reordering feature sounds good to me. This could replace/refine the paste interleaved functions, too. I think there were also requests for other paste interleave functions like every third or forth page. (Maybe in another issue tracker, could have been qpdf.)

nicos68:
I'm in the hunt for basic open source imposition software in the likes of [BookMaker](https://www.devalipi.com/moreinfobm.html), with crop marks being a welcome bonus. Your software looks like a solid basis and I could implement that on top of it, but I see you are questioning the direction the project is going to, so I'm hesitating between either just start from your project and go my own way or try to stick with your project and produce something you could integrate. What would be your advice ?

[...]

PR #500, opened 2021-06-26
Create a booklet from multiple pages

nicos68:
Merge and rearrange pages for booklet printing.

Issue:  #438

The action is available through a new entry in the edit dialog. The generated pages are sized using the source pages (adding widths and using max of height of the two pages) and cannot be sized manually (as this would involve rescaling the source pages).

Note that
- I didn't implement tests. If you want I can have a look to it.
- The behavior of the "cancel" action is not atomic, as I use some other actions (add, delete, insert)
- Rotated pages are not properly handled, I can also have a look to it if you wan.

This PR might be a bit premature, however there already is some working code and it doesn't break anything else, so I created it anyway, also because I would be happy to have feedback on my code. It's my very first time sending some (more or less) serious amount of code to an open source project, so sorry if things are a bit clumsy.

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -381,13 +381,14 @@ class PdfArranger(Gtk.Application):
         blank_page_count = 0 if len(pages) % 4 == 0 else 4 - len(pages) % 4
         if blank_page_count > 0:
             file = exporter.create_blank_page(self.tmp_dir, pages[0].size)
-            for _ in range(blank_page_count):
-                self._insert_pages(model, file, selection)
-                added_page_index = selection[-1].get_indices()[-1] + 1
-                # Fetch the additional blank pages and remove them from the model.
-                added_page = model.get_value(model.get_iter(added_page_index), 0)
-                pages.append(added_page)
-                model.remove(model.get_iter(added_page_index))
+            with self.model_lock:
+                for _ in range(blank_page_count):
+                    self._insert_pages(model, file, selection)
+                    added_page_index = selection[-1].get_indices()[-1] + 1
+                    # Fetch the additional blank pages and remove them from the model.
+                    added_page = model.get_value(model.get_iter(added_page_index), 0)
+                    pages.append(added_page)
+                    model.remove(model.get_iter(added_page_index))
 
         self.clear_selected()
 
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -334,6 +334,7 @@ class PdfArranger(Gtk.Application):
             ('select-same-format', self.on_action_select, 'i'),
             ('about', self.about_dialog),
             ("insert-blank-page", self.insert_blank_page),
+            ("generate-booklet", self.generate_booklet),
         ])
 
         main_menu = self.uiXML.get_object("main_menu_button")
@@ -361,6 +362,9 @@ class PdfArranger(Gtk.Application):
                 adder.move(Gtk.TreeRowReference.new(model, selection[-1]), False)
             adder.addpages(exporter.create_blank_page(self.tmp_dir, page_size))
             adder.commit(select_added=False, add_to_undomanager=True)
+
+    def generate_booklet(self, _, __, ___):
+        print('hello')
 
     @staticmethod
     def __create_filters(file_type_list):
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -357,14 +357,22 @@ class PdfArranger(Gtk.Application):
             size = model[selection[-1]][0].size_in_points()
         page_size = croputils.BlankPageDialog(size, self.window).run_get()
         if page_size is not None:
-            adder = PageAdder(self)
-            if len(selection) > 0:
-                adder.move(Gtk.TreeRowReference.new(model, selection[-1]), False)
-            adder.addpages(exporter.create_blank_page(self.tmp_dir, page_size))
-            adder.commit(select_added=False, add_to_undomanager=True)
+            self._insert_page(model, page_size, selection)
+
+    def _insert_page(self, model, page_size, selection):
+        adder = PageAdder(self)
+        if len(selection) > 0:
+            adder.move(Gtk.TreeRowReference.new(model, selection[-1]), False)
+        adder.addpages(exporter.create_blank_page(self.tmp_dir, page_size))
+        adder.commit(select_added=False, add_to_undomanager=True)
 
     def generate_booklet(self, _, __, ___):
-        print('hello')
+        model = self.iconview.get_model()
+        last_page_size = model[-1][0].size_in_points()
+        first_page_size = model[0][0].size_in_points()
+        page_size = [max(last_page_size[0], first_page_size[0])*2,
+                     max(last_page_size[1], first_page_size[1])]
+        self._insert_page(model, page_size, [])
 
     @staticmethod
     def __create_filters(file_type_list):
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -382,7 +382,7 @@ class PdfArranger(Gtk.Application):
         if len(pages) % 2 != 0:
             self._insert_page(model, pages[0].size, selection)
             if len(selection) < 1:
-                pass
+                pages.append(model[-1][0])
             else:
                 added_page_index = selection[-1].get_indices()[-1] + 1
                 added_page = model.get_value(model.get_iter(added_page_index), 0)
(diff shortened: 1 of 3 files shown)


jeromerobert:
Thank you !

> I didn't implement tests. If you want I can have a look to it.

Not top priority IMHO, this can be done in an other PR.

> The behavior of the "cancel" action is not atomic, as I use some other actions (add, delete, insert)

"cancel" ? Do you mean "undo" ? If yes this deserve to be fixed in this PR. Your action should start with `undomanager.commit("Create booklet")` and do not modify the `undomanager` later on. For the `PageAdder` you can use the `add_to_undomanager` argument.

As `clear_all` is not a end user action it should not update the `undomanager`. I'll need to look at it closer but I'm not even sure you need a `clear_all` function. `model.clear()` should be enough.

> Rotated pages are not properly handled, I can also have a look to it if you wan.

Not top priority IMHO. This could be fixed in an other PR.

> This PR might be a bit premature, however there already is some working code

Yes and yes. Could you rebase your branch on `origin/main` and squash/reorganize your commits so your PR show what you modify in the end and not your whole development history (It's a bit difficult to read as is). May be you could just have:
- commits which add utilities functions like ` make_tmp_file` or `_insert_page` and modify existing code
- one big commit for the feature you add

>  sorry if things are a bit clumsy

They are not. You are almost there.

jeromerobert:
Some qpdf/pikepdf version do not accept `pikepdf.Dictionary` as pages. See #427 and #435 for workaround.

jeromerobert:
Could you rebase your branch on origin/main and squash/reorganize your commits so your PR show what you modify in the end and not your whole development history (It's a bit difficult to read as is).

nicos68:
> Could you rebase your branch on origin/main and squash/reorganize your commits so your PR show what you modify in the end and not your whole development history (It's a bit difficult to read as is).

Sorry I wasn't finished, I didn't think you'd get notified by my push. I still had to handle the pikepdf version issue.

About my commits, I tend to commit very often and I'm used to Gitlab where we usually squash all the commits into one when the code is merged. Can you do that ? About the readability, the "files changed" tab looks like it should allow you to review my code easily.

jeromerobert:
Ok that sounds good.

nicos68:
Any idea when you'll merge this ?

angsch:
@nicos68 Sorry that nobody has responded to your PR for so long. I think I also speak on behalf of the others when I say that it was not clear that you consider the feature complete. Unless someone else is faster, I will have a look at the code tomorrow.

nicos68:
I do consider it complete.

angsch:
Hello nicos68,

Thank you for your PR. I feel really sorry that you had to wait for so long. 

Could you please check one more time that the result is correct when you have 6 pages? I have never used booklets and don't know what is correct. However, I expected that the content is consecutive when folded. In that case, when you fold the 6 pages, don't you need two blank pages in the booklet? [b | 1] [2 | b] [6 | 3] [4 | 5]

I ask you to change two things:

- Please add the functionality to the right-click menu. It is really hidden in the hamburger menu and unless you search for it, you will likely not find it. Most actions are duplicated and this one should be too.
- Please ensure that at least one page is selected (I guess 2 pages makes most sense, but this is up to you) to enable your feature. This is for two reasons. 1) Selecting all pages on the iconview as a default is not consistent with the behaviour of the other actions. 2) You can create a booklet without having a single page on the iconview, which triggers a crash. 

If this is included, I would merge the PR in.

nicos68:
You are absolutely right, my algorithm was flawed, I fixed it. I also made the changes you requested.

pdfarranger/pdfarranger.py
```diff
@@ -378,12 +378,14 @@ def generate_booklet(self, _, __, ___):
...
-            pages.append(added_page)
-            model.remove(model.get_iter(added_page_index))
+        if blank_page_count > 0:
+            file = exporter.create_blank_page(self.tmp_dir, pages[0].size)
+            for i in range(blank_page_count):
+                self._insert_pages(model, file, selection)
+                added_page_index = selection[-1].get_indices()[-1] + 1
+                added_page = model.get_value(model.get_iter(added_page_index), 0)
+                pages.append(added_page)
+                model.remove(model.get_iter(added_page_index))
```
kbengs, line 390: I could be wrong but I think a ```self.model_lock()``` is needed before any model.remove, and a ```self.model_unlock()``` after. You could also put the lock before the for loop, and unlock somewhere after the loop. (I think there is a small chance that a page could get wrong thumbnail without theese locks.) Other than that I think it looks good.

nicos68: Done, using the context manager.

angsch:
@nicos68 Great work! Thank you very much.

mara004:
I've just tested this interesting new feature (checked out from master branch) and noticed a few issues:
- the created booklet pages are appended to the end of the thumbnail list, which is confusing - I would have expected them to show up at the same position as the input
- sometimes the "generate booklet" command doesn't work - in the command-line, I get the following Exception:
```
  File "/home/manuel/Downloads/builds/pdfarranger/pdfarranger/pdfarranger.py", line 402, in generate_booklet
    with self.model_lock:
AttributeError: __enter__
```
Apparently you have a context manager that doesn't implement the `__enter__` method?

nicos68:
> I've just tested this interesting new feature (checked out from master branch) and noticed a few issues:
> 
>     * the created booklet pages are appended to the end of the thumbnail list, which is confusing - I would have expected them to show up at the same position as the input
> 
>     * sometimes the "generate booklet" command doesn't work - in the command-line, I get the following Exception:
> 
> 
> ```
> Traceback (most recent call last):
>   File "/home/manuel/Downloads/builds/pdfarranger/pdfarranger/pdfarranger.py", line 402, in generate_booklet
>     with self.model_lock:
> AttributeError: __enter__
> ```
> 
> From looking at the code, this is because `model_lock` is a method and not a context manager, so the false `with` block should be replaced by a pair of `self.model_lock()` `self.model_unlock()`.
> 
> ```python
> self.model_lock()
> for _ in range(blank_page_count):
>     self._insert_pages(model, file, selection)
>     added_page_index = selection[-1].get_indices()[-1] + 1
>     # Fetch the additional blank pages and remove them from the model.
>     added_page = model.get_value(model.get_iter(added_page_index), 0)
>     pages.append(added_page)
>     model.remove(model.get_iter(added_page_index))
> self.model_unlock()
> ```
> 
> However, with the code modified as shown above, the UI sometimes freezes and never unlocks again when triggering "generate booklet". Maybe a model lock/unlock is inappropriate in this place after all? It seems to work when I just comment it out.

Same for me. I've never had any problems before adding the lock. I'm okay to remove it but if more experienced maintainers of the project could give their feedback on this I'd be more comfortable.

mara004:
> Also, I used the context manager because [...]

It is correct that a context manager would be possible/appropriate in theory. However, you certainly cannot use a method like a context manager. You would have to create a class that implements `__init__(self, lock)`, `__enter__(self)` and `__exit__(self, exc_type, exc_value, exc_traceback)`.

> [...] if something crashes between lock and unlock then the app stays locked.

This is equally true for use without context manager. In this case there is no difference, it's just a question of which way of notation you prefer. Context managers can be a bit easier to use, since they safely prevent missing unlock calls.

mara004:
> Also, I used the context manager because [...]

It is correct that a context manager would be possible/appropriate in theory. However, you certainly cannot use a method like a context manager. You would have to create a class that implements `__init__(self, lock)`, `__enter__(self)` and `__exit__(self, exc_type, exc_value, exc_traceback)`.

> as if something crashes between lock and unlock then the app stays locked.

Yes, good point.

angsch:
Thank you for pointing out the problem with the rendering. That as well as the missing cases with cropped pages and rotations should be taken up. An updated TODO-list can be found in #438 - if something is missing, please add it.

mara004:
> Improve rendering

Primary points:
- fix the `with` blunder
- investigate possible exceptions or infinite waits inside the `for` loop (It is really suspicious that wrapping the loop with model lock/unlock causes pdfarranger to freeze. The unlock is never called, so the loop code either crashes before finishing or takes forever.)

After this visuals can be dealt with, like positioning the new booklet pages correctly.

kbengs:
The idea with the model lock is that if python interpreter decides to switch to the other thread between "path" and "ref" lines in core.py, (see below, at read arrow) and then a page is removed from the model in main thread, then path and ref will not point to the same page. I am not sure if that could lead to a problem. If you think it can not cause a problem, feel free to remove the lock.

image

mara004:
I am not experienced in threading, but would vote for keeping the lock to be safe.
However, we need to investigate what is wrong with the `for` loop that causes pdfarranger to freeze forever.

jeromerobert:
Note that there are 2 `model_lock`:

- `PdfArranger.model_lock` which is method and must not be as a `with` statement.
- `PDFRenderer.model_lock` which is a `threading.Lock` (can be used as a `with` statement)

I agree with mara004, that `PdfArranger.model_lock` should be replaced by a context manager and all `PdfArranger.model_lock`/`PdfArranger.model_unlock` pair replaced by a `with` statements. I also think one of those 2 `model_lock` should be renamed to avoid further confusions.

jeromerobert:
I'm on it.

PR #503, opened 2021-06-30
Show right-click menu also when clicking on empty area

kbengs:
I had this change in a old branch. :-)


dreua:
Great, thanks for the super fast response 💯

PR #510, opened 2021-07-15
Improve closing dialog with proper <Esc> functionality

Scafir:
Hi, it's me again,
I noticed that when hitting \<Esc\> while on the "Do you want to save before quitting?" dialog, the whole program would just quit without saving. This does not feel natural, as on any other software (e.g: LibreOffice), this would act as the "Cancel" button.
This commit fixes this.
In addition, if the return code is unknown, we act as if we pressed "Cancel" as a safeguard instead of quitting without saving.

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -826,7 +826,11 @@ class PdfArranger(Gtk.Application):
             response = d.run()
             d.destroy()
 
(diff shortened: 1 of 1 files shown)


Scafir:
I really need to run the tests local before doing a PR, sorry for the spam. I will fix this shortly.

PR #522, opened 2021-08-29
Booklet

jeromerobert:
Implement https://github.com/pdfarranger/pdfarranger/pull/500#issuecomment-898943898 and fix a fix a dead lock in booklet creation.


mara004:
Yes, that works now. Thanks!

kbengs:
Yes seems to work. Looks good imo.

PR #531, opened 2021-09-09
Use normal clipboard in Linux

kbengs:
Could a wayland user test if this make copy-pasting work? (it does work in a virtual machine..)


kbengs:
Tested in Gnome and KDE with wayland

mara004:
Looks good, I'll check this out and test it soon.

mara004:
Yep, I can confirm it works. Great.

jeromerobert:
Great, it also simplifies the code

PR #532, opened 2021-09-11
Simplified rendering status

kbengs:
I didn't like how rendering status hide "selected pages" text. So here is another approach. Cpu usage is noticeably lower also.


jeromerobert:
Great ! I'll add a commit to fix the broken test then I'll merge.

jeromerobert:
Replaced by #535 and merged as 6e52c53efbafdeb13fd18d2da1c4a3b4e73b3fdc

PR #542, opened 2021-10-03
Move export pdf to separate thread

kbengs:
Before I do more work on this I could have your opinion about it. :-)

When saving a larger pdf the app will freeze until file is saved. The save dialog can stay open for a long time when you press save. App can not be quited while saving. No "Saving" info is provided to user while saving. So here I try to do something about these issues. We can wait until 1.9.0 if you think it is to much for 1.8.0.

### DIFF
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -18,6 +18,8 @@
 import pikepdf
 import os
 import traceback
+import sys
+import warnings
 import tempfile
 from . import metadata
 from gi.repository import Gtk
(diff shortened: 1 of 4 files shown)


kbengs:
```remove_unreferenced_resources``` does not release the GIL while it runs meaning application window will freeze even if it is run in a separate thread. I am now experimenting with multiprocessing. It seems to work fine in Linux, but in Windows it seems to be much harder.

Issue/question at pikepdf:
https://github.com/pikepdf/pikepdf/issues/263

kbengs:
Dogtail test obviously still fails but other than that it work as far as I can see. The content loss warning check still freeze the app for a while when it runs. I'd say this dialog could be shown always regardless of if there is forms or outlines in the document. Also maybe the dialog could show up when the app start.

jeromerobert:
Merged as 7f81ffbbdc378d643ef64ad605fe721852812080 and aca0d65e715cf1de614271532ae176bf2fc8949a. Dogtail fix: a45c2ce3b09057951b894acb824a02bbab9b6063

PR #544, opened 2021-10-11
Enable keyboard navigation through the title bar

vvillenave:
This also enables to leave the Icon View through  
either Tab or Shift-Tab, while blocking other   
key events.  (I’m not exactly sure why “propagation”  
is not something we want here, but at least this  
will be the most minimal & non-disruptive change.)

### DIFF
--- a/data/pdfarranger.ui
+++ b/data/pdfarranger.ui
@@ -21,23 +21,19 @@ along with PDF Arranger.  If not, see <http://www.gnu.org/licenses/>.
 <interface>
   <requires lib="gtk+" version="3.12"/>
   <object class="GtkApplicationWindow" id="main_window">
-    <property name="can_focus">False</property>
     <property name="show_menubar">False</property>
     <child type="titlebar">
       <object class="GtkHeaderBar" id="header_bar">
         <property name="visible">True</property>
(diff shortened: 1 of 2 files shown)


dreua:
> I’m not exactly sure why “propagation”
> is not something we want here

I'm too tired to investigate on this right now, but if there is no good reason to prevent propagation we shouldn't do it and go with the simpler code. I'm not a big fan of accumulating "minimal & non-disruptive" changes just to end up with code that no one can understand because it never had a good reason to exist in the first place.

vvillenave:
> I'm not a big fan of accumulating "minimal & non-disruptive" changes just to end up with code that no one can understand because it never had a good reason to exist in the first place.

Hear, hear! I wouldn’t have dared alter this unless someone more knowledgeable  
than me would chime in first, but I certainly agree with you in general.

From what I can see, that logic was introduced by a03ac42c37aa77d2103eb663d089850544c4bcf4 with the following comment:  
`# Prevent propagation ie don't set cursor at first or last item`

Perhaps @kbengs could elaborate on that?

kbengs:
Looks good imo too. I don't see any issues when testing it.

PR #564, opened 2021-11-01
Fix zoom scale issues

kbengs:
Steps to reproduce the issue:
* Open a PDF and zoom to full page
* Select all + Cut
* Paste
-> The pages will not receive a thumbnail. Thumbnail is rendered but update_thumbnail does not accept the page.


kbengs:
To reproduce second issue:
* Open a PDF
* Make window height as small as possible
* Zoom to full page
* Select all + cut
* Paste
-> cairo.Error: invalid value.

PR #565, opened 2021-11-01
Update Pt_BR.po

TulioCaesar:
Update Pt_BR.po according to POT file

### DIFF
--- a/po/pt_BR.po
+++ b/po/pt_BR.po
@@ -1,141 +1,142 @@
-# Brazilian portuguese translation for pdfarranger.
+# Brazilian portuguese translation for PDF Arranger.
 # Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
 # This file is distributed under the same license as the PACKAGE package.
 # Rogênio Belém <rogeniobelem@gmail.com>, 2013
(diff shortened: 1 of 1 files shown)


po/pt_BR.po
```diff
@@ -1,141 +1,142 @@
...
 
-#: pdfarranger/core.py:186
+#: pdfarranger/core.py:204
 msgid "The password will be remembered until you close PDF Arranger."
 msgstr "A senha será lembrada até que você feche o PDF Arranger."
 
-#: pdfarranger/core.py:187
+#: pdfarranger/core.py:205
 msgid ""
+#, python-format
```
dreua, line 35: According to CI this is an error. Try moving it up one line, so it doesn't stand between msgid and the following string. Did you write this by hand or was this error introduced by some tooling?

dreua (changes requested):

po/pt_BR.po
TulioCaesar: Sorry, I made a mistake while correcting this line. I accidentally reversed the lines.

dreua: Don't be sorry, I just wanted to know whether that was a bug somwhere else that needs to be fixed. Thanks for your translations & PR!

dreua:
Now everything is green :tada:

PR #580, opened 2021-11-17
Change config.ini file location in Windows portable

kbengs:


kbengs:
This should be working now. 1.9.0 is fine.

PR #587, opened 2021-11-21
Fix a button location issue in Insert Blank Page dialog

kbengs:
In KDE minimize and close buttons are too far to the left
insert blank page
Before the commit to left, after to right


dreua:
I guess it was never correct to set those after the call to show_all

PR #600, opened 2021-12-20
New non-regression tests and bug fix for page scaling

jeromerobert:
* Add a new test for page scaling and rotate which check the page size
* Add a non-regression test for 5e58d3e (Fix geometric transformation on duplicated pages)
* Check both PikePDF 1.19 and 4.2.0


kbengs (commented):
Looks good. Seems to work. :+1:

Issue #604, opened 2022-01-02
ModuleNotFoundError: No module named 'DistUtilsExtra'


hongyi-zhao:
On Ubuntu 20.04.3 LTS, I install `pdfarranger` according to the guidance [here](https://github.com/pdfarranger/pdfarranger#for-developers) using the following steps:

```
$ git clone https://github.com/pdfarranger/pdfarranger.git
$ pyenv shell datasci
$ python --version
Python 3.9.1
$ cd pdfarranger
$ ./setup.py build
Traceback (most recent call last):
  File "/home/werner/Public/repo/github.com/pdfarranger/pdfarranger.git/./setup.py", line 24, in <module>
    from DistUtilsExtra.command import (
ModuleNotFoundError: No module named 'DistUtilsExtra'
```
Then I try the following but still failed:
```
$ pip install distutils-extra
Looking in indexes: https://pypi.tuna.tsinghua.edu.cn/simple
ERROR: Could not find a version that satisfies the requirement distutils-extra (from versions: none)
ERROR: No matching distribution found for distutils-extra
```
Any hints for fixing this problem? See [here](https://discuss.python.org/t/modulenotfounderror-no-module-named-distutilsextra/12834) for the relevant discussion.

Regards,
HZ

kbengs:
Did you run this line?
```
sudo apt-get install python3-pip python3-distutils-extra python3-wheel python3-gi python3-gi-cairo gir1.2-gtk-3.0 gir1.2-poppler-0.18 python3-setuptools
```

You maybe don't need all of those but at least python3-distutils-extra

hongyi-zhao:
Yep, as shown below:
```
$ for i in python3-pip python3-distutils-extra python3-wheel python3-gi python3-gi-cairo gir1.2-gtk-3.0 gir1.2-poppler-0.18 python3-setuptools; do dpkg --get-selections $i; done
python3-pip					install
python3-distutils-extra				install
python3-wheel					install
python3-gi					install
python3-gi-cairo				install
gir1.2-gtk-3.0:amd64				install
gir1.2-poppler-0.18:amd64			install
python3-setuptools				install
```

dreua:
I'm assuming the pyenv command creates an empty environment for you. I'm not familiar with this so I can't help you any further unfortunately. I guess you'll have to find a way to install / display the system package inside the pyenv. Let us know if you find a solution.

hongyi-zhao:
[...]

PR #605, opened 2022-01-03
Instruct developers to install using pip

dreua:
pip allows you to uninstall while ./setup.py doesn't.

As suggested in https://github.com/pdfarranger/pdfarranger/issues/604#issuecomment-1003955304 - I've been using pip install for a long time and didn't have issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 __pycache__
 venv/
+pdfarranger.egg-info/
+build/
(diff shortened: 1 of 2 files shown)


dreua:
Only possible issue that may come up in the future:

>   DEPRECATION: A future pip version will change local packages to be built in-place without first copying to a temporary directory. We recommend you use --use-feature=in-tree-build to test your packages with this new behavior before it becomes the default.
>    pip 21.3 will remove support for this functionality. You can find discussion regarding this at https://github.com/pypa/pip/issues/7555.

When I do `pip3 install --user . --use-feature=in-tree-build` it still works but modifies `po/pdfarranger.pot`. Does this also happen with setup.py? Do you think this will be an issue?

jeromerobert:
I disagree. See https://github.com/pdfarranger/pdfarranger/issues/604#issuecomment-1004080685

PR #614, opened 2022-01-15
Do not set the init current folder of import file chooser

### LINKED ISSUE #508: Default directory in file chooser
When a document is imported with the file chooser dialog, the path is stored. The next time you open the file chooser, you start in the very same directory.

**Question 1:** The path is stored on a per window basis. When you open a new window (ctrl + n), the path is reset to the default. Should the new window be treated in the same way as a freshly launched instance?

**Question 2:** The default directory is the home directory. Should PDF Arranger default to a path one level deeper into the file tree? On my system, evince defaults to _Documents_, gimp to _recently used_, shutter to _Pictures_. An alternative is a customisation option to set the default path similar to the keyboard shortcuts.

jeromerobert:
* In that case GTK default to "recent" or "cwd" depending on
  gsettings get org.gtk.Settings.FileChooser startup-mode.
  "recent" (the default) open the "Recent" tab of the dialog
* Subsequent import file chooser still use
  self.import_directory
* Fix #508

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -245,7 +245,7 @@ class PdfArranger(Gtk.Application):
         self.target_is_intern = True
 
         self.export_directory = os.path.expanduser('~')
-        self.import_directory = self.export_directory
+        self.import_directory = None
         self.nfile = 0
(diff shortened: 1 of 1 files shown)


kbengs:
:+1:

PR #619, opened 2022-01-22
Show GTK and Python version in About

jeromerobert:
Because we now have a bug which depends on the GTK version.


kbengs:
Looks good

PR #620, opened 2022-01-23
white_border: Use memoryview to speed-up loops

jeromerobert:


kbengs:
Seems to work and looks good imo.

PR #621, opened 2022-01-28
ui: port to libhandy

oscfdezdz:
GtkApplicationWindow is now a HdyApplicationWindow and GtkHeaderBar is now a HdyHeaderBar, a GtkBox is needed now between HdyApplicationWindow and the rest of the UI elements.

This change enables bottom rounded corners (at least on Linux) and will also allow PDF Arranger to follow system theme (Linux) with libhandy 1.5 ([docs](https://gnome.pages.gitlab.gnome.org/libhandy/doc/main/HdyStyleManager.html)). That's not implemented here as libhandy 1.5 is not out yet.

I haven't noticed any regressions, although it was only tested on Linux.

### DIFF
--- a/data/pdfarranger.ui
+++ b/data/pdfarranger.ui
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.22.2 -->
 <!--
 Copyright (C) 2020 Jerome Robert
 
@@ -19,13 +18,15 @@ You should have received a copy of the GNU General Public License
 along with PDF Arranger.  If not, see <http://www.gnu.org/licenses/>.
 -->
 <interface>
-  <requires lib="gtk+" version="3.12"/>
-  <object class="GtkApplicationWindow" id="main_window">
+  <object class="HdyApplicationWindow" id="main_window">
     <property name="show_menubar">False</property>
-    <child type="titlebar">
-      <object class="GtkHeaderBar" id="header_bar">
+    <child>
+      <object class="GtkBox">
         <property name="visible">True</property>
-        <property name="spacing">4</property>
+        <property name="orientation">vertical</property>
+        <child>
+          <object class="HdyHeaderBar" id="header_bar">
+        <property name="visible">True</property>
         <property name="show_close_button">True</property>
         <child>
           <object class="GtkBox">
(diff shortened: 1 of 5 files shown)


jeromerobert:
Could you clean your `pdfarranger.ui` patch ? I guess you did not actually changed every lines.

oscfdezdz:
> Could you clean your `pdfarranger.ui` patch ? I guess you did not actually changed every lines.

I guess the diff shows it as changed because the indentation did change for almost the entire file.

data/pdfarranger.ui
```diff
@@ -19,318 +18,323 @@ You should have received a copy of the GNU General Public License
...
+  <object class="HdyApplicationWindow" id="main_window">
     <property name="show_menubar">False</property>
-    <child type="titlebar">
-      <object class="GtkHeaderBar" id="header_bar">
+    <child>
+      <object class="GtkBox">
         <property name="visible">True</property>
-        <property name="spacing">4</property>
-        <property name="show_close_button">True</property>
+        <property name="orientation">vertical</property>
```
oscfdezdz, line 26: This new GtkBox causes the indentation for the rest of the file to increase. Locally the diff is the same, so I don't think I can do much more.

kbengs:
I am not familiar with libhandy but I read a bit about it. I did test the commit in Kubuntu (had to install libhandy-1-dev) and I could see no visual changes. :wink: @oscfdezdz what would you say is the benefit of using libhandy?

oscfdezdz:
The changes made in this PR are mostly an inizialization of libhandy, the benefits it could bring are: making the application more adaptive for smaller form factors and/or modernizing it thanks to the widgets its library has, facilitating a port to GTK4 + libadwaita or, as stated in the PR description, support for the upcoming light/dark system theme to GNOME, KDE or Pantheon.

dreua:
Does this introduce new dependencies we need take care of / communicate to downstream package maintainers? And are those available on the relevant distributions?

jeromerobert:
> Does this introduce new dependencies we need take care of / communicate to downstream package maintainers?

Possibly yes. It depends on distributions tracks dependencies. On some distributions `libhandy` could already be required by some of our existing dependencies and not required by others. Contributors will have to provide updates to the README.

> And are those available on the relevant distributions?

Yes. libhandy is already used by Gnome.

dreua:
I think we need to update Readme to include the new dependency.

oscfdezdz:
libhandy 1.6 is out and the following code makes pdfarranger follow dark mode:

```
style_manager = Handy.StyleManager.get_default()
style_manager.set_color_scheme(Handy.ColorScheme.PREFER_LIGHT)
```

I have added it in `do_activate` and it works fine with the Flatpak manifest and GNOME master runtime (which will become version 42) but I am not sure how to check libhandy's version for cases where the available version is older than needed.

PR #622, opened 2022-01-29
Fix multiprocessing coverage issue

jeromerobert:
Without this change everything executed in multiprocessing is not
reported in coverage.


jeromerobert:
@kbengs It looks like there is no way to run python coverage on a process which mix threading and multiprocessing. I'll look for a fix but so far, our coverage is broken.

PR #629, opened 2022-02-03
Zoom enhancements

kbengs:


pdfarranger/pdfarranger.py
```diff
@@ -1952,29 +1941,28 @@ def sw_scroll_event(self, _scrolledwindow, event):
...
         return True
 
+    def update_max_zoom_level(self):
+        """Update upper zoom level limit so thumbnails are max 6000000 pixels."""
+        if len(self.model) == 0:
+            return
+        max_pixels = 6000000  # 6000000 pixels * 4 byte/pixel -> 23Mb
+        max_page_size = max(p.size[0] * p.size[1] * p.scale ** 2 for p, _ in self.model)
+        max_zoom_scale = (max_pixels / max_page_size) ** .5
+        self.zoom_level_limits[1] = min(int(log(max_zoom_scale / .2) / log(1.1)), 40)
```
dreua, line 1957: Can you explain what this line does? How did you come up with it? I'm half curious, half worried about the maintainability of our codebase ;)

kbengs: Yeah, I guess you are wondering most about the last line, but:
> max_pixels = 6000000  # 6000000 pixels * 4 byte/pixel -> 23Mb

23 Mb limit tries to limit memory usage to about 500Mb on a 4K monitor.
On a 4K monitor you can roughly have two A4 portrait pages zoomed to 6000000 pixels between each other.
The current implementation of page rendering will load 5 rows before visible range and 5 after. So (2 + 2  * 5 + 2 * 5)pages x 23Mb = 506Mb

> max_page_size = max(p.size[0] * p.size[1] * p.scale ** 2 for p, _ in self.model)
   max_zoom_scale = (max_pixels / max_page_size) ** .5

^2 and ^0.5 is because scale is multiplicated on both width and height

> self.zoom_level_limits[1] = min(int(log(max_zoom_scale / .2) / log(1.1)), 40)

Level is calculated from zoom_scale = 0.2 * (1.1 ** level) (as can be found elsewhere in the app also) with this math:
scale = 0.2 * 1.1^level
1.1^level = scale / 0.2
log(1.1^level) = log(scale / 0.2)
level * log(1.1) = log(scale / 0.2)  # possible according to google ;)
level = log(scale / 0.2) / log(1.1)

40 is the old max limit. It should not be needed but it's still there.

And reason why the limitation is done is:
1. It will avoid "cairo.Error: invalid value (typically too big)" (I think this will happen at 2^32 pixels)
2. App responsiveness
3. Limit memory usage

PR #630, opened 2022-02-04
Refactor code for page width-height swap on rotate

kbengs:


pdfarranger/core.py
```diff
@@ -140,14 +143,6 @@ def duplicate(self, incl_thumbnail=True):
             r.preview = None
         return r
 
-    def set_size(self, size):
-        """set this page size from the Poppler page."""
-        self.size = list(size)
-        rotation = int(self.angle) % 360
-        rotation = round(rotation / 90) * 90
```
dreua, line 147: Are we sure we don't need this rounding any more? Maybe we could consider an assert(angle in [0, 90, 180, 270]) to be sure the values are within the expected range?

kbengs: The rounding would in that case be on line 126. But documentation for modulo operator say that the result will be a integer if both a and b in a % b are integers. Data that is pasted is also checked that it is in [0, 90, 180, 270] So imo this should be safe.

dreua: I'm worried that it may round to 89 or 91 which would break the rest of your logic

PR #643, opened 2022-03-07
Show save warnings in a more informative dialog

kbengs:
Messages like `The metadata field /Trapped with value 'pikepdf.Name("/False")' has no XMP equivalent, so it was discarded` seems to be common. I think users will be confused when this pops up. But I am not sure what the text should be here. Anyone got ideas? Should there be the option to not show warnings? At least I have never seen any issues with the PDF when there has been warnings.

The message is also put in a scrolled window here so the dialog cant grow large.

Current:
image

The PR:
image
image

Idea 2:
image

### DIFF
--- a/pdfarranger/config.py
+++ b/pdfarranger/config.py
@@ -145,6 +145,12 @@ class Config(object):
     def set_content_loss_warning(self, enabled):
         self.data.set('preferences', 'content-loss-warning', str(enabled))
 
+    def show_save_warnings(self):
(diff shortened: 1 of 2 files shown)


jeromerobert:
I think both are good (i.e. better than the current one) but I prefer the second one.

kbengs:
> I think both are good (i.e. better than the current one) but I prefer the second one.

Ok, ready

PR #655, opened 2022-04-02
Make libhandy optional and remove some deprecated GTK3 usage

jeromerobert:


kbengs:
I see no issues with it. I agree it's better this way.

PR #658, opened 2022-04-09
Allow to print the document

### LINKED ISSUE #647: Add printing support
For productivity reasons, please add printing support directly out of the application. It will save a lot of use-time if you don't need to use an external application just for printing.

jeromerobert:
Close #647

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -243,6 +243,9 @@ class PdfArranger(Gtk.Application):
         self.set_iv_visible_id = None
         self.vadj_percent = None
         self.end_rubberbanding = False
+        self.disable_quit = False
+        multiprocessing.set_start_method('spawn')
+        self.quit_flag = multiprocessing.Event()
 
         # Clipboard for cut copy paste
         self.clipboard = Gtk.Clipboard.get(Gdk.SELECTION_CLIPBOARD)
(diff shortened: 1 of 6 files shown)


kbengs:
I noticed one issue:
If printing or clicking on the preview button while iconview still render will result in a segmentation fault. It will also happen if you scroll while you print.

Ctrl+P shortcut could be added imo

kbengs:
Two observations:
It is possible to close the application while printing. It will cause exceptions.

The word "saving" in statusbar is not right. Maybe "printing"? But then again when waiting for the print preview to appear the word "printing" would not be completely true...

jeromerobert:
Ok, I'll fix both. What do you think about disabling the close button (i.e. cheating `on_quit`) also during saving (i.e. for both printing and saving) ?

kbengs:
Yes I think that is ok. Close while saving or printing should not be necessary.

pdfarranger/pdfarranger.py
```diff
@@ -951,6 +954,8 @@ def clear_data(self):
         malloc_trim()
 
     def on_quit(self, _action, _param=None, _unknown=None):
+        if self.disable_quit:
+            return True
         if self.export_process and self.export_process.is_alive():
             msg = _('Abort saving and quit?')
```
kbengs, line 960: I guess this option could never happen, or am i wrong? Other than this I see no issues.

PR #673, opened 2022-05-18
Updated greek translation

manussos:
Following the instructions for translators, as given in the `readme.md` file, a current Greek translation is submited.

### DIFF
--- a/data/com.github.jeromerobert.pdfarranger.metainfo.xml
+++ b/data/com.github.jeromerobert.pdfarranger.metainfo.xml
@@ -38,12 +38,12 @@ See file COPYING or go to <http://www.gnu.org/licenses/> for full license detail
       It is a frontend for pikepdf.
     </p>
 
-    <p xml:lang="es">
+    <p xml:lang="el">
       Το PDF Arranger είναι μια μικρή εφαρμογή, που βοηθά τον χρήστη να συγχωνεύσει η
       να διαχωρίσει αρχεία τύπου PDF και να περιστρέψει, περικόψει η αναδιατάξει τις
       σελίδες τους χρησιμοποιώντας ένα διαδραστικό και διαισθητικό γραφικό περιβάλλον.
     </p>
-    <p xml:lang="es">
+    <p xml:lang="el">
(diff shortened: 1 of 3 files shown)


data/com.github.jeromerobert.pdfarranger.metainfo.xml
```diff
@@ -37,6 +38,15 @@ See file COPYING or go to <http://www.gnu.org/licenses/> for full license detail
       It is a frontend for pikepdf.
     </p>
 
+    <p xml:lang="es">
```
angsch, line 41: "es" -> "el" (same for line 46)

angsch:
Thank you very much for your PR. Could you change the language code for metainfo.xml?

manussos:
Whoops! I've just changed it.
Thanks for the catch!

po/el.po
```diff
@@ -1,506 +1,1370 @@
...
-msgstr ""
+msgstr "Απαιτείται κωδικός"
 
-#: pdfarranger/core.py:186
+#: pdfarranger/core.py:202
 msgid "The password will be remembered until you close PDF Arranger."
-msgstr ""
-
-#: pdfarranger/core.py:187
+<<<<<<< Updated upstream
```
dreua, line 62: These are leftovers from a git merge/rebase conflict. Those need to be cleaned up before we can move on with this.

dreua (changes requested):

po/el.po
dreua: This should also fix CI. I'm not sure how you ended up with this situation, tbh, but I know Git can be confusing at the beginning. Let me know if you need any assistance :wink:

manussos: Yeah, working with git is a joyful experience 😵‍💫
No idea how I got it messed up like that!

manussos: Should be okay now, hopefully 🤞...

dreua: It becomes joyful (or at least useful) when you get used to Git :wink:

dreua (approved):
Looks good to me now, thank you for your contribution!

manussos:
You are welcome @dreua. Thank you for the merge and for taking the time to review!

PR #683, opened 2022-06-17
update ru.po

Ser82-png:


angsch:
@Ser82-png  Thank you for your PR. Could you please be so kind and add translations to [data/com.github.jeromerobert.pdfarranger.metainfo.xml](https://github.com/pdfarranger/pdfarranger/blob/main/data/com.github.jeromerobert.pdfarranger.metainfo.xml) and [data/com.github.jeromerobert.pdfarranger.desktop](https://github.com/pdfarranger/pdfarranger/blob/main/data/com.github.jeromerobert.pdfarranger.desktop), too? See https://github.com/pdfarranger/pdfarranger#for-translators.

PR #685, opened 2022-06-27
Fix duplicated pages management in booklets

### LINKED ISSUE #681: Booklet generation incorrect on specific PDF
Describe the bug
----------------

The position of individual pages in the Generate Booklet feature is off for certain PDFs.

To Reproduce
------------

1.  Import tests/test.pdf
2.  Split the first page.
3.  Generate booklet from the first two pages

Screenshots
-----------

image

Interestingly, if both pages of test.pdf are split and the middle pages are removed, the result is as expected.

image

jeromerobert:
* PikePDF require that all pages have been copied to the temporary PDF
document BEFORE doing any change to the media box
* Fix #681

### DIFF
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -203,10 +203,9 @@ def export_process(*args, **kwargs):
     warn_dialog(export)(*args, **kwargs)
 
 
-def export_doc(pdf_input, pages, mdata, exportmode, file_out, quit_flag):
-    """Same as export() but with pikepdf.PDF objects instead of files"""
-    global _report_pikepdf_err
-    pdf_output = pikepdf.Pdf.new()
+def _copy_n_transform(pdf_input, pdf_output, pages, quit_flag=None):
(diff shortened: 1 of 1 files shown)


angsch:
@jeromerobert Working marvelously, thank you. I would be happy to merge this in.

While reading the code, there are a few occurrences of
```
if quit_flag is not None and quit_flag.is_set():
    return
```
from earlier commits that appear redundant or could be moved to the beginning of the routine. Probably better done in another commit.

jeromerobert:
> I would be happy to merge this in.

So please yourself.

> from earlier commits that appear redundant or could be moved to the beginning of the routine.

I don't think so. `quit_flag` can change at any time during the export because export is run in an separate thread. This was implemented by *kbengs* to allow the main window to be rendered during the export and avoid to have it all grey. If the main thread quit and the export thread keep running many exceptions are raised.

angsch:
>  If the main thread quit and the export thread keep running many exceptions are raised.

Allright, thanks for the explanation.

PR #700, opened 2022-08-02
Update is.po

sveinki:
Updated Icelandic translation, PO merged with today's POT


sveinki:
see also #701 and #702

PR #703, opened 2022-08-03
Update Catalan translation

juxuanu:
For new release, following https://github.com/pdfarranger/pdfarranger/discussions/693


jeromerobert:
Rebased and merged as 284ca31

You did not updated the `.po` file. I guess that's because you can't run the `data/updatepo.sh` script to update from the `.pot` file. Do you want me to do it so you just have to edit the `.po` file ?

juxuanu:
@jeromerobert I ran the script, I don't know what went wrong. I checked the generated files's timestamps and they were from the moment the script was run...

Anyway, I'd appreciate the updated file to complete the translation.

Also, could it be possible to update them in the CI, maybe? It might be handier for contributors... Or use some external website like Weblate or Crowdin.

jeromerobert:
> Anyway, I'd appreciate the updated file to complete the translation.

Done

> Also, could it be possible to update them in the CI, maybe? It might be handier for contributors...

This is a question I have asked myself several times. I just wonder if this is what all translators prefer or if some want to have complete control over the update. In particular should I run `msgmerge -U` and `msgattrib --no-obsolete` and `msgattrib --no-fuzzy`. Let's move this discussion to the release thread where most translators are already notified.

> Or use some external website like Weblate or Crowdin.

I found this difficult to manage on an other project so I would prefer to avoid.

dreua:
All the relevant changes from this PR are already merged, therefore I'm closing it.
Jerome already updated the po-file in a6ee0c0fa773e74bcfa72d34330c2d4e2a7826c9 could you take another look at that? It would be nice to have at least the "fuzzy" translations reviewed (and the "fuzzy" tags removed, otherwise they won't show up).

PR #704, opened 2022-08-04
Use `FileChooserNative` insted of `FileChooserDialog`

oscfdezdz:
Enables native platform file chooser when available.

See https://docs.gtk.org/gtk3/class.FileChooserNative.html

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -94,13 +94,13 @@ try:
 except ValueError:
     Handy = None
 
-if Gtk.check_version(3, 12, 0):
+if Gtk.check_version(3, 20, 0):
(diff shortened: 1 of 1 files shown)


jeromerobert:
See https://github.com/pdfarranger/pdfarranger/commit/46d11ce0302ab81306dafaccc7ccb18783b7c96e

You need to bump GTK version in `Gtk.check_version`.

oscfdezdz:
> See [46d11ce](https://github.com/pdfarranger/pdfarranger/commit/46d11ce0302ab81306dafaccc7ccb18783b7c96e)
> 
> You need to bump GTK version in `Gtk.check_version`.

I just tested it on Windows and got the same result. On Linux, flatpak build works fine.

PR #705, opened 2022-08-06
Update hu.po

ovari:
Fixes Hungarian (magyar) https://github.com/pdfarranger/pdfarranger/discussions/693#discussioncomment-3309466

1. Does `msgid "Zoom _Fit"` mean `Zoom to _fit` (`_Mérethez igazítás`) or `Size to _Fit` (`Ill_eszkedő méret`)?
https://www.microsoft.com/en-us/language/Search?&searchTerm=Fit&langID=382&Source=true&productid=undefined

2. Do `msgid "% of width"` and `msgid "% of height"` have values before or after the string?
`% of &width:` https://www.microsoft.com/en-us/language/Search?&searchTerm=of%20width&langID=382&Source=true&productid=undefined

Thank you

### DIFF
--- a/po/hu.po
+++ b/po/hu.po
@@ -69,19 +69,19 @@ msgstr "mm"
 
 #: pdfarranger/croputils.py:147
 msgid "Left"
-msgstr "Balra"
+msgstr "Balról"
 
 #: pdfarranger/croputils.py:147
 msgid "Right"
-msgstr "Jobbra"
+msgstr "Jobbról"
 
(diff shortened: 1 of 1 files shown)


kbengs:
1. `Zoom fit` will zoom in on a page as much as possible so that it will fit into the application window
2. Before
image

ovari:
Thank you @kbengs for your comment with image.

Have updated the hu.po file to enable it to be merged.

Thank you once again.

PR #713, opened 2022-09-02
Avoid adding mimetypes to GtkFileFilter on Windows

oscfdezdz:
IFileDialog implementation does not handle them, that is why it falls back to GtkFileChooserDialog.

See https://docs.gtk.org/gtk4/class.FileChooserNative.html#win32-details

I am not sure if this was specified in the documentation when I opened #704.


kbengs:
@oscfdezdz Thank you.

PR #718, opened 2022-09-17
pikepdf 6 support

jeromerobert:


dreua:
Furst of all thank you for taking care of this issue!

Die you See the other occurrence for Annotations? Do we have a teatcase for that?

https://github.com/pdfarranger/pdfarranger/blob/58d3646429712b4bb610beac9d5ff627dd795f71/pdfarranger/exporter.py#L234

jeromerobert:
> Die you See the other occurrence for Annotations?

Yes and I checked that it was not affected using the PDF included in the annotation feature request.

> Do we have a teatcase for that?

No, not so far.

PR #721, opened 2022-09-18
Fix booklet creation

### LINKED ISSUE #720: Split Pages and Create Booklet Fails
Describe the bug
----------------

A [document](https://github.com/pdfarranger/pdfarranger/files/9592821/Rain-Harvesting-Handbook-AU-Issue-2-2021-web.pdf) of A4 cover and back sheet with A3 pages appears to split into A4 sheets successfully but fails to successfully reform into a booklet.

The second and third last page are alternately reproduced on the left side.

To Reproduce
------------

1.  Open the linked document
2.  Select page 2 to second last page (page 61)
3.  Split Pages 2 columns
4.  Select page 2 to second last page, Generate Booklet

Expected behavior
-----------------

Document to be reconstructed as original.

Input files
-----------

Document is linked in description.

Screenshots
-----------

System and Versions
-------------------

PDF Arranger V1.9.0, Win10 from GitHub release.
-   It uses libqpdf 10.4.0, pikepdf 4.3.1, GTK 3.24.31 and Python 3.9.7.

jeromerobert:
In python lists a negative index i means (i+len(list)) but in pikepdf it means "the very last element of the array", whatever the value of i.

Fix #720

### DIFF
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -295,10 +295,15 @@ def generate_booklet(pdfqueue, tmp_dir, pages):
     source_files = {n-1: pikepdf.open(pdfqueue[n - 1].copyname) for n in file_indexes}
     _copy_n_transform(source_files, file, pages)
     to_remove = len(file.pages)
-    for i in range(len(pages)//2):
+    npages = len(pages)
+    for i in range(npages//2):
         even = i % 2 == 0
         first_id = -i - 1 if even else i
         second_id = i if even else -i - 1
+        if first_id < 0:
+            first_id += npages
+        if second_id < 0:
+            second_id += npages
         first = pages[first_id]
         second = pages[second_id]
         first_foreign = file.pages[first_id]


dreua:
Is this documented behavior of pikepdf? Otherwise that's either a bug to be fixed or documentation to be improved.

kbengs:
The order of pages seems to be ok but I see strange results on the document linked in 720. The text is overlapped with text from another page:
A

jeromerobert:
@kbengs I do not reproduce that. Is it after export or in the Poppler preview ? With which version of PikePDF & qpdf ?

jeromerobert:
@dreua I think it's a bug in PikePDF.

kbengs:
Poppler preview and exported file looks the same. 
libqpdf 10.6.3, pikepdf 5.4.2, GTK 3.24.33 and Python 3.10.4.

jeromerobert:
I just tried with `libqpdf 10.6.3, pikepdf 5.4.2, GTK 3.24.34 and Python 3.10.6` and still cannot reproduce the issue.

kbengs:
Okay. Anyway it's not related to this pr. I see the issue in pdfarranger 1.8.0 also. Also tested a few versions of pikepdf and I see the issue in all versions I tested. I will investigate more later.

kbengs:
Here is a pdf with 4 pages.
* Open [minimum.pdf](https://github.com/pdfarranger/pdfarranger/files/9601184/minimum.pdf)
* Split page 2, 3 and 4
* Select all and generate booklet
* Export, result: [out2.pdf](https://github.com/pdfarranger/pdfarranger/files/9601200/out2.pdf)

dreua:
> @dreua I think it's a bug in pikepdf.

I think the issue here is that the there are pages appended the loop, i.e. the length of `file.pages` increases with each iteration. I'd say this has been a bug on our side and pikepdf is bug-free concerning negative integers until proven otherwise. (I was about to create an issue with pikepdf about this but couldn't reproduce the issue.)

PR #725, opened 2022-09-21
Add Italian translation

mapi68:
Added Italian translation, fix language order and xml formatting.

### DIFF
--- a/data/com.github.jeromerobert.pdfarranger.metainfo.xml
+++ b/data/com.github.jeromerobert.pdfarranger.metainfo.xml
@@ -6,165 +6,165 @@ License version 3 or later.
 See file COPYING or go to <http://www.gnu.org/licenses/> for full license details.
 -->
 <component type="desktop-application">
-  <id>com.github.jeromerobert.pdfarranger</id>
-  <name>PDF Arranger</name>
-  <name xml:lang="ko">PDF 편집기</name>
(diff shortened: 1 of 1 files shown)


dreua:
For proper review those need to be split up in 3 commits. Furthermore I'm not sure we want to change the formatting, are there good reasons to do that?

dreua:
Nitpick: Commit comments should be in present tense, I.e. Add instead of Added.

mapi68:
> For proper review those need to be split up in 3 commits. Furthermore I'm not sure we want to change the formatting, are there good reasons to do that?

Only cosmetic changes......

mapi68:
> Nitpick: Commit comments should be in present tense, I.e. Add instead of Added.

You're right.

dreua:
The [Freedesktop example](https://freedesktop.org/software/appstream/docs/chap-Quickstart.html) as well as [Nautilus](https://github.com/GNOME/nautilus/blob/master/data/org.gnome.Nautilus.appdata.xml.in.in) use two spaces indentation, therefore I won't merge a commit changing that to tabs without giving a very good reason.

I'd still like to add your translations, would you submit them without any other changes? (If you want to reorder, do that in a separate commit, please.) You may use `--force-with-lease`  to update this branch / PR.

jeromerobert:
Superseded by #734

PR #727, opened 2022-09-21
Update it.po

mapi68:


po/it.po
```diff
@@ -266,23 +271,23 @@ msgstr "_Salva"
...
 
 #: pdfarranger/pdfarranger.py:940
 msgid "Save changes to “{}” before closing?"
-msgstr "Salvare le modifiche apportate a “{}”prima di chiudere?"
+msgstr "Salvo le modifiche apportate a “{}”prima di chiudere?"
 
 #: pdfarranger/pdfarranger.py:943
 msgid "Save changes before closing?"
-msgstr "Salvare le modifiche prima di chiudere?"
+msgstr "Salvo le modifiche prima di chiudere?"
```
dreua, line 286: It looks like you are changing the grammar here. Is this consistent with how other software, ideally Gnome software, is using it?

po/it.po
```diff
@@ -484,15 +491,15 @@ msgstr "_Elimina"
 
 #: data/menu.ui.h:19
 msgid "Dupl_icate"
-msgstr "D_uplica"
+msgstr "Dupl_ica"
 
 #: data/menu.ui.h:20
 msgid "Re_verse Order"
-msgstr "_Inverti Ordine"
+msgstr "In_verti ordine"
```
dreua, line 499: Please explain this change, it's not necessarily wrong, I just want to be sure you know what you are doing here.

dreua (commented):
Thank you for the new translations and fixes! I've just two quick questions to make sure we are heading in the right direction here, I hope you don't mind.

po/it.po
mapi68: "pick a strong letter (e.g. in "Search and replace" rather pick s, r or p than a)"

po/it.po
mapi68: I did a search and saw that many translations use the infinitive form, I think gnome too. Do I have to change them all?

po/it.po
mapi68: > It looks like you are changing the grammar here...

Here must be "Invertire ordine"?

po/it.po
dreua: I don't know about Italian but most software uses the infinitive form. If the remains off my Latin are of any help, you are changing this away from the infinitive form to some other form. Why do you think this is better? I think we should go with what most software does unless there are good reasons or a (Gnome) guideline saying otherwise.

Maybe the other translators, @MichelePezza or @albanobattistella, can chime in with there opinions?

albanobattistella: HI,

"Salvare le modifiche prima di chiudere?"  thisi is the correct form.

mapi68: It's "Inverti ordine" without uppercase "O", the same as "Seleziona tutto" and not "Selezionare tutto".
I finish now to revert those changes. 
How can I do to update file without create a new pull request?

jeromerobert:
Superseded by #734

PR #740, opened 2022-10-02
manpage: fix NAME to be the actual executable name

kapouer:
lexgrog, which is used by mandb, ignores names with spaces (see lexgrog.1).


dreua:
Thank you!

Issue #729, opened 2022-09-22
Wrong text and images location when creating booklet


jeromerobert:
The order of pages seems to be ok but I see strange results on the document linked in 720. The text is overlapped with text from another page:
![A](https://user-images.githubusercontent.com/60842129/190944105-ad32c686-f1f2-4334-b2b2-4dfe1b18de85.jpg)

_Originally posted by @kbengs in https://github.com/pdfarranger/pdfarranger/issues/721#issuecomment-1250522146_

jeromerobert:
Here is a pdf with 4 pages.
* Open [minimum.pdf](https://github.com/pdfarranger/pdfarranger/files/9601184/minimum.pdf)
* Split page 2, 3 and 4
* Select all and generate booklet
* Export, result: [out2.pdf](https://github.com/pdfarranger/pdfarranger/files/9601200/out2.pdf)

_Originally posted by @kbengs in https://github.com/pdfarranger/pdfarranger/issues/721#issuecomment-1251361615_

dreua:
> Here is a pdf with 4 pages.
> 
>     * Open [minimum.pdf](https://github.com/pdfarranger/pdfarranger/files/9601184/minimum.pdf)
> 
>     * Split page 2, 3 and 4

* Duplicate the first page

>     * Select all and generate booklet
> 
>     * Export, result: [out2.pdf](https://github.com/pdfarranger/pdfarranger/files/9601200/out2.pdf)

With this step added, I too see weird overlays.

![image](https://user-images.githubusercontent.com/15114715/191740900-0ec0e96a-a406-4794-918c-c29aa1b299d9.png)

dreua:
Maybe this is due to content which has been just hidden by a Crop-/MediaBox reappearing? Afaik, split page just sets the CropBox. It may not be obvious with each PDF depending on how it was cropped and whether it has transparencies.

kbengs:
To reproduce this since f4d89708f08eef8bf9462b85c5bbbab220d81677 was merged:
1. Open [minimum.pdf](https://github.com/pdfarranger/pdfarranger/files/9601184/minimum.pdf)
2. Delete the first page
3. Select all and split
4. Duplicate the first page
5. Select all and generate booklet

-> There is text from another page on page 2 to the left and on page 3 to the right

kbengs:
This is fixed since fe0699964e8bfed2aeb3080e1b9c337f3e01eb0d when TrimBox and CropBox will be removed on export.

PR #753, opened 2022-10-30
Fix equal page size test in generate booklet

kbengs:
This seems to fix it...


kbengs:
No there is still more to fix with #729 But this is one thing.

PR #760, opened 2022-11-22
Improve multi-export workflow

kbengs:
Ask if you want to know more why the changes are done

### DIFF
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -270,7 +270,7 @@ def _copy_n_transform(pdf_input, pdf_output, pages, quit_flag=None):
         )
 
 
-def export_doc(pdf_input, pages, mdata, exportmode, file_out, quit_flag):
+def export_doc(pdf_input, pages, mdata, files_out, quit_flag):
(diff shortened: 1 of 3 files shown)


jeromerobert:
I can help to fix the tests but let's wait for 1.9.2 before merging that one.

kbengs:
This should be working. This should improve saving/exporting imo. :-)

jeromerobert:
You can cherry-pick 51a38e6 to fix the test.

jeromerobert:
The default file name become `test-0001.pdf` in the export file chooser. If I keep it as is it lead to exported file named like `test-0001-0001.pdf`. Is this expected ?

kbengs:
Yes the filename would be `test-0001-0001.pdf` if exporting to "individual files" The first `0001` say it's 1:st export, the second `0001` say it's page number 1. The good thing about this is that the names should not conflict with already existing names, even if you use the export action several times.

I had doubts about if this was good. It looks complicated and the filename in the dialog will not be the exact name of the first exported file. Should I remove the first `0001` part of `test-0001-0001.pdf` when exporting to "individual files" ? Any opinions?

PR #772, opened 2022-12-05
Update win32 build instructions

jeromerobert:
pikepdf as changed a lot since we wrote those instructions


jeromerobert:
I'll try fix https://github.com/pikepdf/pikepdf/issues/426 before merging

kbengs:
For some reason it still fails for me with  `undefined reference to vtable for QPDFNameTreeObjectHelper`

I guess installation of img2pdf should be moved to after pikepdf is installed, because it otherwise tries to install it and fails.

jeromerobert:
> For some reason it still fails for me with `undefined reference to vtable for QPDFNameTreeObjectHelper`

I replayed the whole installation process and cannot reproduce that issue. Are you sure you are using qpdf 11.2.0 (i.e. that one https://packages.msys2.org/package/mingw-w64-x86_64-qpdf?repo=mingw64) ?

> I guess installation of img2pdf should be moved to after pikepdf is installed, because it otherwise tries to install it and fails.

Indeed, I'll fix the PR.

kbengs:
> Are you sure you are using qpdf 11.2.0

No I forgot the `pacman -Syu` step.. sorry.

Also `cd ~` before `git clone https://github.com/pdfarranger/pdfarranger.git`

PR #776, opened 2022-12-10
GHA: bump windows docker image

### LINKED ISSUE #719: Not remembering window position
Hi, thanks for great app
i'm using app on windows with multi monitor setup
my secondary monitor on left side and app always starts on my secondary monitor even if i close window on primary monitor
secondary monitor is 1080p, primary monitor is 1440p

jeromerobert:
Upgrade Windows docker image to:
* pikepdf 6.2.5
* qpdf 11.2.0
* gtk 3.24.35

I did not tested the generated zip/msi that much and this needs to be done  before merging.

This should fix #719

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -47,7 +47,7 @@ jobs:
   build-win32:
     runs-on: ubuntu-latest
     container:
-      image: jeromerobert/wine-mingw64:1.7.1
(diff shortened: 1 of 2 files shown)


kbengs:
I can't see any new issues caused by the update. I tested both zip and msi. And application window does remember the right screen. Confirmed by me and yw4z. I think we could merge this.

PR #777, opened 2022-12-10
Bump GH actions

dreua:
Fix warnings about Node.js 12 actions being deprecated

Let's run the CI and see how it works...


dreua:
CI runs fine and we are down to one warning: "ubuntu-latest workflows will use ubuntu-22.04 soon"

That is what I expected, can be merged from my side. (I think we'll just stick with this occasional warning whenever they update the runner images. It will vanish automatically "soon" ;)

PR #791, opened 2022-12-28
Minor code quality fixes

jeromerobert:


pdfarranger/pdfarranger.py
```diff
@@ -71,6 +71,7 @@
     try:
         locale.bind_textdomain_codeset(DOMAIN, 'UTF-8')
     except AttributeError:
+        # Was tested just above so cannot happen
```
dreua, line 74: If it can not happen, then why do we want a try/except here at all? Still approving because this is not mission cricital in any way, just me wondering whether this is the 100% perfect solution :wink:

dreua (approved):
Thank you, looks like a good christmas-cleanup to me!

pdfarranger/pdfarranger.py
jeromerobert: Oops.  This comment is wrong.

PR #794, opened 2023-01-01
Allow user to modify export counter value

kbengs:
self.save_file is used for "save all to single file"
self.export_file is used for all other "exportmodes"

@petaflot would you like to test this before merge?


petaflot:
probably didn't test it the way I was supposed to (patched manually), but from what I've seen it works fine.

suggestion : IMHO it would be nice that "ctrl+shift+e" should export+save+delete selection. any thoughts?

kbengs:
Thanks for testing. Not sure about "export+save+delete" But if it's done well and documented maybe it could be merged.  "ctrl+shift+e" is already used though.

PR #807, opened 2023-01-13
Update German translation

davidkyburz:


po/de.po
```diff
@@ -6,21 +6,55 @@
...
+msgstr "Sprache"
+
+#: pdfarranger/config.py:221
+msgid "(Requires restart)"
+msgstr "(Erfordert Neustart)"
+
+#: pdfarranger/config.py:227
+msgid "Theme"
+msgstr "Thema"
+
```
dreua, line 50: I think this is about the color scheme, maybe we can have a look how this is translated and named in other applications.

po/de.po
```diff
@@ -219,161 +253,166 @@ msgstr ""
...
+#: pdfarranger/pdfarranger.py:1055
 msgid "Export…"
 msgstr "Exportieren …"
 
-#: pdfarranger/pdfarranger.py:1051 data/menu.ui.h:1
+#: pdfarranger/pdfarranger.py:1098
+#, python-format
+msgid "A file named \"%s\" already exists. Do you want to replace it?"
+msgstr ""
+"Eine Datei mit dem Namen \"%s\" existiert bereits. Möchten Sie sie ersetzen?"
```
dreua, line 349: I think "Sie diese ersetzen" is more common and smoother.

dreua (approved):
Thanks for the good translations, I've commented on two strings which are okay but not perfect IMO. Would you like to improve those? I'm fine with merging it anyway.

Vielen Dank!

po/de.po
davidkyburz: Yes, it's about the color scheme. I will change it to "Erscheinungsbild". This term is used in Firefox. Other ideas are "Stil" (Gnome) or "Modus" (Windows).

po/de.po
davidkyburz: Thanks for pointing this out.

dreua:
Perfekt, danke!

PR #818, opened 2023-01-20
Fix missing returns

dreua:


dreua:
I just put both commits in this PR; not married to the second let me know if you don't like it ;)

pdfarranger/pdfarranger.py
```diff
@@ -727,12 +727,13 @@ def render(self):
 
     def quit_rendering(self):
         """Quit rendering."""
-        if self.rendering_thread:
-            self.rendering_thread.quit = True
-            # If thread is busy with page.render(cr) it might take some time for thread to quit.
-            # Therefore set a timeout here so app continues to stay responsive.
-            self.rendering_thread.join(timeout=0.15)
-            return self.rendering_thread.is_alive()
+        if not self.rendering_thread:
```
kbengs, line 730: Should this be `if self.rendering_thread is None:`

pdfarranger/pdfarranger.py
```diff
@@ -1252,7 +1254,7 @@ def save_warning_dialog(self, msg):
     def export_finished(self, exportmode, export_msg):
         """Check if export finished. Show any messages. Run any post action."""
         if self.export_process.is_alive():
-            return True
+            return True # continue polling
```
kbengs, line 1257: "Inline comments should be separated by at least two spaces from the statement." :point_up:

pdfarranger/pdfarranger.py
dreua: My version was logically equal to the previous version but yours is even better, I agree.

pdfarranger/pdfarranger.py
dreua: Good catch, thanks!

dreua:
@kbengs I added exactly those two changes you requested to the first commit, thanks for the input!

kbengs:
This looks good to me.

dreua:
https://github.com/pdfarranger/pdfarranger/security/code-scanning

Back to 0 alerts 🎉

PR #819, opened 2023-01-22
Fix error on old libhandy

kbengs:
Issue was introduced in 0c19bec92566dd59919e12189f13d09af3dfac10


dreua (approved):
I'm not familiar with libhandy but this change looks reasonable to me.

Issue #786, opened 2022-12-19
Don't prompt to save when no changes have been made


K4LCIFER:
Describe the bug
----------------

With a PDF open in PDF Arranger, even if you don't make any changes, you are prompted to save on close

To Reproduce
------------

1.  Open a pdf
2.  click exit
3.  observe that a save prompt appears

Expected behavior
-----------------

I expect that if I open a PDF document, and I dont make any changes, I should be able to immediately close it without having to save anything. Sometimes I just want to open PDFs to make a export of a selection, and not create an  overall project with the PDF..

System and Versions
-------------------

-   PDF Arranger version and where you get it (distro package, flatpak, pip, git, download, ...)  1.9.2 (flatpak)
-   pikepdf version and where you get it (distro package, pip, git, ...) 6.2.6 (came with the program)
-   OS name and version (e.g.: Windows 10, Fedora 33, ...) Arch Linux (Gnome 43.2 wayland)

kbengs:
As I see it the idea has been that you start with an empty "workspace" and then import pages to it. So as soon as you add a document the document/workspace is "unsaved", and then it will ask if you want to save changes when closing.
Now that we have a "Open" button this is more confusing..

Here are som ideas what we could do.

Idea 1: (make "Open" behave more like it does in texteditors)
* Allow to open only one pdf with the "Open" action
* "Open" action set self.export_file and set the filename in titlebar
* "Open" action do not set the "unsaved" flag
* Now if user press "Save" there would be a dialog warning about  "forms and outlines are lost on saving" and maybe  something more too. This would have a "Don't show this warning again" checkbutton. And it would have buttons for "Overwrite" and "Save as"
* All other methods of adding documents would be treated as imports and would set document "unsaved".

So in a new pdfarranger instance if you drag in a document and do no changes you would still be prompted to save on close. 
If you open the document with "Open" and do no changes you would not be prompted to save on cĺose.

Idea 2:
Another idea would be to not set it unsaved when the instance is "clean" and exactly one full pdf is added, regardless of how the pdf is added. Maybe rename "Open" to "Import in new window" or something.

Idea 3:
A third idea would be to continue as before. Maybe rename "Open" to "Import in new window".

What do you think?

dreua:
Could we use/modify the undo manager to check whether there is exactly one import in the history? Don't we already have a star or something indicating whether a file had been changed? (I really should test / look at the code before commenting on my phone, sorry.)

[...]

Issue #65, opened 2019-04-28
default save and replace original file


giannino:
I wish it could behave like all file modifiers: upon closing should warns about replacing the original file and save the modified file.
using ubuntu 19.04, could be realted to file attributes?

jeromerobert:
It does not because pdfarranger is not rigorously a file modifier. The PyPDF2 import/export is not a transparent operation. If you add a PDF file to pdfarranger then save, PyPDF2 rewrite the whole file and possibly change it's structure. Most of the time we don't want the original file to be replaced because there is a high risk of data loss.

This is why the *save* action redirect to *save as* until the user has specified a file name to save to.

Here is what I suggest:

* if a single file (or no file) was imported and not modified
     * Quietly close
* else ask about save confirmation, and if yes
     * if the user as already manually run the *save* or *save as* action to select a different file name
          * Quietly save and close
     * else
          * Open a *save as* dialog

The implementation is linked to #59 because this is where I can get the modified status of the file.

giannino:
What you show it's the implementation already in place, so why not add an
other option _with warning of data loss_ or , even better, creating a
backup copy of the original file?
I guess that 99& of the times the common user discard the original file, I
use mostly for deleting pages.

Thanks for your help

dreua:
How is the risk of data loss addressed with this solution? I think it's rather serious since the exported file can be corrupted while the preview in pfarranger looks good. The user has no possibility to check the results before it's too late

jeromerobert:
@dreua I did not implemented what OP asked for. A file chooser still popup at the first save, but now it default to the first imported PDF files. So before overwriting a file the user still have to confirm two times:
* Click save in the file chooser without changing the file name
* Confirm that he wants to overwrite

He can also blindly press `ctrl+s`, `enter`, `enter` which allow to get almost what OP wants.

dreua:
That sounds okay to me, thank you for the clarification!

madduck:
For me, the "save" dialog does not default to a filename, unfortunately. If it did that, I guess I could live with it (and #100 could be ignored).

Also, how do you get `ctrl+s` to work?

dreua:
@madduck For me the "save" dialog always defaults to a filename (pdfarranger 1.3.0) and `ctrl+s` just works. Unfortunately I can only look into your problem if there is a way for me to reproduce it.

petaflot:
[...]

PR #825, opened 2023-01-31
Fixes save button behavior when editing single PDF file.

lavafroth:
When the user has opened and is editing a single PDF file, the save button does not prompt for assigning a new filename. It instead saves the document to its original path.

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -255,6 +255,7 @@ class PdfArranger(Gtk.Application):
         self.rendering_thread = None
         self.export_process = None
         self.post_action = None
+        self.original_path = None
         self.save_file = None
         self.export_file = None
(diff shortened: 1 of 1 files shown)


dreua:
Afaik this is intended behavior and has been discussed before. The gist: We're not editing but rewriting the whole file with potential for data loss, that is why we want explicit consent from the used to overwrite their existing file.

Thank you for your well intended PR and sorry if we are not going to merge it. In general I recommend to discuss changes like this before investing the time to code and create a PR.

@jeromerobert @kbengs @angsch shall we rediscuss this matter or close this PR?

jeromerobert:
See #65 and #786 (and many other). Even if it is well done I'm afraid we have to close that PR.

lavafroth:
Alright, I will close this PR now. Thanks a lot for looking into it!

PR #832, opened 2023-02-25
Add support for pasting a page as overlay & underlay

kbengs:


kbengs:
* Overlays & underlays are rendered with poppler (in memory).
* A page can have multiple overlays & underlays.
* A overlay/underlay page has a offset array (left, right, top, bottom) to keep track of where the overlay/underlay is added onto the destination page. The offset is relative to the destination page. The overlay/underlay is added into this offset box.
* It should not be too hard to implement N-up with this technique. (create a empty page large enough to hold all pages, add the pages on to the empty page)

Sorry for a lot of changes. I will try to fix the codacy issues. Any comments on this?

setup.py
```diff
@@ -51,7 +51,7 @@
     entry_points={
         'console_scripts': ['pdfarranger=pdfarranger.pdfarranger:main']
     },
-    install_requires=['pikepdf>=1.17.0','python-dateutil>=2.4.0'],
+    install_requires=['pikepdf>=2.14.0','python-dateutil>=2.4.0'],
```
dreua, line 54: Are we ready to cut support for pikepdf < 2.14? According to repology that would include Debian 11. (It can probably be installed from pypi.)

pdfarranger/exporter.py
```diff
@@ -75,7 +75,7 @@ def _intersect_rectangle(rect1, rect2):
     ]
 
 
-def _mediabox(page, crop):
+def _mediabox(page, crop=[0., 0., 0., 0.]):
```
dreua, line 78: Good catch by codacy: This is dangerous, and one of the worst pitfalls in Python imo. While it may work with the current implementation it may break horribly when the code is changed to modify this reference to the "crop" singleton, it can lead to bugs which are horribly hard to debug.

Just make the default value None or maybe 0 in this case and add an if to replace it with the list.

dreua (changes requested):
I agree with codacy, this default initializer is dangerous. Haven't looked deeper into the rest.

dreua:
Are those commits strictly necessary for this PR? TBH, I don't understand the necessity, what are the actual issues they solve?

- [Use hash to check if pasted data is valid](https://github.com/pdfarranger/pdfarranger/pull/832/commits/9c39c76a4f863b65fbbfd637e48b093447cc1b98)
[9c39c76](https://github.com/pdfarranger/pdfarranger/pull/832/commits/9c39c76a4f863b65fbbfd637e48b093447cc1b98)
Purpose is to avoid unexpected behavior if paste data is corrupt


- [Remove some unnecessary steps when printing](https://github.com/pdfarranger/pdfarranger/pull/832/commits/99a19caec97593ccd4ac25572192f0fcba2b8d5e)
[99a19ca](https://github.com/pdfarranger/pdfarranger/pull/832/commits/99a19caec97593ccd4ac25572192f0fcba2b8d5e)
Do not merge metadata & remove unreferenced resources


Otherwise this sounds like a big step towards implementing/solving some of the requested features. Not something I personally had missed but a good fit for PDF Arranger anyways.

kbengs:
Thanks for the comments.

> Are we ready to cut support for pikepdf < 2.14? According to repology that would include (or, rather, exclude) Debian 11. (It can probably be installed from pypi.)

No that does not sound good, I hadn't checked that. But we could instead add the version check in `is_paste_layer_available()`, and then say in a dialog that "pikepdf  >= 2.14 is needed for overlay/underlay support" when trying to paste.

> Use hash to check if pasted data is valid

As we use the normal clipboard we could have any data in clipboard when pasting. So some kind of check is needed imo to avoid unexpected behavior/exceptions. The current implementation does this in `is_data_valid()`. Now that pages can have layers there is more data copied to clipboard and more data to check. So doing it the old way would be troublesome. The current implementation also has it limitations (it can't find all types of "data corruption")

I chose `sha1` here because I expected it to be fast and good enough for this purpose.

> Remove some unnecessary steps when printing

When printing calls `export_doc(pdf_input, pages, mdata, files_out, quit_flag)` ,the argument `pdf_input` is a list where some on the items can be `None`. (a list is needed because if a page has overlays/underlays it will need access to all files related to the page) This became a problem in `merge_doc()` Not doing any metadata merging when printing felt like a good way to fix it. 

> Otherwise this sounds like a big step towards implementing/solving some of the requested features. Not something I personally had missed but a good fit for PDF Arranger anyways.

What I would use this feature for most is to add two or several pages to one page.

jeromerobert:
It works like a charm. I though you would implement it as "export page with overlay" + "import page" + "delete old page", and without modifying the `Page` class. But that's nice like that. Thank you !

kbengs:
> "export page with overlay" + "import page" + "delete old page"

This was the idea but didn't like that it had to write temporary files to disk. (maybe I am too concerned about disk wear)
I also think the exported file size would grow larger that way as it does not have a chance to create referenced pages.

The changes:
* Layer support is pikepdf >= 3 now.
* Found a issue with TrimBox and CropBox so they are now removed in _apply_geom_transform()
* Added a workaround for locale LC_NUMERIC issue in export.py
 https://github.com/pikepdf/pikepdf/issues/446
* Some code cleaning

pdfarranger/exporter.py
kbengs: This is fixed

kbengs:
Removing TrimBox and CropBox also fix  #729

kbengs:
The page is now rendered in one piece, page with overlays & underlays.

Once again I think it's ready.

PR #836, opened 2023-03-18
Make Handy a hard dependency and port preferences dialog to HdyPreferencesWindow

oscfdezdz:
Since version 1.6 is now [available](https://packages.ubuntu.com/jammy/gir1.2-handy-1) in the LTS version of Ubuntu, it is no longer necessary to add a PPA. Related #651

Preferences window widgets are now in an UI file, languages and theme preferences options are now stored in GListStore.

Window height and width are not modified since there are some visual glitches in Windows with HdyComboRow popover when window height is around 350 or lower.

imagen

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -37,7 +37,7 @@ jobs:
   build-old-pikepdf:
     runs-on: ubuntu-latest
     # PikePDF 1.19
-    container: jeromerobert/pdfarranger-docker-ci:1.3.1
+    container: jeromerobert/pdfarranger-docker-ci:1.3.2
     steps:
     - uses: actions/checkout@v3
     - name: Install
--- a/README.md
+++ b/README.md
@@ -61,14 +61,14 @@ sudo pacman -S poppler-glib python-distutils-extra python-pip \
 
 ```
 sudo dnf install poppler-glib python3-distutils-extra python3-pip python3-gobject \
-    gtk3 python3-cairo python3-wheel python3-pikepdf python3-img2pdf python3-dateutil
+    gtk3 python3-cairo python3-wheel python3-pikepdf python3-img2pdf python3-dateutil libhandy
 ```
 
 **On FreeBSD**
 
 ```
 sudo pkg install devel/gettext devel/py-gobject3 devel/py-pip devel/py-python-distutils-extra \
-    graphics/poppler-glib textproc/intltool textproc/py-pikepdf x11-toolkits/gtk30
+    graphics/poppler-glib textproc/intltool textproc/py-pikepdf x11-toolkits/gtk30 x11-toolkits/libhandy
 ```
 
 **Then**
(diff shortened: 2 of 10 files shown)


kbengs:
I just tested it and the dialog looks nice. :+1: Some comments/questions:
* At first the app does not start because config.ini has strings for language and theme. After modifying config.ini the app starts.
* What Window do you mean here "Window height and width are not modified.." ?
* The container version bump, what has changed there?
* Debian 11 does not have libhandy 1.6 according to https://repology.org/project/libhandy/versions Are we happy with that?
* Documentation for getdefaultlocale() say "Deprecated since version 3.11, will be removed in version 3.13"
Do we need to set `os.environ['LANGUAGE']`  at all when we want to activate "System setting" ?

```

oscfdezdz:
> At first the app does not start because config.ini has strings for language and theme. After modifying config.ini the app starts.

Checking the latest stable version, it should not happen with the theme preference if these changes make it to a stable version, since it did not exist. For the language, should I change the variable name, or would initializing it earlier prevent it?

> What Window do you mean here "Window height and width are not modified.." ?

Preferences window, there is nothing in libhandy issue tracker related to this problem. Language popover is visually cut off when its position is left or right.

imagen

> The container version bump, what has changed there?

It includes libhandy https://github.com/jeromerobert/pdfarranger-docker-ci/releases/tag/ci-1.3.2

> Documentation for getdefaultlocale() say "Deprecated since version 3.11, will be removed in version 3.13"

I will change it to getlocale() then.

> Do we need to set os.environ['LANGUAGE'] at all when we want to activate "System setting" ?

I am not sure, I ran into several errors when making the changes for the language preference, I will check it out.

kbengs:
Is it cumbersome to continue to use strings in config.ini ? Now that you use numbers, what will happen if we add a new language translation, is it possible that you will get wrong translation?  (until you open the preferences dialog and set the language you want to have again)

oscfdezdz:
> Is it cumbersome to continue to use strings in config.ini ? Now that you use numbers, what will happen if we add a new language translation, is it possible that you will get wrong translation? (until you open the preferences dialog and set the language you want to have again)

Indeed, it was an issue, language is using strings again.

>  Do we need to set os.environ['LANGUAGE'] at all when we want to activate "System setting" ?

This has also been reverted.

kbengs:
On Debian 11 it does not look right:
debian11


Can anything be done to that?

Also it is not possible to change theme. (but that is not a result of this change)

Still one note, for developers it's a bit of an issue that the theme value is changed in config.ini: you may have to modify the config.ini manually if you switch back and forth in git history, and run the app in between.

What do other pdfarranger team members think about this PR, do you support this?

jeromerobert:
> Since version 1.6 is now [available](https://packages.ubuntu.com/jammy/gir1.2-handy-1) in the LTS version of Ubuntu, it is no longer necessary to add a PPA

I think we should not care about Ubuntu for compatibility because it's not the one with longest release cycle.

This PR breaks backward compatibility for little (no ?) improvement from functional point of view. Debian 12 is not yet release and Debian 11 will remains maintained for while. RHEL7 is EOL next year so we also need a solution for Centos 7 for the time being (not sure if libhandy is in conda-forge). Flatpak is only available with root access.

So I would give a NO go to this PR.

PR #838, opened 2023-03-23
Add Handy to missing distributions in README.md

oscfdezdz:


dreua (approved):
Good catch, thank you!

PR #839, opened 2023-03-29
Paste as layer improvements

kbengs:


kbengs:
Should be okay

kbengs:
Should we warn here too that "Placing a page on top of another page does not remove any content from the PDF file, it only hides it" ?

jeromerobert:
> Should we warn here too that "Placing a page on top of another page does not remove any content from the PDF file, it only hides it" ?

I think that if "crop" was named "resize mediabox" we would not have need to warn the user. I think that here the underlay/overlay wording cleary stand we talk about layers and not changing content.

PR #841, opened 2023-04-04
Update de.po

uwe-ss:
Minor fixes in de,po

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -12,7 +12,7 @@ msgstr ""
 "Project-Id-Version: pdfarranger\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2023-01-13 00:51+0100\n"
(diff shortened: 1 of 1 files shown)


dreua (changes requested):
This is only changing typography: Can you explain why you changed this? We want to keep it as close as possible to other GNOME and Linux Software and I think this is not the style used by the majority, but I'd need to check to be sure as using English UIs despite my mother toungue.

dreua:
@jeromerobert Thanks for the reminder I saw this PR but then forgot about it.

uwe-ss:
> We want to keep it as close as possible to other GNOME … Software

Exactly that was my intend. Have a look at [https://wiki.gnome.org/de/UebersetzungsRichtlinien](url)

dreua:
I stand corrected, thank you for improving the translation and the helpful link. I guess having german as my first language doesn't make me a good translator when I never use translated software :thinking:

Issue #849, opened 2023-04-18
Can't change language when the program is set to Hebrew


yorai1212:
Just installed the software, it's in Hebrew and there's not "preferences" option anywhere. Cannot find a way to change the language to English.

dreua:
In order to help you efficiently we need some more information, the [issue template](https://github.com/pdfarranger/pdfarranger/issues/new?assignees=&labels=&template=bug_report.md&title=) is usually a good place to start.

The output of `locale` would also be helpful in your case.

It is likely that there is something wrong with your system, not PDF Arranger. Did you ever set (parts of) your system to hebrew or is this completely random?

yorai1212:
> In order to help you efficiently we need some more information, the [issue template](https://github.com/pdfarranger/pdfarranger/issues/new?assignees=&labels=&template=bug_report.md&title=) is usually a good place to start.
> 
> The output of `locale` would also be helpful in your case.
> 
> It is likely that there is something wrong with your system, not PDF Arranger. Did you ever set (parts of) your system to hebrew or is this completely random?

I have looked at the bug template, but this isn't a bug most likely, so I don't know how to use it.

The timezone on my PC is set to Israel, yes, maybe that's why it autodetcted that I wanted Hebrew as the language.

Locale, as in the folder Locale in folder "pdfarranger"? I can put it in a RAR file and share it via a site like PixelDrain.

dreua:
Forget about the locale, thats a tool in Linux, not helpful when you are on Windows.

kbengs:
1.9.2 does not have the preferences dialog. (it has been implemented later)

Here you can find instructions on how to set language in 1.9.2:
https://github.com/pdfarranger/pdfarranger/tree/1.9.2#set-custom-language-in-windows

Or if you like to install a msi built from the main branch (which has the dialog):
https://github.com/pdfarranger/pdfarranger/actions/runs/4717821797

yorai1212:
> 1.9.2 does not have the preferences dialog. (it has been implemented later)
> 
> Here you can find instructions on how to set language in 1.9.2: https://github.com/pdfarranger/pdfarranger/tree/1.9.2#set-custom-language-in-windows
> 
> Or if you like to install a msi built from the main branch (which has the dialog): https://github.com/pdfarranger/pdfarranger/actions/runs/4717821797

Oh I see, how did you get to that page though? I'm not very keen in GitHub, I just now tried going into the actions page and searching for "language", but nothing came up.

[...]

Issue #578, opened 2021-11-14
Allow user to set UI language


midaspt:
While running the Windows version of **PDFArranger**, there is no way apparent for an user to change the UI language.

While my Windows is English, my system locale isn't `en-US` because I don't live there...

But **PDFArranger** is picking up the system locale to set the program UI language -- which makes it awkwardly run in a language different from the underlying Windows OS.

An easy way to remedy this could be to make it obey a dedicated key in the "`[preferences]`" section of its '`config.ini`' file, like this:

`language=en`

I am aware of the following workaround for translators, but it doesn't seem suitable for common Windows users.

> You may test different languages with `LANG=xx_YY python3 -m pdfarranger`

dreua:
(Just for reference: I've got a [similar issue filed against Nextcloud](https://github.com/nextcloud/desktop/issues/111) since 2018 - gained a lot of up-votes and "+1"s but that's about it :laughing:)

kbengs:
If you are using windows and want pdfarranger to use english, you can remove the translation folder:
`....pdfarranger\share\locale\fi` (to remove the finnish translation)

ovari:
Can the `User Interface Language` be changed with GNU/Linux operating systems? Thank you

kbengs:
In Linux you should/might be able to start pdfarranger with `LANGUAGE=hu pdfarranger` (for me that will only translate some strings for some reason)
Setting language in config.ini will only work in Windows as it is done now. It is probably possible to make it work in Linux also, but I don't know how.

ovari:
[Xournal++](https://github.com/xournalpp/xournalpp) and [Jami](https://jami.net/) allow for changing the UI language on GNU/Linux with the program's GUI. Perhaps their code repositories can help?

Should another issue be created, like `GNU/Linux Allow user to set UI language`?

Thank you

kbengs:
Yeah, you could open a new issue for that. And add those links. :+1:

PR #852, opened 2023-04-22
Some changes related to #849

kbengs:


kbengs:
My Windows 7 install will start pdfarranger in Swedish by default although Windows is in English. This change seems to fix that. If I change Windows language (in Windows 10) it will also change in pdfarranger. This issue has been reported in #578 and #849 (I think)

PR #854, opened 2023-04-22
Replace intltool usage with xgettext

### LINKED ISSUE #853: Look into replacing python-distutils-extra
When trying to introduce automatic update checking for the Flatpak build, I've been told that [intltool was dead](https://github.com/flathub/shared-modules/pull/245#issuecomment-1518719041). Maybe it's worth looking into alternatives before it breaks beneath us, not a high priority though.

According to https://wiki.gnome.org/Projects/GnomeCommon/Migration we should be able to replace it with gettext:

> intltoolize is obsolete: gettext 0.19.7 includes all the functionality GNOME projects used from intltool, so you should port from intltool to gettext.

TingPing:
Closes #853

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ sudo dnf install poppler-glib python3-distutils-extra python3-pip python3-gobjec
 
 ```
 sudo pkg install devel/gettext devel/py-gobject3 devel/py-pip devel/py-python-distutils-extra \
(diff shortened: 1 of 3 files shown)


jeromerobert:
Not merging as is but kept for the record. See #853 for details.

jeromerobert:
genpot.sh change merged: 514edb671725a31d5459d759ef76ad94d97233c3

PR #855, opened 2023-04-26
Improve menu

kbengs:
The wiki has a section about ellipsis in "computer systems":
https://en.wikipedia.org/wiki/Ellipsis

https://stackoverflow.com/questions/278655/when-should-i-use-a-ellipsis-in-a-menu-item

### DIFF
--- a/data/menu.ui
+++ b/data/menu.ui
@@ -87,6 +87,51 @@ along with PDF Arranger.  If not, see <http://www.gnu.org/licenses/>.
       <attribute name="label" translatable="yes">_Edit</attribute>
       <section>
         <item>
+          <attribute name="label" translatable="yes">_Delete</attribute>
+          <attribute name="action">win.delete</attribute>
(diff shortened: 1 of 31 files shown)


kbengs:
I can add ellipsis to all languages before merging this

jeromerobert:
> I can add ellipsis to all languages before merging this

Thank you. I think it has to be done. I can also do it. This is as you wish.

kbengs:
I think I got all. I noticed that `./setup.py build` creates a .pot that is not sorted.

jeromerobert:
> I noticed that ./setup.py build creates a .pot that is not sorted

Yes.  Not only this make the working directory "git dirty" but it also uses `intltool` which we would like to get rid of. I wonder if we should get rid of python-distutils-extra and handle the calls to `msgfmt`, the installation of `.mo` files and the installation of icons ourselves.

PR #856, opened 2023-04-27
i18n: Adds Occitan language

Mejans:
Hello!
My small contribution.

### DIFF
--- a/po/oc.po
+++ b/po/oc.po
@@ -1,0 +1,632 @@
+# Occitan translation file
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# Quentin PAGÈS, 2023.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: \n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2022-12-08 19:32+0100\n"
+"PO-Revision-Date: 2023-04-29 18:08+0200\n"
+"Last-Translator: Quentin PAGÈS\n"
+"Language-Team: \n"
+"Language: oc\n"
+"MIME-Version: 1.0\n"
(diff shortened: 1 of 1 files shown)


po/oc.po
```diff
@@ -0,0 +1,626 @@
...
+msgid "Edit properties"
+msgstr "Modificar las proprietats"
+
+#: pdfarranger/metadata.py:212
+msgid "Property"
+msgstr "Proprietat"
+
+#: pdfarranger/metadata.py:212
+msgid "Value"
+msgstr "Valor "
```
dreua, line 201: Here is an extra space: Typo?

po/oc.po
```diff
@@ -0,0 +1,626 @@
...
+msgid "_Close"
+msgstr "_Tampar"
+
+#: pdfarranger/pdfarranger.py:943
+msgid "Save changes to “{}” before closing?"
+msgstr ""
+
+#: pdfarranger/pdfarranger.py:946
+msgid "Save changes before closing?"
+msgstr ""
```
dreua, line 279: This and the following 5 strings are missing translations which might be especially important so people can make the right decisions and not lose work.

dreua (changes requested):

dreua:
> Partial translation to start

There are some strings missing some less some more important, see my comment above.

I guess there is no harm in merging it. Its probably better than having no translation at all and falling back to English.

po/oc.po
Mejans: Yes, will make a more complete PR later then!

po/oc.po
Mejans: Oups yes typo here!

po/oc.po
dreua: Ideally just add commits to this PR i.e. to your main branch. Thanks!

jeromerobert:
Mercés

PR #859, opened 2023-05-06
No longer use python-distutils-extra

### LINKED ISSUE #853: Look into replacing python-distutils-extra

jeromerobert:
* python-distutils-extra was added by 2077a73
* it relies on intltool which is deprecated as gettext can now do the job alone
* intltool relies on perl which is a tricky dependency on Windows
* PyPA has announced disruptive changes to distutils/setuptools/pip that will propably break python-distutils-extra
* clean_i18n is removed as it was not very useful
* Close #853

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,9 +20,18 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+from setuptools import Command
 from setuptools import setup
-from DistUtilsExtra.command import (
-    build_i18n, clean_i18n, build_extra, build_icons)
+# for the time being do not use setuptools.command because:
+# * setuptools is to old in many distro (ex: Ubuntu 22.04)
+# * The Fedora package relies on wheel so it would not contains translation and
+#   icons
+from distutils.command.build import build
+
+from os.path import join
+import glob
+import os
+import subprocess
 
 data_files = [
     ('share/applications', ['data/com.github.jeromerobert.pdfarranger.desktop']),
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -20,9 +20,18 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+from setuptools import Command
 from setuptools import setup
-from DistUtilsExtra.command import (
-    build_i18n, clean_i18n, build_extra, build_icons)
+# for the time being do not use setuptools.command because:
+# * setuptools is to old in many distro (ex: Ubuntu 22.04)
+# * The Fedora package relies on wheel so it would not contains translation and
```
dreua, line 27: Do you want me to look after the Fedora stuff? Pretty sure we can get rid of the wheel. Wheel's not ideal anyway ;) (It was just the easiest back when I first packaged this.)

jeromerobert: You read my mind. Yes !

jeromerobert:
I cannot test the Windows installer of this PR for now. I'll do it as soon as I have a Windows box. Please do not merge until Windows installer has been tested.

dreua:
More discussion: https://github.com/pdfarranger/pdfarranger/discussions/860

dreua:
Test results on Windows 10, msi and zip from https://github.com/pdfarranger/pdfarranger/actions/runs/4903360382?pr=859:

1. Yeah, dark mode gets recognized and looks super cool 😎
2. Looks all good, manual language change works
4. You need to click through "Windows protected your PC" -> "More info" -> "Run anyway" (once, either during installation or first run of the portable version)

All above is true for both portable zip and installed msi.

setup.py
dreua: Just to be on the same page: The current version still works with the wheel but you want me to get rid of it in order to further improve the thing, correct?

PR #874, opened 2023-05-23
Add Ukrainian translation

Vovkiv:


Vovkiv:
Yay! Yes, i'm okay with that

PR #914, opened 2023-07-15
Add new methods Page.__repr__ and LayerPage.__repr__

m-holger:
To facilitate unit testing.

### DIFF
--- a/pdfarranger/core.py
+++ b/pdfarranger/core.py
@@ -60,7 +60,7 @@ _ = gettext.gettext
 
 
 class Page:
-    def __init__(self, nfile, npage, zoom, copyname, angle, scale, crop, size, basename, layerpages):
+    def __init__(self, nfile, npage, zoom, copyname, angle, scale, crop, size_orig, basename, layerpages):
         #: The ID (from 1 to n) of the PDF file owning the page
         self.nfile = nfile
         #: The ID (from 1 to n) of the page in its owner PDF document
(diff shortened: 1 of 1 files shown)


dreua (changes requested):
Your code looks good and reasonable to me, however, the existing class has a flaw which might become worse when adding these methods:

If angle = 90, the size tuple is reversed by the init function. I don't even know why that happens and whether the size parameter is the rotated size or the unrotated  size but once you add a `__repr__` someone might expect that its output size might be the same as the input size (or you could use it to create an identical page object) and that won't be true.

Assuming that your intention is just to write unit tests I really don't want to stand in your way, maybe we can just add a comment to warn future users about this inconsistency. (Or maybe just revert the rotation as necessary.)

Ideally however, we would clean the whole thing up by renaming the parameter unrotated/rotated_size and saving it to its own field, so that it can be easily used by the `__repr__` method. (I think I'd prefer that solution.)

Further improvement ideas:
- The whole thing would ideally become a @<no-username>dataclass which would generate the `__init__` and `__repr__` methods automatically, lowering manual maintenance, repetition and risk of error. The size calculation could happen in `__post_init__`. TBD: Would it be reasonable to require Python >= 3.7? I'd hope so since it just reached its end of life.
- Why are there two nearly identical Page classes? Pretty sure we should derive one from the other or from one common ancestor.

m-holger:
> If angle = 90, the size tuple is reversed by the init function ...

I fixed that as this would made repr behave unexpectedly. I have also renamed the init parameter so that the parameter name matches the name of the data item.

> Ideally however,  ...
> Further improvement ideas: ...

I'd rather leave that to others and focus on #462 for now.

dreua (approved):
lgtm!

PR #918, opened 2023-07-16
Harden pikepdf version checking

m-holger:
Avoid future problems with eg 10.0.0 testing as less than 8.0.0.

### DIFF
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -24,6 +24,7 @@ import tempfile
 import io
 import gi
 import locale
+import packaging.version as version
 from . import metadata
 from gi.repository import Gtk
 gi.require_version("Poppler", "0.18")
@@ -447,7 +448,7 @@ def generate_booklet(pdfqueue, tmp_dir, pages):
             )
 
         # workaround for pikepdf <= 2.6.0. See https://github.com/pikepdf/pikepdf/issues/174
-        if pikepdf.__version__ < '2.7.0':
+        if version.parse(pikepdf.__version__) < version.Version('2.7.0'):
             newpage = file.make_indirect(newpage)
         file.pages.append(newpage)
     for __ in range(to_remove):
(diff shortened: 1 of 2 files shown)


pdfarranger/exporter.py
```diff
@@ -447,7 +449,7 @@ def generate_booklet(pdfqueue, tmp_dir, pages):
             )
 
         # workaround for pikepdf <= 2.6.0. See https://github.com/pikepdf/pikepdf/issues/174
-        if pikepdf.__version__ < '2.7.0':
+        if pre_pike_2_7:
```
dreua, line 452: Good catch:
```
>>> '2.10.0' < '2.8.0'
True
```

pdfarranger/exporter.py
```diff
@@ -399,6 +400,7 @@ def num_pages(filepath):
 
 
 def generate_booklet(pdfqueue, tmp_dir, pages):
+    pre_pike_2_7 = version.parse(pikepdf.__version__) < version.Version('2.7.0')
```
dreua, line 403: Why not use parse or Version constructor on both sides? What is the difference actually?

setup.py
```diff
@@ -117,7 +117,7 @@ def run(self):
     entry_points={
         'console_scripts': ['pdfarranger=pdfarranger.pdfarranger:main']
     },
-    install_requires=['pikepdf>=1.17.0','python-dateutil>=2.4.0'],
+    install_requires=['pikepdf>=1.17.0','python-dateutil>=2.4.0', 'packaging'],
```
dreua, line 120: packaging is also required by pikepdf :heavy_check_mark:

pdfarranger/exporter.py
m-holger: I believe there is no real difference. Using both is a stylistic preference - it feels more natural to me for a literal and parse for an unknown version, but if you want me to use one or the other I don't have a problem with changing it.

PR #920, opened 2023-07-17
Convert comments in core.Page.__init__ to doc comments

m-holger:
Give IDEs a chance to display meaningful tooltips.

(Especially useful when variables don't follow Python conventions, e.g. 1-based indexing)


m-holger:
> Are you using VSCode 

I am using pyCharm thanks to a license kindly donated by JetBrains to QPDF. From a quick glance at your link, pyCharm does what is requested there - which I find very useful for C++. For Python it is not quite as useful because parameters are untyped and pyCharm often cann't deduce variable types. It would probably do better with type annotations, but I haven't tested that.

m-holger:
Obsoleted by #949

PR #922, opened 2023-07-18
Add basic unit tests for exporter.export

m-holger:
This is an illustration of the approach I am suggesting for unit testing export.

Before I continue with this PR can I have an indication whether my approach is acceptable.


jeromerobert:
It looks good. It's a good complement to dogtail testing because it's much easier to maintain. I hope it will increase our coverage rate because it's really too low at the moment.

m-holger:
@jeromerobert 

As an aside, PDFArranger set /MediaBox to letter size if no media box is defined for a page. Have you come across pages with undefined media boxes in real life PDF files?

jeromerobert:
> As an aside, PDFArranger set /MediaBox to letter size if no media box is defined for a page.

Yes, that's what the PDF specification says.

> Have you come across pages with undefined media boxes in real life PDF files?

I don't remember if I did that because it was in the specification or because I found a real PDF without media box.

m-holger:
I couldn't find that in the specification. Also, Acrobat Reader does not default to letter.

jeromerobert:
Indeed. The specification even says that MediaBox is mandatory. I really do not remember why I did that.

m-holger:
Good enough for me.

Pantivashka:
> It looks good. It's a good complement to dogtail testing because it's much easier to maintain. I hope it will increase our coverage rate because it's really too low at the moment.

Hello.
Discovered your PDF Arranger program on Ubuntu. She is beautiful.
Couldn't find another way to contact you.

Please tell me if I understand correctly that the program does not degrade the quality of images (JPEG, PNG) from which I create a PDF file?

Thank you in advance.

jeromerobert:
Trying to fix Codecov before merging...

PR #923, opened 2023-07-18
Try setuptools

dreua:
Let's see if this works :nerd_face:


dreua:
rpmbuild is green using setuptools, translations appear in the logs. Looks like the problem we had with wheels is fixed?!

PR #942, opened 2023-08-15
added the snap install in readme

soumyaDghosh:


soumyaDghosh:
Forgot to mention, the pdfarranger has 2185 active snap users. :tada:

PR #950, opened 2023-08-20
Change Page.size to a NamedTuple

m-holger:
To improve readability and reduce code duplication.

### DIFF
--- a/pdfarranger/core.py
+++ b/pdfarranger/core.py
@@ -32,7 +32,7 @@ import pathlib
 import shutil
 import tempfile
 import threading
-from typing import Optional, Tuple
+from typing import NamedTuple, Optional, Tuple, Union
 import gettext
 import gi
 from gi.repository import GObject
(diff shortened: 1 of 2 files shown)


pdfarranger/core.py
```diff
@@ -58,11 +59,34 @@
...
+        """Swap height and width"""
+        return Dims(self.height, self.width)
+
+    def scaled(self, factor: float) -> "Dims":
+        """Scale by factor"""
+        return Dims(self.width * factor, self.height * factor)
+
+    def int_scaled(self, factor: float) -> "Dims":
+        """Scale by factor and round to nearest int"""
+        return Dims(round(self.width * factor), (self.height * factor))
```
kbengs, line 79: When the thumbnail is created int(0.5 + ...) is used (line 661-662). round() will not always give exactly the same result. I think you should use int(0.5 + ..) here too.

m-holger: Changed. Also updated unit tests.

I originally updated PDFRenderer.update to use Dims but this caused CI patch coverage check  to fail because the method is currently not covered by testing. What is the policy on that?

jeromerobert: I would say there are no strong requirement that new PR get covered by CI. I put a hold on some of your PR because I'll have time during the next week to add more functional tests. Your PRs are good opportunities for that. If I do not manage to add those test before the end of next week, I'll merge your PR as is.

Also I would like to avoid to have too many unit tests (i.e. which test low-level functions). My experience is that it make refactoring a nightmare. We should not spend more time to maintain tests than too maintain PDF-Arranger itself. IMHO the target should be 100% coverage on functional tests (dogtail or if not possible or too difficult call of high-level functions) and have unit tests only when it make sense.

jeromerobert:
I'll add a dogtail test for "past as overlay" before merging that PR.

PR #954, opened 2023-08-22
Improve "paste as overlay/underlay" and "Crop margins". Implement a "Hide margins" feature

kbengs:
All use a common DrawingAreaWidget.

### DIFF
--- a/data/menu.ui
+++ b/data/menu.ui
@@ -150,6 +150,10 @@ along with PDF Arranger.  If not, see <http://www.gnu.org/licenses/>.
           <attribute name="action">win.crop</attribute>
         </item>
         <item>
+          <attribute name="label" translatable="yes">_Hide Margins…</attribute>
+          <attribute name="action">win.hide</attribute>
+        </item>
+        <item>
           <attribute name="label" translatable="yes">Crop White Borders</attribute>
           <attribute name="action">win.crop-white-borders</attribute>
         </item>
@@ -381,6 +385,10 @@ along with PDF Arranger.  If not, see <http://www.gnu.org/licenses/>.
       <item>
         <attribute name="label" translatable="yes">_Crop Margins…</attribute>
         <attribute name="action">win.crop</attribute>
+      </item>
+      <item>
+        <attribute name="label" translatable="yes">_Hide Margins…</attribute>
+        <attribute name="action">win.hide</attribute>
       </item>
       <item>
         <attribute name="label" translatable="yes">Crop White Borders</attribute>
--- a/data/menu.ui
+++ b/data/menu.ui
@@ -142,8 +142,12 @@ along with PDF Arranger.  If not, see <http://www.gnu.org/licenses/>.
           <attribute name="target" type="i">90</attribute>
         </item>
         <item>
-          <attribute name="label" translatable="yes">_Page Format…</attribute>
-          <attribute name="action">win.page-format</attribute>
+          <attribute name="label" translatable="yes">_Page Size…</attribute>
+          <attribute name="action">win.page-size</attribute>
+        </item>
+        <item>
+          <attribute name="label" translatable="yes">_Crop…</attribute>
+          <attribute name="action">win.crop</attribute>
         </item>
         <item>
           <attribute name="label" translatable="yes">Crop White Borders</attribute>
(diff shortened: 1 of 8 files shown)


m-holger:
@kbengs 

Is there actually a need  to search for a blank page of the required size? It seems to me that all blank pages are the same, up to scaling and trimming.

Also, what is the need for having a pdf with blank pages prior to exporting / printing?

m-holger:
Thinking about it - is there any need for a temporary file with blank pages at all?

kbengs:
> Is there actually a need to search for a blank page of the required size?

All layerpages that are put on top of this blank page will be placed relative to the non-cropped page edges, so yes the blank page need to be the right size. 

> Also, what is the need for having a pdf with blank pages prior to exporting / printing?

The blank page feature is implemented this way (I guess, it was implemented by Jerome) because it was a easy way to do it which required no changes to the rendering of the page.  Here I am continuing with the same approach.

But you are right, it could probably be done another way so that we would not need any blank temporary files at all. I have to think about it for a while. But I think it will be something for another PR in the future.

m-holger:
Even if you want to have a blank page of the required size to keep the handling of blank page consistent with normal pages, that still does not require having multiple pages of the same size, or having separate files for blank pages of different sizes. Why not keep all the blank pages in a single file?

kbengs:
This has to do with the way description under the thumbnail is created: The page number is read from "npage". 
The description would need to be handled another way if we would have all blank pages in one file. "npage" is maybe not important in "Hide margins" but at least in "Merge pages" it is.

m-holger:
Sorry to keep harping on about this, but I think this should be documented as code comments to make contributing a less steep experience. (Not saying that I practice what I preach re documenting my own code)

m-holger:
Creating multiple booklets and/or inserting multiple blank pages results in a number of "page 1" appearing, Also, if "page 1" is copied and pasted, it also appears as "page 1". Presumably it would make sense to have distinct page numbers?

kbengs:
Yeah documentation could overall be improved. I think someone who reads the code and is not very familiar with it can best see where the code should be better documented.. ;-) 

My initial thought about page numbering is that it is not wrong as it is. Maybe a first line should be added, so that it would be "Booklet 1 \n page 1", "Booklet 1 \n page 2". For the second booklet: "Booklet 2 \n page 1", "Booklet 2 \n page 2". 
But this is another case.

m-holger:
@kbengs , sorry to be a pain, but ....

In the paste as overlay, what is the reasoning for the limits on the offsets.
consider the case with page  height > layer page height, page width < layerpage width

Why is there the requirement  that the entire height of page has to be covered by the overlay, but none of the width overlay can go outside the page width? Why not allow all cases where the page and the overlay intersect?

kbengs:
> but none of the width overlay can go outside the page width

I am not exactly sure what you mean here. I would say that the overlay can go outside of the page width in the case you describe.
The reasons with the limits is that you could easily drag a page to the edge and be sure it "snap" there.  I have considered to have limits beyond 0 and 100%., but I think the in most common use cases it make sense to have limits.

I  haven't, yet, found out why build-pikepdf-1 fails..

m-holger:
I can't run build-pikepdf-1 locally, but build fails with the same error. If I run build locally, it succeeds. 

I ran the test on github with the folling modified test code:
```python
    def _assert_page_size(self, width, height, pageid=None):
        import time
        if pageid is not None:
            for icon in self._icons():  # NEW
                icon.click()
                time.sleep(2)
                print(self._status_text()) # END NEW

            self._icons()[pageid].click()
            print(self._status_text()) # NEW
            self._wait_cond(lambda: self._status_text().startswith(f"Selected pages: {pageid+1}"))
        label = " {:.1f}mm \u00D7 {:.1f}mm".format(width, height)
        self.assertTrue(self._status_text().endswith("Page Size:" + label))
```
I got the output
```
Clicking on [icon | ]
  Mouse button 1 click at (330.0,299.5)
  Selected pages: 2 / 2 | Page Size: 489.0mm × 212.2mm
  Clicking on [icon | ]
  Mouse button 1 click at (330.0,454.5)
  Selected pages:  / 2
  Clicking on [icon | ]
  Mouse button 1 click at (330.0,299.5)
  Selected pages: 2 / 2 | Page Size: 489.0mm × 212.2mm
```
When I run it locally I get erratic results - sometimes it succeeds and sometimes it behaves as if you click on the zoom in button rather than the page icon.

It looks to me like it might be a dogtail test issue and/or a issue with the status message.

jeromerobert:
`self._icons()[pageid].click()` fails because for some reason the iconview is not clickable at the time this line of code is called. As far as remember we have a transparent panel ontop the iconview when processing is running. So we need a condition to check that this panel is gone before clicking.

jeromerobert:
May be this:

```patch
--- a/tests/test.py
+++ b/tests/test.py
@@ -227,6 +227,8 @@ class PdfArrangerTest(unittest.TestCase):
 
     def _assert_page_size(self, width, height, pageid=None):
         if pageid is not None:
+            viewport = self._app().child(roleName="layered pane")
+            self._wait_cond(lambda: viewport.sensitive)
             self._icons()[pageid].click()
             self._wait_cond(lambda: self._status_text().startswith(f"Selected pages: {pageid+1}"))
         label = " {:.1f}mm \u00D7 {:.1f}mm".format(width, height)
```

kbengs:
Thanks.

Jeromes suggested change did not fix the problem. I also tried to add a `viewport.grabFocus()`. But no.

I can try to improve the coverage later but it might take some weeks.

PR #955, opened 2023-08-22
Add unit tests for BasePage.rotate_times and other methods

m-holger:


m-holger:
I am not sure that the rounding of angles makes much sense. Anything other than an integral multiple of 90 is illegal and QPDF and I believe pretty much everybody else will treat them as 0 when rotating a page. 

To me it would feel more sensible to treat illegal angles as 0.

PR #957, opened 2023-08-27
Change Page.crop and LayerPage.offset to named tuples

m-holger:


pdfarranger/pdfarranger.py
```diff
@@ -1573,7 +1574,7 @@ def paste_as_layer(self, data, destination, laypos, offset_xy=None):
                     lp.offset[i] += outside
```
kbengs, line 1574: The lines lp.crop[i] +=  and lp.offset[i] +=  cause "TypeError: 'Sides' object does not support item assignment"
To test this code you need to copy and paste as overlay a page which already has a overlay.

jeromerobert: Thank you, I'll add a test for that

m-holger: Sorry about this. #960 should fix it

PR #959, opened 2023-08-29
dogtail test: paste page with overlay as underlay

jeromerobert:
The lines `lp.crop[i] +=`  and `lp.offset[i] +=`  cause `"TypeError: 'Sides' object does not support item assignment"`
To test this code you need to copy and paste as overlay a page which already has a overlay.

_Originally posted by @kbengs in https://github.com/pdfarranger/pdfarranger/pull/957#discussion_r1308133936_

Detect a regression in #957 

@m-holger any opinion on how to fix it ?


m-holger:
> any opinion on how to fix it ?

Yes, I'll do a PR today

PR #961, opened 2023-08-30
Fix reversion in #950

m-holger:
To reproduce:
- open tests/test.pdf
- rotate page 2
- duplicate page 1
-  undo

Page 2 is not rendered correctly


kbengs:
I found one more regression, not sure where it was introduced:
* open tests/test.pdf
* Merge the two pages
* Export

The green rectangle is lost.

Pikepdf  8.4.0

m-holger:
I'll look into it

PR #962, opened 2023-09-01
Fix regression in #957

m-holger:
To reproduce:

- open tests/test.pdf
- Merge the two pages
- Export

The green rectangle is lost.

Reported by @kbengs in #961

### DIFF
--- a/pdfarranger/core.py
+++ b/pdfarranger/core.py
@@ -879,7 +879,7 @@ class PDFRenderer(threading.Thread, GObject.GObject):
             if rotation > 0:
                 cr.translate(*lp.size.scaled(0.5))
                 cr.rotate(rotation * pi / 180)
-                cr.translate(*lp.size.scaled(-0.5))
+                cr.translate(*lp.size_orig.scaled(-0.5))
             self.render(cr, lp)
             cr.restore()
 
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -1551,10 +1551,12 @@ class PdfArranger(Gtk.Application):
             dwidth, dheight = dpage.size[0] * dpage.scale, dpage.size[1] * dpage.scale
             scalex = (dpage.width_in_points() - lp0.width_in_points()) / dwidth
             scaley = (dpage.height_in_points() - lp0.height_in_points()) / dheight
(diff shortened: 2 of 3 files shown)


m-holger:
@kbengs

I believe that the PR fixes the issue, could you please verify.

Dogtest batch 2 fails with the fix. I think this is a problem with the test. When I go through the test sequence manually, with the fix the page 1 output matches the rendered image on screen as well as the output produced by the flatpack version of PDFArranger. Without the fix, page 1 output is different.

Is there an idiot's guide to running the dogtest suite locally somewhere?

kbengs:
It appears to fix the problem!

If you want one more I have one: :-)
To reproduce:

* Open tests/test.pdf
* Rotate the green rectangle  to the left
* Merge the two pages 

The green rectangle does not look as it should in the iconview

I run the test with:
`rm ~/.config/pdfarranger/config.ini`
`python3 -X tracemalloc -u -m unittest -v -f tests.test`

For me it fails at test_03_booklet (tests.test.TestBatch5) But this might be some problem on my machine only...

m-holger:
> I run the test with: ...

Thanks. That was easier than expected. From testing.md I was under the impression that it required to be run in a docker image.

> If you want one more I have one: :-)  ...

That seems to have been a local problem. It passed for me.

I slightly changed the file size for the output for batch2. I think the batch works correctly but someone needs to verify it.

jeromerobert:
> Is there an idiot's guide to running the dogtail suite locally somewhere?

To run in docker/podman: https://github.com/pdfarranger/pdfarranger/blob/main/TESTING.md#dogtail. This is good to reproduce what happen in Github Action but you won't see the UI.

Some tips when running out of docker: https://github.com/pdfarranger/pdfarranger/blob/main/tests/test.py#L36-L40

> Dogtail batch 2 fails with the fix. I think this is a problem with the test. 

I check the size of the exported files to check it has the expected number of triangles, that is the "past as overlay" added a second triangle to the page. The tolerance may be too small but this may also be suspect.

m-holger:
I fixed the line in PDFRenderer and added some code to explicitly check for the expected triangles

m-holger:
@kbengs 

>    If you want one more I have one: :-) ...
> That seems to have been a local problem. It passed for me.

Sorry, that was meant as a response to 

> For me it fails at test_03_booklet (tests.test.TestBatch5)

PR #965, opened 2023-09-04
Support both new setuptools and old distutils

Cimbali:
- Old distros still didn’t update their setuptools and we still want to build for them
- New/rolling distros officially deprecated distutils

This works for both setups without deprecations/warnings.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,11 +22,14 @@
 
 from setuptools import Command
 from setuptools import setup
-# for the time being do not use setuptools.command because:
-# * setuptools is to old in many distro (ex: Ubuntu 22.04)
-# * The Fedora package relies on wheel so it would not contains translation and
-#   icons
-from distutils.command.build import build
+from setuptools import __version__ as setuptools_version
+
+# support distros that ship old setuptools
+setuptools_version = tuple(int(n) for n in setuptools_version.split('.')[:2])
+if setuptools_version < (65, 2):
+    from distutils.command.build import build
+else:
+    from setuptools.command.build import build
 
 from os.path import join
 import glob


Cimbali:
It seems CI errors are unrelated to this PR except for the slight decrease in coverage. Essentially just an import and parsing a version string.

Note that the version string parsing only uses the 2 first (major, minor) numbers, as that is all we need, and to avoid breaking on alpha/beta versions and strings like `51.1.0.post20201221`.

jeromerobert:
Yes, it looks like github action had some failure, but there seems to be a real problem with rpmbuild. Unless I'm wrong, it's the only place where we check that the translations are properly installed. I'll be looking at it in the next few hours.

jeromerobert:
Ah it's in the comment you removed. Using setuptools breaks wheels because it forbid to have data files outside of `site-packages` (see #860).

jeromerobert:
Indeed I was wrong, it seems to work properly.

@dreua Do you understand why `python3 setup.py bdist_wheel` does not embed `*.mo` files in rpmbuild ? When I run the same command locally on Debian `*.mo` files are included in the `*.whl` file.

Cimbali:
I thought the error was just `%changelog not in descending chronological order`, but you’re right, translations seem to be missing. It’s strange because the log says explicitly:
```
+ /usr/bin/python3 setup.py bdist_wheel
running bdist_wheel
running build
running build_py
```

I think it’s `build.sub_commands` that breaks with this PR.

Cimbali:
Doesn’t seem to be that, I’ve tried with all implicated versions of setuptools, the command always gets called.

Cimbali:
I tried to run the rpmbuild from various vanilla Fedora containers, and it always seems to build the translation files fine: https://github.com/Cimbali/pdfarranger/actions/runs/6112654784

(Except on Fedora 31 which fails due to `pdfshuffler` not being listed)

https://github.com/pdfarranger/pdfarranger/blob/0a816457c2ae4e6cc9632a528e00213d3a1ad5f5/pdfarranger.spec#L91-L93

jeromerobert:
Indeed this is failing to install `*.mo` files:

```
podman run -v $PWD:$PWD -w $PWD docker.io/dreua/pdfarranger-docker-rpm bash -c "yum install -y gettext python3-devel python3-setuptools python3-wheel python3-pip && pip install -v ."
```

while

```
podman run -v $PWD:$PWD -w $PWD docker.io/fedora:38 bash -c "yum install -y gettext python3-devel python3-setuptools python3-wheel python3-pip && pip install -v ."
```
succeed. That's surprising because https://github.com/pdfarranger/pdfarranger-docker-ci/blob/main/rpmbuild/Dockerfile doesn't contain anything odd.

jeromerobert:
Even:

```
podman run -v $PWD:$PWD -w $PWD docker.io/fedora:38 bash -c "dnf -y install 'dnf-command(builddep)' rpm-build rpmdevtools && dnf -y builddep pdfarranger && pip install -v ."
```

works.

jeromerobert:
And if I rebuild https://github.com/pdfarranger/pdfarranger-docker-ci/blob/main/rpmbuild/Dockerfile from source it also works. So it's definitly a transient bug in Fedora. @dreua will just have to rebuild and push a new image to fix the issue.

Issue #972, opened 2023-09-16
Github Action rpmbuild image need to be rebuild


jeromerobert:
Indeed this is failing to install `*.mo` files:

```
podman run -v $PWD:$PWD -w $PWD docker.io/dreua/pdfarranger-docker-rpm bash -c "yum install -y gettext python3-devel python3-setuptools python3-wheel python3-pip && pip install -v ."
```

while

```
podman run -v $PWD:$PWD -w $PWD docker.io/fedora:38 bash -c "yum install -y gettext python3-devel python3-setuptools python3-wheel python3-pip && pip install -v ."
```
succeed. That's surprising because https://github.com/pdfarranger/pdfarranger-docker-ci/blob/main/rpmbuild/Dockerfile doesn't contain anything odd.

_Originally posted by @jeromerobert in https://github.com/pdfarranger/pdfarranger/issues/965#issuecomment-1722261184_
            
            
And if I rebuild https://github.com/pdfarranger/pdfarranger-docker-ci/blob/main/rpmbuild/Dockerfile from source it also works. So it's definitly a transient bug in Fedora. @dreua will just have to rebuild and push a new image to fix the issue.

_Originally posted by @jeromerobert in https://github.com/pdfarranger/pdfarranger/issues/965#issuecomment-1722264999_

dreua:
Done, sorry I'm currently unable to do much here.

PR #985, opened 2023-09-30
Fix iconview in right-to-left locale

kbengs:
This fixes:
* Drag and drop of pages
* Dragging in a file
* Selecting pages by clicking and dragging mouse
* Navigating and selecting with keyboard arrows

### DIFF
--- a/pdfarranger/iconview.py
+++ b/pdfarranger/iconview.py
@@ -171,9 +171,15 @@ class IconviewCursor(object):
             step = 0 if self.cursor_page_nr + columns_nr > len(self.model) - 1 else columns_nr
             self.cursor_page_nr_new = self.cursor_page_nr + step
         elif self.event.keyval == Gdk.KEY_Left:
-            self.cursor_page_nr_new = max(self.cursor_page_nr - 1, 0)
+            if self.iconview.get_direction() == Gtk.TextDirection.LTR:
+                self.cursor_page_nr_new = max(self.cursor_page_nr - 1, 0)
+            else:
+                self.cursor_page_nr_new = min(self.cursor_page_nr + 1, len(self.model) - 1)
         elif self.event.keyval == Gdk.KEY_Right:
-            self.cursor_page_nr_new = min(self.cursor_page_nr + 1, len(self.model) - 1)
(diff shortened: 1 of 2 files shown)


kbengs:
Looks like `./setup.py build` does not create `build/mo` since 3477ee615d089483d36337a9c1adeda4d97b3df3

Tried to "fix" it with 307958c2dc0c29779b1f3f4e160bb9a325c978f0 but I guess os.name does not return 'nt" here

Any ideas how to fix it properly?

jeromerobert:
> Looks like `./setup.py build` does not create `build/mo` since [3477ee6](https://github.com/pdfarranger/pdfarranger/commit/3477ee615d089483d36337a9c1adeda4d97b3df3)

I'm on it.

jeromerobert:
This is very odd:
- https://github.com/pdfarranger/pdfarranger/commit/3477ee615d089483d36337a9c1adeda4d97b3df3 built one time properly: https://github.com/pdfarranger/pdfarranger/actions/runs/6208430514/job/16936487582
- I cannot reproduce the issue when running in Docker locally with the very same Docker image
- In Github Action https://github.com/pdfarranger/pdfarranger/blob/3477ee615d089483d36337a9c1adeda4d97b3df3/setup.py#L103 seems to have no effect
- It remind me https://github.com/pdfarranger/pdfarranger/issues/972 but I tried to upgrade setuptools and python with no luck
- A workaround is to explicitly invoke `./setup.py build_i18n`. I'll push that if I cannot quickly find good fix.

jeromerobert:
This is the same bug as https://github.com/pypa/setuptools/issues/4040. Any package can override the distutils or setuptools build command. In our case the culprit is also distutils-extra because it's still in the build Docker image even if we no longer use it. In general we cannot assume that `./setup.py build` will be either `setuptools.command.build` or `distutils.command.build`. It can be anything. I'm looking for a clean solution.

jeromerobert:
https://github.com/pdfarranger/pdfarranger/pull/986

kbengs:
Tested in Ubuntu 23.04 with language set to Hebrew. As far as I can see it works.

PR #987, opened 2023-10-03
Add missing release to metainfo

dreua:
Flathub shows the wrong version without this data, but I guess this is not worth a bugfix release or anything. It would be nice to have it right in the coming releases though, do we have a release checklist somewhere?


soumyaDghosh:
Same issue is with the snap. But, I can force the snap to show the correct version.

PR #998, opened 2023-10-22
GHA: Compute coverage in only one job

jeromerobert:
This might create race as we don't know how codecov is interpreting this


jeromerobert:
Wrong way. I'll switch to https://about.codecov.io/blog/uploading-code-coverage-in-a-separate-job-on-github-actions/

Issue #927, opened 2023-07-29
Copy text, copy image to clipboard


jeromerobert:
### Discussed in https://github.com/pdfarranger/pdfarranger/discussions/905

<div type='discussions-op-text'>

<sup>Originally posted by **JunglePredator** June 26, 2023</sup>
I don't know if it's what you intend for the scope of the project but integrating some of poppler's features to pdf arrange would make it a better all in one solution.

Like right now you can only extract a page to another pdf; with poppler integration it'd allow users to extract to text or the images on a page etc.

It of course would add quite a bit of work for things users can already do; it'd make the work flow a lot nicer however.

Thought it'd be nice to see what other users think.</div>

jeromerobert:
https://lazka.github.io/pgi-docs/Poppler-0.18/classes/Page.html#Poppler.Page.get_text
https://lazka.github.io/pgi-docs/Poppler-0.18/classes/Page.html#Poppler.Page.get_image_mapping
https://lazka.github.io/pgi-docs/Poppler-0.18/classes/Page.html#Poppler.Page.get_image
https://lazka.github.io/pgi-docs/index.html#Gdk-4.0/functions.html#Gdk.pixbuf_get_from_surface

_Originally posted by @jeromerobert in https://github.com/pdfarranger/pdfarranger/discussions/905#discussioncomment-6391913_

kbengs:
Was implemented in https://github.com/pdfarranger/pdfarranger/pull/1002

PR #1002, opened 2023-11-15
Copy image and text from pdf

kbengs:
Discussed in https://github.com/pdfarranger/pdfarranger/discussions/905 and https://github.com/pdfarranger/pdfarranger/issues/927

If a page has several images the idea is to use "Explode into Images" first. The images are then added as new pages to model. After that you can select the page you want to copy the image from.


kbengs:
Fixed missing indentation and added handling of very small and large images

kbengs:
Works as far as I can see

PR #1010, opened 2023-12-02
Fix reversed page order when dragging in file "after"

kbengs:


kbengs:
When adding pages after a page the pages need to be reversed. This is now done in the commit() method. The old approach was to either avoid adding after by converting to a "add before" action, or the pages were reversed earlier.

PR #1012, opened 2023-12-03
Add a keyboard shortcut to open the popup menu

### LINKED ISSUE #915: Can't open page context menu via keyboard
While translating and testing accelerators I noticed that I can no longer open the context menu (right click) of a page without using the mouse. Tried the context menu key (between right super and ctrl), different F-keys, also with modifiers. I guess this was possible in a previous release but I rarely use this feature so it might have vanished anywhere between PDF Arranger versions or Gnome / OS updates.

Any pointers appreciated, once I find the time I might test older versions :thinking:

### LINKED ISSUE #1007: Open the right-click context menu when "Document key" is pressed on the keyboard
Right now the only way to open the context menu is to right-click on the pages of PDF. I want to request an additional way to do this. 

When we select a page in the opened PDF file and press the Document key/ Right-click key then it should open the context menu.

Here is the mentioned key on the keyboard
image

And it should open this context menu
image

jeromerobert:
* Close #915
* Close #1007

### DIFF
--- a/pdfarranger/config.py
+++ b/pdfarranger/config.py
@@ -58,7 +58,7 @@ _DEFAULT_ACCELS = [
     ('select(1)', '<Primary><Shift>a'),
     ('main-menu', 'F10'),
     ('metadata', '<Alt>Return'),
-    ('context-menu', '<Shift>F10 Multi_key'),
+    ('context-menu', '<Shift>F10 Menu'),
 ]
 
 
@@ -128,7 +128,8 @@ class Config(object):
 
     def is_popup_key_event(self, keyevent):
         for key, mods in self.popup_menu_accels:
-            if keyevent.state == mods and keyevent.keyval == key:
+            # `& mods` lets us ignore all other modifiers (e.g. numlock)
+            if keyevent.state & mods == mods and keyevent.keyval == key:
                 return True
         return False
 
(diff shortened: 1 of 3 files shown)


jeromerobert:
Setting to draft until I add a test and fix coverage.

dreua:
How would you feel about changing the name to context menu and the default key to Multi_key? That is how I believe most apps work or at least how I expect the menu key to work. I can dontbe changes if you approve.

jeromerobert:
Done. I've added `Multi_key` and left `shift+F10` as most laptop don't have a `Multi_key`.

dreua:
Great, I'll test and merge asap, thank you!

dreua:
For some reason it doesn't work for me, nothing happens when pressing any of the keys. Even tried https://github.com/pdfarranger/pdfarranger/commit/9299e8c94f4acb7e8aa06db08a4877b28bd32b76 but the same results.

Not sure whether this is due to differences on my system or if I'm missing something.

dreua:
Okay I think I'm on it: Numlock on is also a modifier `GDK_MOD2_MASK` so we need to check only whether shift modifier is present not if the current modifier is exactly and only shift. For the different keycode of my menu key further research is needed.

dreua (approved):
Approved with minor fixes, thanks you!

PR #1020, opened 2023-12-27
Fix typos again

kianmeng:
Found via `codespell -S po,*.pdf -L te`


kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #1022, opened 2023-12-27
Fix a "raise KeyError(key) from None" in pikepdf

### LINKED ISSUE #1019: Unable to Save PDF with Edits
Describe the bug
----------------

Hello, I was unable to save any edits made to certain PDFs in PDFarranger today. I was able to complete my edits in the application (e.g., rotate a single page, duplicate a page) however, when attempting to save this edited PDF, I would receive a pop-up notification warning: "Forms and outlines are lost on saving." This in itself has not been a problem for me. However, this was followed by another pop-up notification with a series of letters and numbers (see two examples below) and only the option to click "Ok" of close that window; neither of which led to saving the edited file.  

So far, the only work around I have found is to "print" the target PDF to a new PDF. Then I can successfully edit and save that new "printed" file. 

To Reproduce
------------

1.  Go to the target PDF file in a file manager.
2.  Right click on target PDF file, select to open with PDFarranger. 
4.  Edit the file (e.g., rotate a page) and click Save or Save As.
5.  Receive the pop-up notifications and cannot save edited file. 

Expected behavior
-----------------

Successfully save the PDF file with edits retained.

Input files
-----------

Only appears to affect _some_ PDF files. I cannot attach an example here because they are copyrighted academic journal articles. However, I confirmed that the PDF files that cannot be saved after editing are PDF v. 1.7 and the PDF files that _can_ be saved after editing are PDF v. 1.3 if that is useful.

Screenshots
-----------

Screenshot from 2023-12-26 14-16-57
Screenshot from 2023-12-26 14-38-04
Screenshot from 2023-12-26 14-04-04
Screenshot from 2023-12-26 14-15-25

System and Versions
-------------------

-   PDF Arranger 1.10.1 (flatpak)
-   pikepdf 8.10.1 (installed with PDF Arranger flatpak)
-   Linux Mint 21.2 Cinnamon. 

I greatly appreciate any and all help. 

Thank you.

jeromerobert:
* At pikepdf/models/metadata.py:706
* Close #1019

### DIFF
--- a/pdfarranger/metadata.py
+++ b/pdfarranger/metadata.py
@@ -87,6 +87,9 @@ def _safeiter(elements):
             break
         except ValueError:
             traceback.print_exc()
+        except KeyError:
+            # Workaround for https://github.com/pdfarranger/pdfarranger/issues/1019
+            pass
 
 
 def merge_doc(metadata, input_docs):


dreua (approved):
lgtm

PR #1024, opened 2024-01-01
Setup dependabot for github actions

dreua:


jeromerobert:
I'm not sure to understand what this is supposed to do ? Could you explain ?

dreua:
I shouldn't have created this on my phone, sorry. I hope this clears it up:

> Setup dependabot for github actions
> 
> Check the used actions (e.g. checkout, upload-artifacts) for new
> versions and open PRs with updates.

Here is an example: https://github.com/cntools/cnping/pull/108

jeromerobert:
Indeed, that was needed.

PR #1042, opened 2024-01-27
Improve handling of scaling and rotation at printing

### LINKED ISSUE #1015: Print from PDF Arranger ignores page orientation
Describe the bug
----------------

To Reproduce
------------

Print a PDF from within PDFarramger with portrait and landscape pages

Expected behavior
-----------------

Pages are printed with the orientation like in the PDF

Actual behaviour
-----------------

Landscape oriented pages are cut and printed in portrait orientation

System and Versions
-------------------

-   1.10.1 (flatpak)
-   debian 12

kbengs:
This is needed to fix #1015 imo

I am not planning to write tests for this now.

### DIFF
--- a/pdfarranger/config.py
+++ b/pdfarranger/config.py
@@ -22,6 +22,8 @@ import gettext
 from gi.repository import Gdk
 from gi.repository import Gtk
 
+from .exporter import PrintSettingsWidget
+
 _ = gettext.gettext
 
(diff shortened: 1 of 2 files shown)


kbengs:
Also on flatpak print dialog the "Page Handling" tab is not visible. But there is not much we can do about that as I see it. At least the options are available in the preferenses dialog. 

Evince has a open bug about this too. 
https://gitlab.gnome.org/GNOME/evince/-/issues/1015
https://gitlab.gnome.org/GNOME/evince/-/merge_requests/394

PR #1043, opened 2024-01-28
Show content loss warning on import

kbengs:
This does also not check for what content is in the pdf, which will speed up adding of large pdfs when "content_loss_warning" is enabled.

What do you think about this? Also @m-holger

### DIFF
--- a/pdfarranger/core.py
+++ b/pdfarranger/core.py
@@ -32,6 +32,7 @@ import pathlib
 import shutil
 import tempfile
 import threading
+import pikepdf
 import packaging.version as version
 from typing import NamedTuple, Optional, Tuple, Union
 import gettext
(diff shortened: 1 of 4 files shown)


m-holger:
I thought forms are preserved for pikepdf 8.

To my mind the general warning is sufficient. I don't feel that checking the input file for outlines or forms as a condition for triggering the warning adds much, especially if we assume that most users will tick the "don't show this dialog again" box. 

(Going beyond the scope of this PR, a help option displaying a short help file might be useful. This could eg contain a list of operations that preserve or destroy outlines and links.)

kbengs:
> I thought forms are preserved for pikepdf 8.

Yeah, even pikepdf 2.7.0 preserve the forms on all example pdfs I have tested... Does anyone have a pdf where forms will not be preserved?

raph82:
> Does anyone have a pdf where forms will not be preserved?

I have a pdf with a form that is not preserved by pdfarranger, I haven't looked up how to test it with pikepdf yet.
The attached form come from here: https://www.dispositif-rexbp.com/ressource/fiche-de-reception-menuiserie-exterieure
First I opened the source pdf with xed and the form is editable.
Then I opened the source pdf in pdfarranger 1.10.1, duplicated one page and saved. There are two warnings: "forms and bookmarks are not preserved" and "The metadata field /Trapped with value '/False' has no XMP equivalent, so it was discarded".
Then I opened the result pdf and got a "flat" pdf with no more form fields.
[F-Reception-Menuiserie-Ext_0.pdf](https://github.com/pdfarranger/pdfarranger/files/14088962/F-Reception-Menuiserie-Ext_0.pdf)

kbengs:
@raph82 thank you but I can not reproduce the issue. I tested the pdf you attached (which does have editable forms) with several versions of pdfarranger/pikepdf. I used evince and okular to view the pdf. Is xed a pdf viewer, do you have a link?

m-holger:
This looks like a linearized file to me. I don't recall PDF Arranger writing linearized files.

raph82:
Sorry all, I'm not sure if I was very clear...

The problem I have is not with the input pdf, the one linked in the previous post, but with the pdfarranger output of this pdf, after duplicating one of the pages. [Here is that pdf output](https://github.com/pdfarranger/pdfarranger/files/14104725/pdfarranger.output.pdf).
And I wrote xed when I meant [xreader](https://github.com/linuxmint/xreader/), the Linux Mint default pdf reader, sorry again :(

While double-checking, I opened the linked pdf arranger output with firefox and it _has_ editable form fields like I wanted! So maybe it's a xreader bug/limitation, which only display form fields from the original file and not with the pdfarranger-modified file. I'll check at work with another pdf reader!

Many thanks for your replies!

PR #1045, opened 2024-01-31
Add MacOS support

Endle:
1) Load libintl.8.dylib on mac
2) Check GTK schema


dreua (commented):
I think this needs a quick test on windows, otherwise looks good to me, thanks for the contribution!

Would you like to add a few lines to the readme as well?

kbengs:
It does not start on Windows because of `check_gtk_schema_exists() `. Is that needed on Linux and Windows?

Endle:
Thank you. I don’t know the mechanism behind it. How about changing from raising an exception into printing a warning?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

kbengs:
Yes I think so and run the check only on "Darwin", if it is a potential problem there?

pdfarranger/pdfarranger.py
```diff
@@ -66,14 +67,21 @@
...
 DOMAIN = 'pdfarranger'
 ICON_ID = 'com.github.jeromerobert.' + DOMAIN
+
+def get_libintl_path():
+    f = 'libintl.so.8'
+    if os.name == 'nt':
+        f = 'libintl-8'
+    if platform.system() == 'Darwin':
+        f = 'libintl.8.dylib'
+    return f
```
dreua, line 77: Can we get rid of the variable `f` here and return directly? Reorder the statements if overwriting was part of the logic but I guess that's not necessary here.
Also add an empty line after the function for readability.

pdfarranger/pdfarranger.py
```diff
@@ -119,6 +127,14 @@
...
 from .core import Sides, _img_to_pdf
 
+def check_gtk_schema_exists():
+    schemas = subprocess.run(["gsettings", "list-recursively"],
+        stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+        check=True,
+        text=True)
+    return 'org.gtk.Settings.ColorChooser' in schemas.stdout
+if not check_gtk_schema_exists():
+    raise Exception('Found no schema files. You may need to set GSETTINGS_SCHEMA_DIR.')
```
dreua, line 137: I totally understand that you added this to help/guide users with the same issue. However we don't want it to break stuff with tests that would be fine without therefore we should either:

- Run it only on darwin if it is a common enough problem here
- Make it a warning only. On windows where it is not supposed to work ever we still shouldn't run it because spamming warnings that are unfixable and not helpful are still a bad idea imo.

Feel free to do a combination of both.

dreua (changes requested):

Endle:
Thank you @dreua 

I totally agree with your comments. I'm sorry that I'm a bit distracted in this week. I'll modify my code and add a short note in the README when I have time.

dreua:
Could you rebase the branch on current main? Ideally also getting rid of the merge commits? Jerome recently changed something in the tests maybe that already fixes the failed test here.

kbengs:
I still think this should be darwin only, .ie. `if platform.system() == 'Darwin' and not check_gtk_schema_exists():`

There is not much reason to run it on Linux. It will slow down upstart of application (by 37ms).

Endle:
I'm sorry for my delay. I've updated my PR. May I ask for review it again?

@dreua

cerenno:
Hello. Is it possible to make this into a dmg for mac? I tried all your instructions and can't get it to work. Have also been trying VMs in UTM to try to use it in Linux and can't get it to work. I'm really struggling with this. If there's anything you could do it would be greatly appreciated.

Endle:
@cerenno How do you set up dev env on your mac? Nix/homebrew/macPort/etc.. ?

cerenno:
@Endle I use homebrew. I'm not a programmer though. I've had a bit of help but didn't succeed in getting it to work. I think I managed to export the variables and stuff but when it came to using the pip/python there was no module for pdfarranger or similar errors along those lines.

Endle:
@cerenno I don't use homebrew but I know it is the most popular option. I'll try to find a method to run pdfarranger with homebrew. Please give me a few days

cerenno:
@Endle Alright. Thank you for taking the time to try. I'll wait for your response.

Endle:
@cerenno can you try the following homebrew & pip dependencies?

```
git clone https://github.com/Endle/pdfarranger.git
cd pdfarranger; git checkout mac
pip3 install --user pikepdf PyGObject python-dateutil img2pdf
brew install gettext
	# msgfmt: command not found
ln -s /opt/homebrew/lib/libintl.8.dylib .
brew install  pygobject3 gtk+3 poppler
python3 -m pdfarranger
```

cerenno:
@Endle Hello. Wow thank you so much! It worked seamlessly. I can't thank you enough for this. One question, for me to open it do I have to type on the terminal `cd pdfarranger` followed by `./setup.py build python3 -m pdfarranger` everytime I want to use it or is there a way to automate it or make it into an icon that you can click and it opens?

Endle:
@cerenno Sorry, I don't know packaging in macOS. And I don't know how to send a package request for homebrew (all I can find is https://github.com/orgs/Homebrew/discussions/533)

For now, I think you can set `alias` for pdfarranger
1. run `echo $0` to tell if your terminal is bash or zsh (ref https://www.moncefbelyamani.com/which-shell-am-i-using-how-can-i-switch/)
2. If you're using bash, then edit `~/.bash_profile`. If you're using zsh, edit `~/.zprofile`
3. Add `alias pa='cd xxx; python3 -m pdfarranger'` into this file
4. restart the terminal

Now you should be able to just type `pa` to start pdfarranger

cerenno:
Hello. Sorry for the delay. It didnt work and adding the alias adjustment makes the app not open, needing to reinstall with the git clone command. When using the command it says that theres no module named pdfarranger and directly after editing the alias, it said something like pdfarranger is an executable or something like that.

kbengs:
This looks good to me. I also tested in Windows. Are you @dreua also happy with this?

dreua:
Yes, sorry for the late response!

I squashed everything in one commit in order to keep our git history small, hopefully that is okay with you. Thanks for your work @Endle!

Endle:
@dreua Thank you for reviewing and squashing them!

Yes, definitely I'd prefer to squash merge this PR.

PR #1048, opened 2024-02-04
update screenshot

### LINKED ISSUE #170: New screenshot
We need a new screenshot with the new look & feel. It should match the [freedesktop recommendation](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#qsr-app-screenshots-info), so have a 16:9 ratio (the current one is 4:3).

bertob:
Closes #170


dreua:
I don't get why the CI keeps failing here. Screenshot looks good, thanks!

dreua (changes requested):
The ci is failing because one of the tests uses the screenshot and probably fails due to a size difference or something. (Not a great idea to start with, I admit.) Is anyone interested in investigating further? Otherwise I propose to readd and rename the old screenshot and pointing the test to the old version. Feel free to move it to another directory if appropriate.

jeromerobert:
I'll have a look at it.

dreua (approved):
With all tests green now I'd say that is good to be merged? Thanks @jeromerobert !

Issue #1038, opened 2024-01-18
Better appdata


bertob:
Since there was a request to re-review the metadata and I didn't see any changes, here's some more detailed/actionable feedback:

- The summary `PDF Merging, Rearranging, Splitting, Rotating and Cropping` is too long (35 characters max) and should use sentence case capitalization. There's no need to list every feature here, you can just focus on some of the important ones and leave the rest for the description, e.g. `Merge, shuffle, and crop PDFs`
- The screenshots should be taken using the "window screenshot" feature of the shell screenshot tool, that way they get the native shadow with transparency
- On the screenshot the window controls should be the platform default (close-only in this case)

dreua:
> Since there was a request to re-review the metadata

This is about flathub, right? If I requested this accidentally today the UX is seriously broken.

Furthermore I will not do anything for flathub until they acknowledge and fix the error in their build system: https://github.com/flathub-infra/buildbot/pull/19

dreua:
@bertob thanks for your otherwise useful feedback and please don't take my frustration with the flatpak/flathub ecosystem and the surrounding culture personally. Maybe I need a break from flatpak maintenance.

jeromerobert:
> Since there was a request to re-review the metadata and I didn't see any changes

Could you elaborate ? Which request ?

> The summary PDF Merging, Rearranging, Splitting, Rotating and Cropping is too long (35 characters max) and should use sentence case capitalization. There's no need to list every feature here, you can just focus on some of the important ones and leave the rest for the description, e.g. Merge, shuffle, and crop PDFs

Ok ! Let's go with `Merge, shuffle, and crop PDFs`. This must also be changed here: https://github.com/pdfarranger/pdfarranger/blob/7b73781bd8ac98d86bc6f8a39fae5ad04acd07ee/data/com.github.jeromerobert.pdfarranger.desktop#L13-L35

> The screenshots should be taken using the "window screenshot" feature of the shell screenshot tool, that way they get the native shadow with transparency

"shell screenshot tool" ? I guess you mean that the screenshot must be taken using a Gnome desktop. PR welcomed (I'm not using Gnome).

> On the screenshot the window controls should be the platform default (close-only in this case)

Sorry but here I don't understand what you mean.

bertob:
Sorry, forgot to link the new app metadata documentation for context: https://docs.flathub.org/docs/for-app-authors/appdata-guidelines/quality-guidelines

[...]

PR #1057, opened 2024-02-17
Add developer appstream data

dreua:
Required to fix builds on flathub.

The value is just a suggestion, we can also name ourselves "The PDF Ararnger Project", foobar or list one or more names of the actual developers. Let me know what you think, I don't really care as long as it builds :shrug:


dreua:
Closes: #1038 (together with #1048)

PR #1060, opened 2024-02-23
Add more languages

mapi68:


dreua:
I think there is something wrong with this PR

jeromerobert:
@mapi68 I guess that's this is automated translation. Which tool did you use ?

mapi68:
ChatGPT

jeromerobert:
I'm sorry but there are by far too many mistakes in this PR. You removed Name and change Comment (so all translations are wrong) in the .desktop for example.  Also please don't bump the version.

mapi68:
But now .desktop file is complete

Screenshot_20240224_184648

dreua:
I'm against using AI generated translations, I don't trust those systems and there is no way I could verify them and I don't think you could or have verified them either. Thanks for the initiative but I'm against merging this.

PR #1066, opened 2024-03-02
Add standard paper sizes to "insert blank page" and "page size" dialogs

kbengs:
Old - New
old -new
Scaling to a paper size is what I usually want to do. The old (fit to) "width" and "height" can be achieved by setting for example width to desired value and height to a large value.

As the translation files was updated, this can wait to next release if you like to.

### DIFF
--- a/pdfarranger/core.py
+++ b/pdfarranger/core.py
@@ -306,7 +306,7 @@ class BasePage:
 
 
 class Page(BasePage):
-    def __init__(self, nfile, npage, zoom, copyname, angle, scale, crop: Sides, hide: Sides, size_orig: Dims, basename, layerpages):
+    def __init__(self, nfile, npage, zoom, copyname, angle, scale, crop: Sides, hide: Sides, size_orig: Dims, description, layerpages):
         super().__init__(nfile, npage, copyname, angle, scale, Sides(*crop), size_orig)
         self.zoom = zoom
         self.hide = Sides(*hide)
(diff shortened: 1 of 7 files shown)


dreua:
Support for common pape sizes is a great improvment imo, thanks for the initiative!

> Scaling to a paper size is what I usually want to do. The old (fit to) "width" and "height" can be achieved by setting for example width to desired value and height to a large value.

Haven't tried your branch yet but from reading this sounds not ideal for the user experience. Had to read that twice to even understand it. How hard would it be to change the other value based on the value a user just changed manually? We don't support disproportionate scaling at all, do we? (I mean stretching in one dimension more or less than in the other. It's not what someone wants to do with PDFs I hope :wink: )

kbengs:
> change the other value based on the value a user just changed manually

I think that was a good idea so now it look like this:

old -new 2

kbengs:
With "Fit mode" added:
Screenshot_20240315_094105

bompstable:
This PR is exactly what I need to fix a problem I'm currently experiencing. I have a 3 page document that I want to make into a booklet, but the pages are ever so slightly different in size, so Generate Booklet reports "All pages must have the same size". The pages are 209.9x297.0, 210.1x297.0 and 210.0x296.8 mm respectively - ie ~A4.

With this PR version installed, I have selected all pages, Page size... and then A4 with "Scale and add margins" selected, and all my pages have changed to A4 and can now be made into a booklet.

It did take me a while to work out that I had to use "Scale and add margins" - I intuitively expected that the default Scale option would do what I wanted.

I found it quite a pain selecting A4 from the Paper Size list. It started off with Custom selected and I had to scroll manually upwards through a large list to find A4. This is not a blocker - having the standard sizes to select from is still a huge improvement. It's maybe worth considering if the list is currently too big and if the most common sizes could be prioritized somehow.

kbengs:
Thanks for comments.

I am planning to move "Custom" to the beginning of the list instead of the end. That would make it easier to find A4. The list is big but I am not sure how to select only the common ones, as people probably has different opinions about what is common.

I'll look into this later after summer when I have more time for programming.

bompstable:
The ISO A3, A4 and A5 sizes should be in the top of the list along with Letter, Legal and Tabloid. See https://en.wikipedia.org/wiki/Paper_size for usage information.

It would be helpful if people could type when the list is focussed  to find or jump to what they are looking for.

kbengs:
The list now only have A3, A4, A5, Letter, Legal and Tabloid. It does work as far as I can see.

kbengs:
Ok to merge this?

PR #1084, opened 2024-04-16
Set program name to fix the missing icon when running under Wayland

Integral-Tech:


dreua:
Huh, I'm surprised that didn't cone up earlier. Is everyone like me still on X?

kbengs:
I am also still on X11. But I can confirm that this fixes the icon on wayland. This line actually make the test fail. Don't know why though.

Integral-Tech:
> I am also still on X11. But I can confirm that this fixes the icon on wayland. This line actually make the test fail. Don't know why though.

Wayland is mature and worth trying in 2024 :)

dreua:
> Wayland is mature and worth trying in 2024 :)

Tried it multiple times and went back to X in less than a day each time. Mainly because I have some custom shortcuts to activate/bring to front or run applications on Gnome which rely on wmctrl+xprop which somene (me) would need to port to some alternative in wayland world. And then there used to be always a lot of bugs like this or keepassxc not using dark mode or whatever. But that is all quite off topic here, if you want to help with our migration to wayland we should create a ticket somewhere else :wink: 

It would be nice to get the dogtail tests fixed but on first glance I don't see the solution either.

kbengs:
I tested it only on KDE. @Integral-Tech are you also on KDE?

This should fix the test:

```
--- a/tests/test.py
+++ b/tests/test.py
@@ -185,7 +185,7 @@ class PdfArrangerTest(unittest.TestCase):
         return [
             a
             for a in root.applications()
-            if a.name == "__main__.py" or a.name == "pdfarranger"
+            if a.name in ["__main__.py", "pdfarranger", "com.github.jeromerobert.pdfarranger"]
         ]
```

However on KDE the name on the panel will be:
image

Also in Windows 11, in Task Manager the name will be 'com.github.jeromerobert.pdfarranger' which I think is not right.

Is there a better way to fix the icon?

Integral-Tech:
> I tested it only on KDE. @Integral-Tech are you also on KDE?
> 
> This should fix the test:
> 
> ```
> --- a/tests/test.py
> +++ b/tests/test.py
> @@ -185,7 +185,7 @@ class PdfArrangerTest(unittest.TestCase):
>          return [
>              a
>              for a in root.applications()
> -            if a.name == "__main__.py" or a.name == "pdfarranger"
> +            if a.name in ["__main__.py", "pdfarranger", "com.github.jeromerobert.pdfarranger"]
>          ]
> ```
> 
> However on KDE the name on the panel will be: image
> 
> Also in Windows 11, in Task Manager the name will be 'com.github.jeromerobert.pdfarranger' which I think is not right.
> 
> Is there a better way to fix the icon?

I submitted a new commit to specify the application name. The display name should be correct now :)

PR #1086, opened 2024-04-19
Do not show full path name in headerbar (fixes #1085)

### LINKED ISSUE #1085: Full path+filename in headerbar's title makes it hard to read the filename
Currently, the app puts the entire combined path and filename into the headerbar's title property.

However, paths (especially nested) can be pretty long, and usually are not the main important thing, compared to the filename.

My suggestion would be to split the path from the filename, only show the filename in the headerbar's "title" property, and put the path (maybe stripping out the `/home/theusername/` part of it, too) in HeaderBar's "subtitle" property.

b-sharman:
see #1085 

It looks like this for me now:
image
Please let me know if I've missed anything!

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -533,9 +533,11 @@ class PdfArranger(Gtk.Application):
             filter_list.append(f_img)
         return filter_list
 
-    def set_title(self, title):
+    def set_title(self, title, subtitle = None):
         if Handy:
             self.uiXML.get_object('header_bar').set_title(title)
+            if subtitle is not None:
+                self.uiXML.get_object('header_bar').set_subtitle(subtitle)
         else:
             self.window.set_title(title)
 
@@ -822,8 +824,9 @@ class PdfArranger(Gtk.Application):
         GObject.idle_add(self.retitle)
 
     def retitle(self):
+        subtitle = None
         if self.save_file:
-            title = self.save_file
+            subtitle, title = os.path.split(self.save_file)
         else:
             title = _('untitled')
         if self.is_unsaved:
@@ -835,7 +838,7 @@ class PdfArranger(Gtk.Application):
             title += ' [' + ', '.join(sorted(all_files)) + ']'
 
         title += ' – ' + APPNAME
-        self.set_title(title)
+        self.set_title(title, subtitle)
         return False
 
     def update_thumbnail(self, _obj, ref, thumbnail, zoom, scale, is_preview):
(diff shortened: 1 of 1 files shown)


pdfarranger/pdfarranger.py
```diff
@@ -533,9 +533,11 @@ def __create_filters(file_type_list):
             filter_list.append(f_img)
         return filter_list
 
-    def set_title(self, title):
+    def set_title(self, title, subtitle = None):
         if Handy:
             self.uiXML.get_object('header_bar').set_title(title)
+            if subtitle is not None:
```
dreua, line 539: It might be slightly better to set this always, so that a previously set subtitle can be reset. (I don't know if it should be set to None or something else, someone would need to look that up.)

dreua: I think that is even necessary, what happens if you start empty, load a file and Ctrl+Z back? I think the subtitle might be left behind in this case. Could you check that?

dreua (changes requested):
The `if Handy` makes me wonder in which cases we don't have `Handy`. Are those substantial enough to reconsider this change? (Probably not, but I really don't know.)

pdfarranger/pdfarranger.py
b-sharman: You're right, looks like `None` is an acceptable argument anyway: https://lazka.github.io/pgi-docs/#Handy-1/classes/HeaderBar.html#Handy.HeaderBar.set_subtitle

b-sharman: > I think that is even necessary, what happens if you start empty, load a file and Ctrl+Z back? I think the subtitle might be left behind in this case. Could you check that?

I'm not sure I understand this. For me, the subtitle doesn't appear until a file is saved, as there is no path to show until then. Is that the desired behavior?

b-sharman:
In regards to dependence on `Handy`, I thought it was a little strange to check for it since non-libhandy headerbars have a `set_title` too. I tried setting `Handy=None` and removing the check like this:
```py
    def set_title(self, title, subtitle = None):
        self.uiXML.get_object('header_bar').set_title(title)
        self.uiXML.get_object('header_bar').set_subtitle(subtitle)
```
and it seemed to work fine, although I'm not sure how I would test it more rigorously.

pdfarranger/pdfarranger.py
dreua: You're right, sorry. It is now fixed and future proof anyway, thanks!

kbengs:
I think it would be fine to remove the `if Handy` as @b-sharman suggest. But can be done later also.
Seems to work well.

Issue #1094, opened 2024-06-01
Failure when generating booklet since pikepdf v9 update


sotirismorf:
Describe the bug
----------------

When using the "generate booklet" button under edit, nothing happens.
This is because of this error:

```txt
Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/pdfarranger/pdfarranger.py", line 476, in generate_booklet
    booklet = exporter.generate_booklet(self.pdfqueue, self.tmp_dir, pages)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/pdfarranger/exporter.py", line 452, in generate_booklet
    file.pages.append(newpage)
TypeError: only pikepdf.Page can be appended to PageList
```

pikepdf v9 release notes: https://pikepdf.readthedocs.io/en/latest/releasenotes/version9.html

Quote from the release notes:

> Pdf.pages no longer coerces PDF dictionaries to page objects. You must explicitly insert/add pikepdf.Page objects

To Reproduce
------------

1.  Open pdfarranger
2.  Open any pdf file suitable for pdf creation (here is a test one: https://api.printnode.com/static/test/pdf/multipage.pdf)
3.  Select all pages with ctrl+a
5.  Click on Main Menu > Edit > Generate Booklet

Expected behavior
-----------------

The pages should be converted to a booklet layout.

Input files
-----------

I have checked with multiple pdf files. Even ones that I remember working in the past.

Screenshots
-----------

N/A

System and Versions
-------------------

`pdfarranger --version` output:

```txt
PDF Arranger-1.10.1
pikepdf-9.0.0
libqpdf-11.9.0
```

Distro: Arch Linux

dreua:
That looks very much like current PDF Arranger is not compatible with the breaking change in pikepdf 9, thanks for the very detailed bug report and further research!

dreua:
The fix was quite simple, you could just edit the export.py file in the meantime if you need a quick fix.

PR #1095, opened 2024-06-04
Fix booklet for pikepdf 9

dreua:
Tested with pikepdf 7.0.0, 8.0.0, 9.0.0
Fixes: https://github.com/pdfarranger/pdfarranger/issues/1094

We probably should run the whole testsuite against pikepdf 9.

### DIFF
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -544,7 +544,7 @@ def generate_booklet(pdfqueue, tmp_dir, pages):
                 Contents=pikepdf.Stream(file, content_txt.encode())
             )
 
-        file.pages.append(newpage)
+        file.pages.append(pikepdf.Page(newpage))
     for __ in range(to_remove):
         del file.pages[0]
     file.save(filename)


kbengs:
@dreua  do you understand why the test fail? Other than that it looks fine to me.

dreua:
No, I tried reproducing it locally but the tests fail at a different place for me. Not really familiar with the tests and its failure modes unfortunately.

kbengs:
Looks like this is needed to fix the failing test:
```
diff --git a/pdfarranger/exporter.py b/pdfarranger/exporter.py
index 3f74a45..1603296 100644
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -178,7 +178,7 @@ def _scale(doc, page, factor):
     # when calling as_form_xobject in generate_booklet
     # It's also needed for pikepdf 8.8 but not for pikepdf 6.0.1
     new_page = doc.make_indirect(new_page)
-    return new_page
+    return pikepdf.Page(new_page)
```

PR #1097, opened 2024-06-05
Improvements to dogtail tests

dreua:


dreua:
Theory: On the second TestBatch the App might come up with the initial dialogue because the config has not been saved. Not sure how to best address that. Either we check for the initial dialogue on each run or we need to clear this configuration setting before running batch 1 in order to allow recurring local tests.

dreua:
Another thought: We also need to consider the possibility to run only a single batch (very important if you want to go in a direction of test driven development), that should be kept in mind when deciding on an approach here. (To be researched: Would it be even possible to allow running only a single test routine out of a test batch?)

PR #1102, opened 2024-06-17
Fix booklet for scaled page for pikepdf 9

kbengs:


dreua:
Ah, finally all checks green angain :smile:  Thanks and sorry for the delay.

PR #1108, opened 2024-08-08
Add Arabic translation

albaroty:


albaroty:
sorry, I don't understand

‫في الجمعة، 23 أغسطس 2024 في 2:15 م تمت كتابة ما يلي بواسطة ‪x9a‬‏ <‪
***@***.***‬‏>:‬

dreua:
What mnemonics are: https://help.gnome.org/users/gnome-terminal/stable/pref-keyboard-access.html.en
In translations they are defined by the charachter after the underscore (_). I don't know the best practice regarding mnemonics for arabic, maybe you could ask other translators or check other translations on that. (Or check other not latin script translations of PDF Arranger, e.g. Chinese).

I actually don't know if menmonics are absolutely necessary but I'd rather have them, it can be a speedup if you want to use the keyboard rather then pointing and an accessibility feature as well.

dreua (changes requested):
Mnemonics are missing, as already discussed in comments

albaroty:
Okay, I get it. Thank you. I'll do it.

dreua:
You have the underscore sometimes at the end of a word and sometimes at the beginnging, is this correct? Is Arabic a right to left script and my system is just not displaying it properly? (I'm not familiar with rtl to be honest.)

Also some translations are missing.

It would be nice if someone could test this. If you think it is fine that way I'm open to merging it we'll see if someone complains. Keep in mind though that any fixes for a translation will only be published in the next release so it might take a while until this is released in the first place and then until fixes are released afterwards.

albaroty:
I've tested it on Windows 10 and 11, Fedora Workstation, and it worked flawlessly.

dreua:
Alright that is more platform testing than I expected, thanks a lot!

PR #1141, opened 2024-10-30
Simplify statusbar properties

linkmauve:
Very little of these properties make sense, we can replace the GtkGrid with a GtkBox and remove almost all of them.

### DIFF
--- a/data/pdfarranger.ui
+++ b/data/pdfarranger.ui
@@ -282,49 +282,20 @@ along with PDF Arranger.  If not, see <http://www.gnu.org/licenses/>.
           </packing>
         </child>
         <child>
-          <!-- n-columns=2 n-rows=1 -->
-          <object class="GtkGrid">
+          <object class="GtkBox">
             <property name="visible">True</property>
-            <property name="row_spacing">10</property>
             <child>
               <object class="GtkStatusbar" id="statusbar">
                 <property name="visible">True</property>
-                <property name="margin_start">4</property>
-                <property name="margin_end">4</property>
-                <property name="margin_top">4</property>
-                <property name="margin_bottom">4</property>
-                <property name="hexpand">True</property>
-                <property name="orientation">vertical</property>
-                <property name="spacing">1</property>
+                <property name="hexpand">1</property>
               </object>
-              <packing>
-                <property name="left_attach">0</property>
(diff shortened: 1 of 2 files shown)


kbengs:
The tests does fail...
To run the test on your machine you can:
`sudo docker run -w /src -v $PWD:/src jeromerobert/pdfarranger-docker-ci:1.5.0 sh -c "pip install .[image] ; python3 -X tracemalloc -u -m unittest tests.test"`
The window will not be visible when run in docker.

To run the test with appllication visible you should be able to:
`python3 -X tracemalloc -u -m unittest -v -f tests.test`
I could not however get this to work now.

At least it looks like you have to:
```
--- a/tests/test.py
+++ b/tests/test.py
@@ -229,7 +229,7 @@ class PdfArrangerTest(unittest.TestCase):
         self._app()
         allstatusbar = self._find_by_role("status bar")
         # If we have multiple status bar, consider the last one as the one who display the selection
-        statusbar = allstatusbar[-1]
+        statusbar = allstatusbar[0]
         return statusbar.name
```
But there seems to be a lot more to do. :smile:

kbengs:
The issue with `python3 -X tracemalloc -u -m unittest -v -f tests.test` was that I tried to run it on Wayland. On X11 it worked.
This seems to fix the test:
```
--- a/tests/test.py
+++ b/tests/test.py
@@ -216,7 +216,7 @@ class PdfArrangerTest(unittest.TestCase):
 
     def _is_saving(self):
         allstatusbar = self._find_by_role("status bar")
-        statusbar = allstatusbar[0]
+        statusbar = allstatusbar[-1]
         return statusbar.name.startswith("Saving")
 
     def _wait_saving(self):
@@ -229,7 +229,7 @@ class PdfArrangerTest(unittest.TestCase):
         self._app()
         allstatusbar = self._find_by_role("status bar")
         # If we have multiple status bar, consider the last one as the one who display the selection
-        statusbar = allstatusbar[-1]
+        statusbar = allstatusbar[0]
         return statusbar.name
```

linkmauve:
Thanks for the patch!  I’m not sure I understand why the order of statusbars changed though, it shouldn’t have according to the .ui file at least.

kbengs (approved):

PR #1150, opened 2024-11-20
Compatibility switch for saving document-level information

### LINKED ISSUE #1144: Inconsistent handling of password-protected PDFs 1.11.x vs 1.10.0
Describe the bug
----------------

Up until version 1.10.0 we used PDF Arranger to decrypt encrypted PDF documents. 
Open encrypted file with password -> Export file / Save as = Exported PDF is not encrypted 

Since version 1.11.0, PDF Arranger saves PDF files with the password of the first imported/opened file if the first imported file was encrypted. I did not see this mentioned in the release notes. This “magical” behavior may be useful when working with only one document, but it can be confusing when working with different documents, wanting to save unencrypted documents or wanting to change the password / PDF restrictions of a PDF document, as there are no settings to configure the export settings.

To Reproduce
------------
Please always restart PDF Arranger 1.11.1 prior to testing any of the following scenarios: 
-> If you import an **unencrypted file first** the exported documents are not encrypted even if you import encrypted documents later.
-> If you import two encrypted documents, the **password of the document imported first is used for the export,** even if you change the order of the pages so that the pages of the document imported later appear first.
-> If you import a unencrypted PDF document first and encrypted documents later the export is always not encrypted
-> If you add a blank page to an empty PDF Arranger window, delete it and then import an encrypted PDF the export has no encryption 
-> Open document that is not encrypted and delete all pages - import an encrypted document in the now empty PDF Arranger window. Exported document is encrypted and has the password of the first document you opened  

Expected behavior
-----------------

Same as 1.10.x: Saved PDFs are not password protected.

Future feature: Add password settings (and maybe file restrictions) to document settings / options menu (or at least an option "Keep encryption / document restrictions of first imported document) 

System and Versions
-------------------

-   PDF Arranger Portable Windows versions 1.10.0 + 1.11.0 + 1.11.1 (Github releases)
-   Windows 10

StefRe:
Let user decide whether document-level information, such as outlines, tags, encryption passwords, etc., shall be taken from the primary input file and be preserved in the output file. This became the default behavior with PDF Arranger 1.11 using pikepdf version 8 or newer, which uses QPDF jobs. Prior to that, the output was always built starting with a new, empty pdf document and thus all the document-level information from the input file was discarded in the output. This old behavior can now be enforced by deactivating the check button
"Keep document-level information from first page's file" in the save file chooser dialog. If this option is deactivated, the "--empty" qpdf option is used inplace of the primary input file. As a side effect, this for instance removes passwords when saving password-protected input files.

grafik

Further, if the check button is kept activated (default), the pdf file of the first page to be saved is used as the primary input file. Previously the first opened file was used, which lead to unexpected/surprising behavior if all pages of this file are removed before saving or if other files' pages were moved to the beginning of the resulting document.

Closes #1144

### DIFF
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -432,7 +432,7 @@ def _add_json_entries(json: Dict[str, Any], files: List[List[str]], page: Page) 
 
 
 def _create_job(files: List[List[str]], pages: List[Page], files_out: List[str], quit_flag=None,
-                test_mode: bool = False):
+                test_mode: bool = False, start_with_empty: bool = False):
     """ Same as _copy_n_transform, except it use the pikepdf Job interface. Requires pikepdf >= 8.0 """
     # Generate the output PDF file including temporary overlay/ underlay pages. We don't need to call
     # _append_page as the Job interface copies pages / annotations as necessary. We can also delay getting
@@ -440,12 +440,16 @@ def _create_job(files: List[List[str]], pages: List[Page], files_out: List[str],
     json = dict(outputFile=files_out[0], pages=[], removeUnreferencedResources="yes")
     if test_mode:
         json.update(qdf="", staticId="", compressStreams="n", decodeLevel="all")
-    if len(files) > 0 and len(files[0][0]) > 0:
-        json["inputFile"] = files[0][0]  # We are treating files [0] as the main document
-        if len(files[0][1]) > 0:
-            json["password"] = files[0][1]
+    if start_with_empty:
+        json["empty"] = ""
     else:
-        json["inputFile"] = "."
+        primary_file = files[pages[0].nfile - 1]  # treat file of first page as primary input file
+        if len(primary_file[0]) > 0:
+            json["inputFile"] = primary_file[0]
+            if len(primary_file[1]) > 0:
+                json["password"] = primary_file[1]
+        else:
+            json["inputFile"] = "."
 
     for page in pages:
         if quit_flag is not None and quit_flag.is_set():
@@ -458,9 +462,9 @@ def _create_job(files: List[List[str]], pages: List[Page], files_out: List[str],
 
 
 def export_doc_job(pdf_input: List[pikepdf.Pdf], files: List[List[str]], pages: List[Page], mdata, files_out: List[str],
-                   quit_flag, test_mode: bool = False) -> None:
+                   quit_flag, test_mode: bool = False, start_with_empty: bool = False) -> None:
     """  Same as export() but uses the pikepdf Job interface. Requires pikedf >= 8.0. """
-    job = _create_job(files, pages, files_out, quit_flag, test_mode)
+    job = _create_job(files, pages, files_out, quit_flag, test_mode, start_with_empty)
     pdf_output = job.create_pdf()
     max_version = max(
         pdf_output.pdf_version,
@@ -489,14 +493,14 @@ def export_doc_job(pdf_input: List[pikepdf.Pdf], files: List[List[str]], pages: 
         job.write_pdf(pdf_output)
 
 
-def export(files, pages, mdata, files_out, quit_flag, _export_msg, test_mode=False):
+def export(files, pages, mdata, files_out, quit_flag, _export_msg, test_mode=False, start_with_empty=False):
     pdf_input = [
         pikepdf.open(copyname, password=password) for copyname, password in files
     ]
     if version.parse(pikepdf.__version__) < version.Version("8.0"):
         export_doc(pdf_input, pages, mdata, files_out, quit_flag, test_mode)
     else:
-        export_doc_job(pdf_input, files, pages, mdata, files_out, quit_flag, test_mode)
+        export_doc_job(pdf_input, files, pages, mdata, files_out, quit_flag, test_mode, start_with_empty)
 
 
 def num_pages(filepath):
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -15,6 +15,7 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 import os
+import packaging.version as version
 import platform
 import ctypes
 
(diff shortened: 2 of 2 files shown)


StefRe:
The test matrix below uses the following abbreviations:
symbol | meaning
----|----
u	 |unencrypted
e1	|encrypted with password 1
e2	|encrypted with password 2
x y	|open x, then import y
(x)	|open/import and remove all pages of x
x ↔ y	|import x, then y, then swap places of pages
x + y	|concatenation of pages of x and y 

Results of manual testing ($\color{red}{\text{red}}$   = wrong/unexpected result):

\#	|	Input	|	Output content this PR	|	Output password this PR	|	Output password version 1.11.1
:---:	|	:---:	|	:---:	|	:---:	|	:---:
1	|	e1	|	e1	|	1	|	1
2	|	u e1	|	u + e1	|	none	|	none
3	|	e1 u	|	e1 + u	|	1	|	1
4	|	(u) e1	|	e1	|	1	|	$\color{red}{\text{none}}$  
5	|	(e1) u	|	u	|	none	|	$\color{red}{\text{1}}$  
6	|	u ↔ e1	|	e1 + u	|	1	|	$\color{red}{\text{none}}$  
7	|	e1 ↔ u	|	u + e1	|	none	|	$\color{red}{\text{1}}$  
8	|	e1 ↔ e2	|	e2 + e1	|	2	|	$\color{red}{\text{1}}$   

No output passwords are set if the new check button is deactivated. This corresponds to the old behavior (or if pikepdf version < 8 is used).

Not sure if we need dogtail tests for all test cases - maybe it's enough to just test case \#  4 with and without activated check button (=2 tests). If you're OK with this PR in general I'll try to write the tests.

dreua:
Is it safe to assume this checkbox in the file dialog will also work on Windows and Mac? Could anyone give that a try?

kbengs:
The checkbox will be visible on Windows but it will change from using Windows native dialog to Gtk. I do like Windows own save dialog in Windows.

I also tested flatpak in KDE by replacing pdfarranger.py and exporter.py in the flatpak install. The checkbox is not visible there. I don't know it it might be when the flatpak is built as it should be?

When user has installed Pikepdf < 8, shouldn't the checkbox be greyed out and unchecked?

Should we add the checkbox in preferences dialog instead (or also) ?

StefRe:
> When user has installed Pikepdf < 8, shouldn't the checkbox be greyed out and unchecked?

Nice find - yes, I guess it should definitely be greyed out and unchecked.

Also, on second thought, maybe the checkbox should also be available for export? Right now it's only available for saving, not for exporting. Maybe making it availble (sensitive) and pre-setting it to unchecked is a better choice in this case?

kbengs:
> Tucking away the checkbox into the prefences dialog

I do agree with you. But on Windows I would like to have Windows own dialog. And if it is not visible in KDE+flatpak (which use portals?) then we would need to have the checkbox also in preferences dialog.

> Previously the first opened file was used, which lead to unexpected/surprising behavior

For preserving of outlines and links we have the popup at start which links to this https://github.com/pdfarranger/pdfarranger/wiki/User-Manual#preserving-of-outlines-and-links

I'd maybe argue that m-holgers concept "The first (opened) file is treated as the file being edited, and pages are imported into this file from other files. " is more intuitive (than "file from first page in iconview being treated as the file being edited"). But I don't have a clear opinion.

Do you have a link to documentation for for example ` json["empty"] = ""` ?

StefRe:
> ... then we would need to have the checkbox also in preferences dialog.

I agree and will add the checkbox to the preferences dialog too and pre-select the checkbox in the file chooser dialog accordingly.

> I'd maybe argue that m-holgers concept "The first (opened) file is treated as the file being edited, and pages are imported into this file from other files. " is more intuitive.

At first glance yes. The main motivation for me changing it to first page's file was that I (obviously incorrectly) use PDF Arranger when treating a couple of files in a row: I do the first file (append, reshuffle, etc.) then save it, when I'm done with this file I just remove all pages and drag the next file to be treated into PDF Arranger and when I then save this one it takes outlines etc. from the previous "job". This is described in test cases 4 and 5 above. I know that I should open a new instance and close the old one but my workflow is more intuitive to me. But if you prefer the existing approach (take outlines etc from first file opened) it's also fine with me. Maybe in this case it would make sense to add another menu item "Reset workspace" to clear the internal files list.

> Do you have a link to documentation for for example  json["empty"] = ""?

documentation of the _empty_ option: https://qpdf.readthedocs.io/en/stable/cli.html#option-empty and the link there to _Page Selection_

documentation on how to specify flags with no parameter: https://qpdf.readthedocs.io/en/stable/qpdf-job.html  (the third last item on this page):
> Bare command-line options (flags with no parameter) map to config functions that take no options and to **JSON keys whose values are required to be the empty string**.

dreua:
I'm really not convinced this is worth the hassle. The user experience doesn't seem good to me, there is a high chance that many people will not understand this feature and will be confused by it. If we lose the native dialog on windows and the possibility to properly use file portals one day (and it seems to cause trouble already today) then I vote against it.

If the only known use case is to remove passwords then we should just have a menu item that does that imo. Much better UX, people who want to remove passwords will actually find it (it is unlikely that the feature as proposed in this PR will be recognized as the button to click by the majority of users).

And of course it needs to be the same for all ways of exporting / saving.

StefRe:
> And of course it needs to be the same for all ways of exporting / saving.

I don't quite think so:  First it turns out that the job interface is only used when saving/exporting to a single file. When exporting to multiple files, the old approach (starting each output file from an empty pdf) is used even in `export_doc_job` (i.e. its docstring is not exactly correct). So the new checkbox must be grey out (set insensitive) in these cases.  

Second, I think that when selecting only certain pages for exporting into one file the natural user expactation would be to not have the complete outline added to this document. This would only make sense if you select all and export to a single file, but this is the same as "Save to" a single file, i.e. not the normal use case for exporting.

So in summary I think the following pre-settings should be chosen for the checkbox: 

ALL_TO_SINGLE: sensitive and activated
SELECTED_TO_SINGLE: sensitive and deactivated
ALL_TO_MULTIPLE, SELECTED_TO_MULTIPLE: insensitive (and deactivated)

kbengs:
> .. Maybe in this case it would make sense to add another menu item "Reset workspace" to clear the internal files list...

There is already close, ctrl +w which clears everything.

>..But if you prefer the existing approach (take outlines etc from first file opened) it's also fine with me..

I think this is better.

> ..documentation of the empty option:..

Okay. The code could have a comment above the  `json["empty"] = ""` line explaining what trick it does, as it is not very obvious. (add empty pdf first which makes qpdf drop any document level information, as this info is taken from the first document)

I think dreua has a point. It might be hard to understand that the checkbox does. I am fine with either your approach or the approach dreua suggest. @dreua may decide :-)

StefRe:
> There is already close, ctrl +w which clears everything.

Thank you so much for pointing this out - I just overlooked the obvious. My issue is resolved and there's no need anymore for taking outlines etc. from the file of the first page to be saved instead of from the first file opened.  

So in view of the difficulties of the proposed checkbox in terms of availability on all platforms and that it might be difficult to understand by some users what exactly it will effect I suggest closing this PR.

PR #1151, opened 2024-11-23
Update README.md: macOS guide

Endle:


dreua:
This is either use the nix installer which does erverything or do the install yourself from source, right? If so, that should be made clearer, maybe make this two subsections with a clear "either or" introduction. (Assume that people reading this have no clue what they are doing and can't piece that together from the context you and me have.)

Furthermore I have no idea about nix and the alternative installer you linked and if we want to endorse those in our Readme. Maybe we could write that more like a users choice than a recommendation, something like "If you want to use nixos, here is the command and the package. For Flatpaks we just link to without further ado or explanation on how to install it.

Is the alternative installer important enough to warrant a link in our readme?

Endle:
>  If so, that should be made clearer, maybe make this two subsections with a clear "either or" introduction.

I totally agree with you. I'll add a bit more context to it. I appreciated your feedback that "we need more context" or "we need to keep it brief". I'll update my PR.


> Furthermore I have no idea about nix and the alternative installer you linked and if we want to endorse those in our Readme. 

That's my concern as well. IMHO the determined installer (alternative) provide a simple way to install nix on mac (just one command), and the dev team had a good reputation in nix community (https://www.reddit.com/r/NixOS/comments/11dfiqq/comment/jaot5uc/). However, I don't know if it's a good idea that pdfarranger endorses this installer.

kbengs:
How about if we add the mac stuff in Mac.md and link from README.md just under the Windows link:
"For Mac see [Mac.md](Mac.md)."

Endle:
> How about if we add the mac stuff in Mac.md and link from README.md just under the Windows link:
> "For Mac see [Mac.md](Mac.md)."

@kbengs That's a good advice. I've updated my PR. Sorry for my late reply.

kbengs:
Thanks!

PR #1154, opened 2024-11-30
Add more explicit deps and setup instructions

selfhoster1312:
These are the versions i tested with. I have no idea if you want more relaxed requirements, but i was confused that i can not run pdfarranger because missing dependencies so i thought it's a good idea to declare them.

Also i added instructions to run uv because it's so much faster than setup tools starting the venv from scratch takes <1s.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,12 +72,28 @@ In addition, *PDF Arranger* supports image file import if [img2pdf](https://gitl
 
 ## For developers
 
+### Using setuptools
+
 ```
 git clone https://github.com/pdfarranger/pdfarranger.git
 cd pdfarranger
 ./setup.py build
 python3 -m pdfarranger
 ```
+
+### Using uv
+
+[uv](https://docs.astral.sh/uv/) is a modern alternative to setuptools. You can setup a development environment like this:
+
+```
+git clone https://github.com/pdfarranger/pdfarranger.git
+cd pdfarranger
+uv venv
+uv pip sync setup.py
+uv run python3 -m pdfarranger
+```
+
+### On Windows
 
 For Windows see [Win32.md](Win32.md).
 
--- a/setup.py
+++ b/setup.py
@@ -121,7 +121,7 @@ setup(
     entry_points={
         'console_scripts': ['pdfarranger=pdfarranger.pdfarranger:main']
     },
-    install_requires=['pikepdf>=6','python-dateutil>=2.4.0', 'packaging'],
+    install_requires=['pikepdf>=6','python-dateutil>=2.4.0', 'packaging', 'pillow>=11', 'lxml>=5.3', 'pygobject>=3.50', 'six>=1.16', 'pycairo>=1.27'],
     extras_require={
         'image': ['img2pdf>=0.3.4'],
     },


kbengs:
The versions available on Debian are:
pillow
debian 10 5.4.1
debian 11 8.1.2
debian 12 9.4.0

lxml
debian 10 4.3.2
debian 11 4.6.3
debian 12 4.9.2

pygobject
debian 11 3.38.0
debian 12 3.42.2

python3-six
debian 10 1.12.0
debian 11 1.16.0
debian 12 1.16.0

pycairo
debian 11 1.16.2
debian 12 1.20.1

(from repology.org)
I think we need to support at least Debian 11.  What does @dreua think?

dreua:
Some thoughts in no particular order:

- Are we using six anywhere? If so, why?
- Our readme is not free advertising space, I'd say we only reference to new stuff there if there is good reason and we have a maintainer or at least regular contributor who can vouch for it and can be assumed to respond to any questions regarding its usage. I'm not using uv.
- CI is failing, I just approved running the jobs. (I think you can do that too, or are there any access rights missing for you, kbengs?)
- Yes, I agree, at least Debian 11 should still be supported.
- Dependencies are documented in the readme and for some should be installed as system packages, I am not convinced declaring those in setup.py is even a good idea to start with. (I can't remember anyone having problems or asking about installation before.)
- While this is probably a well intentioned PR (thank you for taking the time!) I'm leaning towards not merging anything at all. (Sorry.)

StefRe:
also, `Pillow` and `lxml` are dependencies of `pikepdf` and hence will be installed anyway.

selfhoster1312:
Some good points @dreua. I don't take it personally, but let me share my own experience. The only reason i did this is because following the "Debian-based distributions" instructions failed for me on PopOS. After installing the system dependencies, i get:

```
$ ./setup.py build
running build
running build_py
creating build
creating build/lib
creating build/lib/pdfarranger
copying pdfarranger/core.py -> build/lib/pdfarranger
copying pdfarranger/__main__.py -> build/lib/pdfarranger
copying pdfarranger/splitter.py -> build/lib/pdfarranger
copying pdfarranger/__init__.py -> build/lib/pdfarranger
copying pdfarranger/undo.py -> build/lib/pdfarranger
copying pdfarranger/config.py -> build/lib/pdfarranger
copying pdfarranger/metadata.py -> build/lib/pdfarranger
copying pdfarranger/iconview.py -> build/lib/pdfarranger
copying pdfarranger/exporter.py -> build/lib/pdfarranger
copying pdfarranger/pageutils.py -> build/lib/pdfarranger
copying pdfarranger/pdfarranger.py -> build/lib/pdfarranger
running build_i18n
running build_icons
user@pop-os:~/pdfarranger$ python3 -m pdfarranger
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/selfhoster1312/pdfarranger/pdfarranger/__main__.py", line 2, in <module>
    from pdfarranger import pdfarranger
  File "/home/selfhoster1312/pdfarranger/pdfarranger/pdfarranger.py", line 38, in <module>
    import pikepdf
ModuleNotFoundError: No module named 'pikepdf'
```

I am aware that setuptools is mostly considered deprecated by much of the python ecosystem, that `pyproject.toml` is  the shiny "new" ([8 years old](https://peps.python.org/pep-0518/)) standard way to do things, and that dependency management and virtualenvs are a mess. 

I heard good things about uv but it's my first time trying it. It's not paid advertising and it's not even a favorite tool of mine. I know i hope never to have to touch Python code ever again :smile:. In the meantime, this was the easiest solution i found to run pdfarranger from source. An alternative could be nix (or guix) but i'm not good even at that to write a good shell.nix/flake.nix.

Overall, i don't care for uv. I just care to find setup instructions i can follow to get started, and the current instructions failed me. I'm pretty confident the current setup instructions will not work on a fresh Debian.

That's just my two cents. I'll leave the solution up to you people :+1:

dreua:
Most of the time it is better to start with an issue / discussion. That is much easier to follow along for the other side (us) than coming from the other way around. There is spam on the internet and also on Github so one never knows the intentions, especially from a user you haven't interacted before, hence my skepticism when new tools are PRed towards the readme.

I understand now where you are coming from and agree that the "for developers" instructions could be better. Note that the setup.py call does not install any dependencies which is expected

If you just want to install from source, there is a section with this title above it. It ends with a call to `pip3` which should install the declared dependencies including pikepdf, pillow, lxml and six (needed for dateutils, not required by us directly).

Now pycairo and pygobject from your PR are still missing but I believe those are better installed as system package on most distributions. (Open to discussions and other opinions, I haven't checked this in depth for both packages right now.)

If you want to use a venv (or alternative) that should work too, but you'll need to tell the venv to also allow usage of the system side packages. I think we already had an issue about that here. We need the right `gi` import which is hard or impossible to do in a clean venv iirc.

PR #1156, opened 2024-12-01
Add switch to save into empty file

### LINKED ISSUE #1144: Inconsistent handling of password-protected PDFs 1.11.x vs 1.10.0

StefRe:
Let user decide whether document-level information, such as outlines, tags, encryption passwords, etc., shall be taken from the first opened file and be preserved in the output file when saving or exporting to a single file. This became the default behavior with PDF Arranger 1.11 using pikepdf version 8 or newer, which uses QPDF jobs. Prior to that, the output was always built starting with a new, empty pdf document and thus all document-level information from the input file was discarded in the output. This old behavior can now be enforced by deactivating the check button "Retain document-level information of the first opened file" in the preferences dialog.
This new check button is only available when working with pikepdf 8 or newer.

Closes #1144

### DIFF
--- a/pdfarranger/config.py
+++ b/pdfarranger/config.py
@@ -19,6 +19,8 @@ import configparser
 import os
 import sys
 import gettext
+import packaging.version as version
+import pikepdf
 from gi.repository import Gdk
 from gi.repository import Gtk
 
@@ -129,6 +131,7 @@ class Config(object):
         self.popup_menu_accels = [
             Gtk.accelerator_parse(x) for x in a["context-menu"].split()
         ]
+        self.has_pikepdf8 = version.parse(pikepdf.__version__) >= version.Version("8.0")
 
     def is_popup_key_event(self, keyevent):
         for key, mods in self.popup_menu_accels:
@@ -181,6 +184,15 @@ class Config(object):
 
     def set_theme(self, theme):
         self.data.set('preferences', 'theme', theme)
+
+    def start_with_empty(self):
+        return (self.data.getboolean('preferences', 'start-with-empty', fallback=False)
+                if self.has_pikepdf8 else True)
+
+    def set_start_with_empty(self, start_with_empty):
+        if not self.has_pikepdf8:
+            start_with_empty = True
+        self.data.set('preferences', 'start-with-empty', str(start_with_empty))
 
     def scale_mode(self):
         return self.data.get('print-settings', 'scale-mode', fallback="PRINTABLE")
@@ -261,11 +273,22 @@ class Config(object):
         psettings = PrintSettingsWidget(self.scale_mode(), self.auto_rotate())
         frame3.add(psettings)
         d.vbox.pack_start(frame3, False, False, 8)
+        if self.has_pikepdf8:
+            frame4 = Gtk.Frame(label=_("Saving/exporting to single file"), margin=8)
+            cb_retain = Gtk.CheckButton(
+                label=_("Retain document-level information of the first opened file"),
+                margin=8)
+            cb_retain.set_active(not self.start_with_empty())
+            frame4.add(cb_retain)
+            d.vbox.pack_start(frame4, False, False, 8)
+        else:
+            # prevent CodeQL false positive "uninitialized local variable"
+            cb_retain = None
         t = _("For more options see:")
-        frame4 = Gtk.Frame(label=t, shadow_type=Gtk.ShadowType.NONE, margin=8)
-        label4 = Gtk.Label(self._config_file(self.domain), selectable=True, margin=8)
-        frame4.add(label4)
-        d.vbox.pack_start(frame4, False, False, 8)
+        frame5 = Gtk.Frame(label=t, shadow_type=Gtk.ShadowType.NONE, margin=8)
+        label5 = Gtk.Label(self._config_file(self.domain), selectable=True, margin=8)
+        frame5.add(label5)
+        d.vbox.pack_start(frame5, False, False, 8)
 
         langs = []
         if os.path.isdir(localedir):
@@ -299,6 +322,8 @@ class Config(object):
             num2 = combo2.get_active()
             theme = themes[num2] if num2 != 0 else ""
             self.set_theme(theme)
+            if self.has_pikepdf8:
+                self.set_start_with_empty(not cb_retain.get_active())
             self.set_scale_mode(psettings.get_scale_mode())
             self.set_auto_rotate(psettings.get_auto_rotate())
         d.destroy()
--- a/tests/test.py
+++ b/tests/test.py
@@ -377,6 +377,12 @@ class PdfArrangerTest(unittest.TestCase):
             time.sleep(0.1)
         registry.generateKeyboardEvent(code, None, KEY_RELEASE)
 
+    def _config(self):
+        """Get config from config file used in test"""
+        config = configparser.ConfigParser()
+        config.read(self.__class__.config_file)
+        return config
+
     def _quit(self):
         self._app().child(roleName="layered pane").keyCombo("<ctrl>q")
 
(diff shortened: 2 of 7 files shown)


tests/test.py
```diff
@@ -898,3 +895,60 @@
...
+            self.assertNotIn(warning, stdout)
+        else:
+            self.assertIn(warning, stdout)
+
+    def test_02_legacy(self):
+        """Test legacy setting"""
+        if have_pikepdf8():
+            self._mainmenu("Preferences")
+            dialog = self._app().child(roleName="dialog")
+            panels = self._find_by_role("panel", dialog)
```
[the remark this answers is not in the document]
StefRe: fixed

pdfarranger/config.py
```diff
@@ -299,6 +319,8 @@
             num2 = combo2.get_active()
             theme = themes[num2] if num2 != 0 else ""
             self.set_theme(theme)
+            if self.has_pikepdf8:
+                self.set_start_with_empty(not cb_retain.get_active())
```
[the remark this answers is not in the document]
StefRe: is false positive, fixed by checking if variable exists

pdfarranger/config.py
```diff
@@ -299,6 +319,10 @@
             num2 = combo2.get_active()
             theme = themes[num2] if num2 != 0 else ""
             self.set_theme(theme)
+            if self.has_pikepdf8:
+                # prevent CodeQL false positive "uninitialized local variable"
+                if cb_retain in vars():
```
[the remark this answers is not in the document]
StefRe: fixed by defining it (anyways should have been `'cb_retain' in vars()` instead of `cb_retain in vars()`)

kbengs:
`start_with_empty` does not sound perfect to me. What about preserve.. something or retain.. something ? Other than that I think it looks good. Are you able to fix the tests?

StefRe:
> `start_with_empty` does not sound perfect to me

I know (especially because it's the negation of the label text) but it is the more technical internal description of what is going on (we start with an empty pdf and add all the pages to it). The user is not supposed to directly edit the config file, so I thought the name might be OK nonetheless.
As you have seen, I didn't use the job interface (option `--empty`) for pikepdf >= 8 as it turned out that it yields slightly different results for files with annotations. It didn't, however, investigate why exactly.

> Are you able to fix the tests?

I thinks so. The problem is that I can't run the dogtail tests locally as I'm on Arch Linux with at-spi2 version 2.54.0 and they renamed the push button role in 2.53.1 which breaks dogtail. I opened an [issue](https://gitlab.com/dogtail/dogtail/-/issues/35) and the dogtail people already came up with a fix which I'll test in the next couple of days so that I can fix the tests locally.

StefRe:
The Snap build error seems to me to be unrelated to the PR. Could you have a look, please. I have no knowledge of snap.

kbengs:
I think I have seen the snap fail sometimes but then work the next time again.

There are issues if you try to save a pdf which produces warnings, for example this:
https://drivecentre.ca/wp-content/uploads/2017/03/Vacon-NXP-NXC-User-Manual-DPD00890B-UK.pdf

kbengs (changes requested):
If a PDF produces warnings at saving you will see exceptions in terminal and no warning window.

StefRe:
> If a PDF produces warnings at saving you will see exceptions in terminal and no warning window.

Hm, I'm not sure what this is supposed to refer to? This PR doesn't change anything about error handling.

kbengs:
wrapper() (line 245 exporter.py) is expecting export_msg (=warnings and errors at saving) as last argument. You have changed that to self.config. Becuase of that the export process cant pass messages to main process. You should try to open and save the file I linked a few post above. Try it before your commit and then after. If you have clicked to not show warnings on saving you need to edit config.ini and set "show-save-warnings = True"

StefRe:
@kbengs I fixed the `export_msg` issue and tested it manually, but I'm not sure if I'll have the time to write a test for `show-save-warnings` within the next few days. If you like you can merge it without the test and I'll make a separate PR for the test. If not I'll mark this PR as draft and see if I can finish it next week.

kbengs (approved):

kbengs:
I think it looks good. Do you agree @dreua ?

StefRe:
Converting to draft until #1166 goes in.

tests/test.py
```diff
@@ -924,3 +931,89 @@ def test_01_cli_version(self):
         self.assertIn("libqpdf", stdout)
         self.assertIn("OS-", stdout)
         self.assertIn("OS_Version-", stdout)
+
+
+class TestBatch10(PdfArrangerTest):
+    """Test show-save-warnings config settings"""
```
kbengs, line 937: You are not testing show-save-warnings here, are you?

tests/test.py
```diff
@@ -924,3 +931,89 @@ def test_01_cli_version(self):
...
+            self.assertTrue(checkboxes[0].checked)
+        else:
+            panels = self._find_by_role("panel", dialog)
+            self.assertNotIn(savepanel_name, [p.name for p in panels])
+        dialog.child(name="OK").click()
+        self._wait_cond(lambda: dialog.dead)
+
+        self._mainmenu("Save As…")
+        self._save_as_chooser("encrypted.pdf", ["encrypted.pdf"] )
+        with pikepdf.open(os.path.join(self.__class__.tmp, "encrypted.pdf"), password="foobar"):
```
kbengs, line 963: Why is this "with pikepdf.." here, is it a mistake?

tests/test.py
```diff
@@ -924,3 +931,89 @@ def test_01_cli_version(self):
...
+        with pikepdf.open(os.path.join(self.__class__.tmp, "unencrypted.pdf")):
+            # there should be no pikepdf._core.PasswordError
+            pass
+
+    def test_03_quit(self):
+        self._quit()
+
+
+class TestBatch11(PdfArrangerTest):
+    """Test metadata"""
```
kbengs, line 999: Here you are testing show-save-warnings ?

tests/test.py
StefRe: Yes, thanks. This used to be `"""Test saving config settings"""` but I thought this is ambiguous (test the config settings for saving vs. test that config settings are correctly saved), so I decided to re-word it but got myself confused.  This should read 
```
class TestBatch10(PdfArrangerTest):
    """Test start-with-empty config settings""" 
```
which is easier to relate to the actual code. Will fix it.

tests/test.py
StefRe: Yes, thanks for spotting this mistake. This should read
```
class TestBatch11(PdfArrangerTest):
    """Test show-save-warnings config setting"""
```
This TestBatch11 will later also be used in # 1169 for metadata testing, but I'll amend the docstring then. 
Will fix it.

tests/test.py
StefRe: Yes - another good find by you :). Lines 963 and 964 should be deleted. Will fix it.

StefRe:
@kbengs Thank you again very much for the careful reading of this PR. Next time I'll try to cause less confusion by doing one thing after the other and not cherry pick from my local development branch with different issues fixed at the same time that depend on each other.

kbengs:
Thanks!

Issue #1067, opened 2024-03-02
Split and reorder booklet to pages


souslaplage:
There is the option to put several pages in booklet format. Can there be reverse options (re-format page by page)?
Integrate different bookletimposer options?

thank you to you

dreua:
Would that be a combination of split (we have that) and reorder (would currently be manual work I think)?

dreua:
In general that would be possible but it is the question if someone invests the time to implement and maintain it. On my first thought this seems like a niche use case and therefore unlikely to be implemented, but I'm in no way authority on that.

souslaplage:
Yes, that's the right combination: the 1st page cut becomes the last, the 3rd the penultimate and so on. It's true that you can do it manually, but on large documents it can be time-consuming.

In terms of needs, I don't know how many people use it. I was thinking that as you already have the great page-by-page option to booklet, having the reverse would be very beneficial for reworking with a single program in one direction or another. I was also talking about booklet impose, which I don't think is developed any more, but which a lot of people who do paper distribution still use. It's present by default on the tails system, for example, so I imagine that it's used a lot, and reinstating its options will ensure that it's also functional in the long term. I've already noticed that some documents that can't be put in booklet format on bookletimposer work on pdfarranger. 

It's up to you to see what you think.

m93a:
Hey! I've been semi-active in the DIY zine community and we've been using PDF Arranger to create booklets to print our zines! I just wanted to let you know that having an inverse operation would be great help for rescuing old zines (eg. those which can only be downloaded in booklet form).

selfhoster1312:
Thanks to @dreua and @kbengs reviewing #1157 there should be unimposition available in the next PDFArranger release in the "Booklet->Split" menu action.

If someone would like to test before release, let me know and we can give instructions on how to try it out!

Let the DIY zines flourish :fire:

kbengs:
Implemented in #1157
@selfhoster1312 Thanks!

PR #1157, opened 2024-12-01
Allow unimposition from the "Split pages" menu

selfhoster1312:
This is a much requested feature by my friends who make zines, and was also requested in #1067. This is the only feature from booklet-imposer missing in pdfarranger (booklet-imposer is deprecated).

This PR adds an unimposition checkbox in the split pages menu.

TODO:

- [x] work on page selection, not just entire document
- [x] make sure page selection is contiguous
- [x] update translations
- [ ] tests pass

I wanted to have feedback to help finalize this PR so i'm posting early. Here's a preview:

unimpose

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -704,7 +704,7 @@ class TestBatch5(PdfArrangerTest):
 
     def test_03_booklet(self):
         self._popupmenu(0, ["Select", "Select All"])
-        self._popupmenu(0, ["Generate Booklet"])
+        self._popupmenu(0, ["Booklet", "Generate (imposition)"])
         self._wait_cond(lambda: len(self._icons()) == 2)
         self._app().child(roleName="layered pane").keyCombo("Home")
         self._assert_page_size(491.1, 213.3)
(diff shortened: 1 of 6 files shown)


kbengs:
Imo it would fit better in a "Booklet" menu with two submenus "Impose booklet" and "De-impose booklet"
ie "generate booklet" is renamed to "Impose booklet".

> make sure page selection is contiguous

It would be possible to grey out "De-impose booklet" when selection is not continuous.

When doing model changes you should use `with self.render_lock():` I'll look more at the code later.

dreua:
With only a first glance at the code I'd say the "split_pages" function is getting way to long. Would it be feasible to split (haha) into at least two functions?

kbengs:
Unimpose is fine for me. I am unfamiliar with all the suggested words. :smiley: I suggest the names should make it clear that one is the opposite of the other.

selfhoster1312:
Does this look better? Gathering new feedback before trying to write tests and updating translations...

There's now a Booklet menu containing:

- Generate (imposition)
- Split (unimposition)

I've moved the `_crops` splitter method out of the `Dialog` class so i can reuse it without magic tricks, maybe we could add some tests here too... I've added an actual error dialog when the selection is not contiguous.

Maybe we should make sure page size is the same in selection?

selfhoster1312:
So i can't run tests myself but otherwise i think this is ready for review. Some notes:

- the translation/testing process is very badly explained in docs (but i will open dedicated issue for this)
- it's strange that imposition adds blank pages even for 2 pages (which could be a simple A4 pane that you'd like to duplicate to print 2 times the same A5 leaflet for each A4 sheet), and maybe stranger that it adds those white pages at the end? i would think adding blank pages on page 1/n-1 would make most sense so you keep the cover intact
- i'm really not sure about the end of the split_booklet method, i just copied that from split_pages, and it looks very different from generate_booklet method end... i don't know what i'm doing, i don't know why i would use the exporter, or not... please pay attention to that part :)

Otherwise i think it's looking good?

kbengs:
Some comments:
* Generate and Split sound good to me
* Mnemonics is missing ie _Generate and _Split in menu.ui (to allow use of Alt + G or S)
*  Split need to be greyed out when nothing is selected, otherwise there are errors.
* the end of split_booklet looks good. The comment can be removed. (end of generate_booklet has clear_selected() which does the same lines)

I'll look more later.

selfhoster1312:
I rebased and addressed your comments. I think you can review :)

kbengs (commented):
Sorry for commenting again. I had this written before but forgot to push on "submit review" Also if you can clean the history (force push) so that you do not have so many commits. (ie not have commits where you fix your own code) Maybe one commit for the code changes, one for the test and one for the translation is enough.

selfhoster1312:
Applied your suggestions and rebased on main. I believe the test may still fail.

If you know how to fix the test feel free to push on my branch directly! Since i cannot run the tests locally it's not very convenient for me...

tests/test.py
```diff
@@ -729,7 +729,19 @@ def test_05_buggy_exif(self):
         self._wait_cond(lambda: filechooser.dead)
         self.assertEqual(len(self._icons()), 3)
 
-    def test_06_quit(self):
+    def test_06_split_booklet(self):
+        self.test_O3_booklet()
```
kbengs, line 733: You probably just need to remove this line self.test_O3_booklet()

kbengs: or change O to 0..

kbengs:
I think its not possible to force push to your branch so I could not make the tests in one commit. Could you fixup the two test commits into one?

PR #1167, opened 2024-12-15
Fix 1161

kbengs:


kbengs:
I don't know if the bug is in poppler or in the print dialog or if printing in pdfarranger was done in a "unsupported way". Anyway this fixes it. Printing will be somewhat slower because it now always exports the pages first (even when there are no modifications done to pages).

PR #1169, opened 2024-12-18
Retain XMP metadata when saving

### LINKED ISSUE #1168: Metadata cut off at saving
Describe the bug
----------------

PDFArranger takes metadata primarily from the docinfo dictionary. If they differ from the corresponding XMP metadata the latter will be overwritten at saving with the docinfo metadata. This is in constrast to pikepdf where XMP metadata are retained at saving.

To Reproduce
------------

1.  Download this file where the creator is a single name in docinfo and an ordered array in XMP: https://link.springer.com/article/10.1007/s10853-017-1978-5
2.  Open it with PDFArranger
3.  Save it to say output-pdfarranger.pdf
4.  check the metadata with your favorite pdf viewer

Expected behavior
-----------------

Keep author (=creator) metadata as is, as does pikepdf:
```
with pikepdf.open("s10853-017-1978-5.pdf") as pdf:
    pdf.save("output-pikepdf.pdf")
```

Input files
-----------

https://link.springer.com/article/10.1007/s10853-017-1978-5

Screenshots
-----------

Wrong result when saving with PDFArranger:
grafik
```
         <dc:creator>
            <rdf:Seq><rdf:li>O. Barrera </rdf:li></rdf:Seq></dc:creator>
```

Correct result when saving with pikepdf:
grafik
```
			<dc:creator>
				<rdf:Seq>
					<rdf:li>O. Barrera </rdf:li>
					<rdf:li>D. Bombac </rdf:li>
					<rdf:li>Y. Chen </rdf:li>
					<rdf:li>T. D. Daff </rdf:li>
					<rdf:li>E. Galindo-Nava </rdf:li>
					<rdf:li>P. Gong </rdf:li>
					<rdf:li>D. Haley </rdf:li>
					<rdf:li>R. Horton </rdf:li>
					<rdf:li>I. Katzarov </rdf:li>
					<rdf:li>J. R. Kermode </rdf:li>
					<rdf:li>C. Liverani </rdf:li>
					<rdf:li>M. Stopher </rdf:li>
					<rdf:li>F. Sweeney </rdf:li>
				</rdf:Seq>
```

System and Versions
-------------------

-   1.11.1
-   8.5.0
-   Windows 10

StefRe:
XMP metadata should not be overwritten by their corresponding docinfo entries.

Fixes #1168

### DIFF
--- a/pdfarranger/metadata.py
+++ b/pdfarranger/metadata.py
@@ -93,16 +93,27 @@ def _safeiter(elements):
 
 
 def merge_doc(metadata, input_docs):
-    """Same as merge but with pikepdf.PDF object instead of files"""
+    """Same as merge but with pikepdf.PDF object instead of files
+
+    XMP metadata take precedence over equivalent docinfo metadata,
+    metadata of later opened files are merged into these of earlier opened ones
+    """
     r = metadata.copy()
     for doc in input_docs:
         with doc.open_metadata() as meta:
+            for k, v in _safeiter(meta.items()):
+                if not _pikepdf_meta_is_valid(v):
+                    # workaround for https://github.com/pikepdf/pikepdf/issues/84
+                    del meta[k]
+                elif k not in r:
+                    r[k] = v
+            # workaround for https://github.com/pdfarranger/pdfarranger/issues/1168
             load_from_docinfo(meta, doc)
             for k, v in _safeiter(meta.items()):
                 if not _pikepdf_meta_is_valid(v):
                     # workaround for https://github.com/pikepdf/pikepdf/issues/84
                     del meta[k]
-                elif k not in metadata:
+                elif k not in r:
                     r[k] = v
     return r
 
--- a/tests/test.py
+++ b/tests/test.py
@@ -140,6 +140,13 @@ class DogtailManager:
 # dogtail does not support change of X11 server so it must be a singleton
 dogtail_manager = DogtailManager()
 
+def setUpModule():
+    # if we can't kill the dbus process on GHA, let's at least suppress the warning
+    if dogtail_manager.xvfb and "GITHUB_ACTIONS" in os.environ:
+        warnings.filterwarnings("ignore",
+                                f"subprocess {dogtail_manager.xvfb.dbus_proc.pid} is still running",
+                                ResourceWarning)
+
 def tearDownModule():
     dogtail_manager.kill()
 
--- a/tests/test.py
+++ b/tests/test.py
@@ -375,6 +375,7 @@ class PdfArrangerTest(unittest.TestCase):
         return config
 
     def _quit(self):
+        self._app().child(roleName="layered pane").grabFocus()
         self._app().child(roleName="layered pane").keyCombo("<ctrl>q")
 
     def _quit_without_saving(self):
@@ -979,10 +980,11 @@ class TestBatch10(PdfArrangerTest):
 
     def test_03_quit(self):
         self._quit()
+        self._process().wait(timeout=22)
 
 
 class TestBatch11(PdfArrangerTest):
-    """Test show-save-warnings config setting"""
+    """Test show-save-warnings config setting and metadata"""
 
     def test_01_save_and_disable_warning(self):
         """Show warning discarded metadata (default)"""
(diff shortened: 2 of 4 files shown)


StefRe:
This is a quick and dirty fix. It would be easier if [`load_from_docinfo`](https://pikepdf.readthedocs.io/en/latest/api/models.html#pikepdf.models.PdfMetadata.load_from_docinfo) had a parameter `overwrite=True` which we could set to `False` here.

kbengs:
I guess it does the job.

Another strange thing I noticed while looking at this is that pdfarranger tries to preserv "outlines and links" from the first opened document. But with metadata it is the opposite way: metadata in documents added later does overwrite metadata added earlier. Or am I wrong?

StefRe:
> metadata in documents added later does overwrite metadata added earlier

Yes this is correct. I guess it's indeed more intuitive to do it the other way round (to be in-line with the way outlines and links are handled).  

I think this can be achieved easily by just changing
https://github.com/pdfarranger/pdfarranger/blob/3680096f818693628919ff71c8ea093dc99db523/pdfarranger/metadata.py#L98

to 
```python
for doc in reversed(input_docs): 
```

Strictly speaking this is, however, an API change. Not sure what PDFArranger's versioning policy is. Does it follow semantic versioning?

dreua:
> Not sure what PDFArranger's versioning policy is. Does it follow semantic versioning?

I'm for roughly following semver but not overthinking it :wink: Semver as I understand it is very useful for libraries, it is not really designed for end user GUI applications. To my knowledge the metadata import precedence is not documented so I hope no one is relying on it being consistent. I'd say we can change that in any version if it make sense. (Which it does imo.)

StefRe:
I added the change of the sequence of metadata merging as a separate commit for clarity.

StefRe:
Converting to draft until #1166 goes in.

StefRe:
This one turns out to be tricky:
* Regarding https://github.com/pdfarranger/pdfarranger/pull/1169#issuecomment-2555006463: just moving `load_from_docinfo` already fixes the correct merging sequence (no need to reverse the list of `input_docs`), although I must admit that I don't completely understand why. The fact is, however, that the merging sequence was wrong and now it works, see test `test_03_retain_metadata`.
* Handling of custom entries in the info dict was obviously changed (fixed) between pikepdf 6 and 8 but I couldn't find the relevant commit. This is the reason for `if not have_pikepdf8():; del expected["/Custom"]` in the test.
* The main problem I'm struggling with is that I can't wait for the pdfarranger process terminating after quitting in `TestBatch10.test_03_quit` and `TestBatch11.test_03_retain_metadata`: `self._process().wait(timeout=22)` will time out after 22 seconds. I can leave this line out and the process will be killed in `tearDownClass` but this leads to the ugly warnings `UserWarning: resource_tracker: There appear to be 5 leaked semaphore objects to clean up at shutdown`, see for instance [here](https://github.com/pdfarranger/pdfarranger/actions/runs/12512097952/job/34905025562#step:9:244). I probably can't see the forest for the trees but I can't get it to work on CI. It runs fine locally: the process terminates in less than 1 sec after `<ctrl>q` (python 3.13.1). I though that some kind of deadlock appears (introduced maybe by capturing stdout in 81c867c) so I tried redirecting stdin to `DEVNULL` as recommened on SO but it doesn't help. **Maybe someone can take a look and give me a tip.**

StefRe:
@kbengs Thanks a lot for this fix. This is really strange, obviously there are some subtleties in the setup that make the test run fine locally but fail on CI. Usually there are warnings like "Node is not focusable; trying key combo anyway" in this case that I don't see here.  
I think the remaining `ResourceWarning` most probably comes from `test_01_cli_version` in `TestBatch9`, will take a closer look at it later.

tests/test.py
```diff
@@ -1002,8 +1005,40 @@ def test_02_save_without_warning(self):
...
+        self._process().stdout.close()
+        self._process().wait(timeout=22)
         self.assertEqual(self._config()["preferences"]["show-save-warnings"], "False")
 
+    def test_03_retain_metadata(self):
+        """Merge metadata into first file and don't overwrite from info dict"""
+        self._start(["tests/test_metadata1.pdf", "tests/test_metadata2.pdf"])
+        self._mainmenu("Save")
+        self._save_as_chooser("output2.pdf", ["output2.pdf"])
+        self._wait_saving()
```
kbengs, line 1017: _save_as_chooser already has _wait_saving

pdfarranger/metadata.py
```diff
@@ -93,17 +93,28 @@ def _safeiter(elements):
...
+    """
     r = metadata.copy()
     for doc in input_docs:
         with doc.open_metadata() as meta:
-            load_from_docinfo(meta, doc)
             for k, v in _safeiter(meta.items()):
                 if not _pikepdf_meta_is_valid(v):
                     # workaround for https://github.com/pikepdf/pikepdf/issues/84
                     del meta[k]
                 elif k not in metadata:
```
kbengs, line 108: This actually still use the old behavior (later added values overwrites previous added). Insted this could also be `elif k not in r:`

I think the reason the test files work fine for the "Title" is that the value is not i XMP but only in docinfo, and docinfo overwriting is fixed but not xmp overwriting.

StefRe: Yes, you're right. Fixed and amended test pdf files to cover this case.

tests/test.py
StefRe: Yes, thanks. This was a leftover from my desperate attempts to let the process terminate. Fixed.

StefRe:
> I still see `./usr/lib/python3.10/subprocess.py:1070: ResourceWarning: subprocess 48 is still running` which I think is caused by capture of stdout also.

This is the D-Bus daemon started in XvfbManager. Killing it on GHA will terminate the whole build job (not sure why exactly, but I couldn't find a way to do it and keep the job running), so we will at least suppress this warning to not pollute the log.

PR #1173, opened 2024-12-23
Show program name in Windows "open with" context menu

### LINKED ISSUE #1170: `Right-Click on PDF > Open With` is not showing "PDF Arranger"
I installed PDF Arranger (version 1.11.1) on Windows 11 (for all users of a workstation) with this winget command, in an (admin) Powershell:

`winget install PDFArranger.PDFArranger --scope machine --source winget`

While this did successfully install PDF Arranger, when a user right-clicks on a PDF file and goes to the `Open With` context-menu option, PDF Arranger isn't listed there by default for my users. To remedy this I copied and pasted this into a Powershell as admin:

```
# Define the application path for PDF Arranger
$appPath = "C:\Program Files\pdfarranger\pdfarranger.exe"

# Define the registry path for adding the "Open with PDF Arranger" option for PDF files
$RegistryPath = "HKLM:\SOFTWARE\Classes\PDFArranger.PDF\shell\Open with PDF Arranger\command"

# Create the registry keys if they do not exist
if (-not (Test-Path $RegistryPath))
{
    New-Item -Path $RegistryPath -Force | Out-Null
}

# Set the default value for the command to open PDF files with PDF Arranger
$Command = "`"$appPath`" `"%1`""
Set-ItemProperty -Path $RegistryPath -Name "(Default)" -Value $Command

# Register the new handler under the .pdf file association
$FileAssociationPath = "HKLM:\SOFTWARE\Classes\.pdf\OpenWithProgids"
if (-not (Test-Path $FileAssociationPath))
{
    New-Item -Path $FileAssociationPath -Force | Out-Null
}

# Add the new file association for PDFs with PDF Arranger
New-ItemProperty -Path $FileAssociationPath -Name "PDFArranger.PDF" -PropertyType String -Value "" -Force | Out-Null

```

**This issue requests that the PDF Arranger installer automatically register itself as a PDF file handler during Windows installation.**

This issue also requests that the application's menu entry be changed:

See where it says: "A simple application for PDF Merging, Rearranging, and Splitting" ?

That should probably be either "PDF Arranger" (by itself) or "PDF Arranger: A simple application for PDF Merging, Rearranging, and Splitting". Typically, (as you can see from other entries in that menu) it would just be the application's name alone.

The commands I provided above do not control the text shown in the screenshot above. That's likely coming from a registry entry added by the PDF Arranger installer or by the [package metadata provided to winget](https://github.com/microsoft/winget-pkgs/issues/199790).

Personally, I don't mind the length of the entry you're seeing in the menu above, but I do think that it should first say "PDF Arranger:" prior to that pertinent description.

In closing, I want to thank the developers of PDF Arranger. I've been using it for years on Linux, but only recently discovered it was cross-platform, and I'm very happy to extend its goodness to all my users. In my opinion, the application was very well done and ought to be on every business computer in the world!

StefRe:
instead of program description. The existing description is moved to `long_description` which goes into the `comment` field of the `VersionInfo` struct.

Closes #1170.


dreua:
To clarify: This just changes the name which is shown, not whether it is or isn't shown at all.

PR #1178, opened 2024-12-28
Update German translation

StefRe:
and show number of untranslated messages in `updatepo.sh`. Restore `pdfarranger.pot` from HEAD if it was generated for a single locale, following the instructions in [README.md](https://github.com/pdfarranger/pdfarranger?tab=readme-ov-file#for-translators).

There are a number of changes of line wrapping in de.po that I don't know where they come from. We don't use any options or config files for msgmerge and I didn't update gettext (still using version 0.22.5). Maybe we should explicitly specify the width ([--width=n](https://www.gnu.org/software/gettext/manual/html_node/msgmerge-Invocation.html#Operation-modifiers)) to keep it constant?

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -12,8 +12,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: pdfarranger\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-11-10 19:24+0100\n"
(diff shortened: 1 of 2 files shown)


dreua (approved):

kbengs:
> There are a number of changes of line wrapping

I have not noticed that before. Lets see if it happens again.

PR #1182, opened 2025-01-04
Update hu (Hungarian - magyar) translations

ovari:


dreua:
What is the source of those translations?

ovari:
> What is the source of those translations?

Computer translation, except for `hu`. Feel free to update `de` and/or any other ones you like.

dreua:
I'm against merging this with computer generated translations. Furthermore it is not clear to me why we would need the "caption" here at all or what value it provides.

ovari:
Revert computer generated translations.

The caption would show on https://flathub.org/hu/apps/com.github.jeromerobert.pdfarranger

Thank you

dreua:
Yeah I already had a feeling that this might have something to do with flathub. Still it does not answer my question of what purpose it servers or what value it adds. (We only have one screenshot and that is obviously "The PDF Arranger main window" :shrug:

dreua:
Full disclosure: I have some issues with how flathub operates, some of the decisions they are making and especially with the lack of transparency and community involvement in their decision making process,

ovari:
Reverted `<caption>` tag to enable this patch to be merged.

Please advise if you would like to discuss on [Jami](https://jami.net/) reasons for and against including `<caption>` tags. Jami is more effective than writing here on GitHub.

Thank you

dreua:
I didn't mean to say that I'm decidedly against merging the captions, just that I'd like the reasoning behind it in order to make that decision. (Or I might have left that decision to someone else.) But I'm happy with the translations only, thanks!

PR #1185, opened 2025-01-08
TST: Verify that setting are saved to config file

StefRe:
For content-loss-warning this was implicitly verified by running the next test batch. Now that each test batch starts with a fresh config file, this should be verified explicitly.

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -569,6 +569,7 @@ class TestBatch1(PdfArrangerTest):
         self._quit()
         # check that process actually exit
         self._process().wait(timeout=22)
+        self.assertEqual(self._config()["preferences"]["content-loss-warning"], "False")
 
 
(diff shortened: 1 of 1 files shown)


StefRe:
Hm, there seem to be some changes in the Ubuntu 24.04 image between version 20241215.1.0 and 20250105.1.0 that finally breaks to dogtail test setup. The [relevant commit](https://github.com/actions/runner-images/commit/bb3043b2c5169c9d969cc4ffd02856b5bc5d3b9c#diff-cc12e5c5d005932feaacf280af949d83e2cb6020dff8203d9ead2e00c30e8b7fR10) doesn't show anything obvious.  I am quite at a loss.

dreua:
I just reran the test :shrug:

Issue #1131, opened 2024-10-20
Split Pages… tool duplicates pages’ content


linkmauve:
Describe the bug
----------------

After having used the Split Pages… tool, the resulting PDF contains the original merged pages duplicated and cropped, which looks correct when displayed (e.g. in Evince) but not for editing (e.g. in Inkscape).

To Reproduce
------------

1.  Open a PDF
2.  Select all pages
3.  Right click, Split Pages…
4.  Save the result
5.  Open the result in Inkscape

Expected behavior
-----------------

pdfarranger could try to remove the objects outside of the canvas of each page after the Split Pages… operation.

Input files
-----------

This happens with all PDF files, but here is a sample file exhibiting this issue: ![drawing.pdf](https://github.com/user-attachments/files/17449911/drawing.pdf)

Screenshots
-----------

Once open in an editing tool (such as Inkscape), every page is duplicated:
![image](https://github.com/user-attachments/assets/91c63efe-db9f-48f9-8d44-aa020698106a)

System and Versions
-------------------

-   PDF Arranger version and where you get it: 1.11.1-1 from ArchLinux
-   pikepdf version and where you get it: 9.3.0-1 from ArchLinux
-   Inkscape version and where you get it: 1.4-1 from ArchLinux
-   OS name and version: ArchLinux up to date

dreua:
This is expected and should be explained in the first run dialog. I believe this is actually not duplication but it displays the same content at different places of different pages. (i.e. there is no storage wasted; it's just a visual problem in Inkscape.)

## Workarounds
I wonder if there is any "box" (media, crop, etc) we can set which would cut it away in Inkscape, too. Maybe there is or could be a feature in inkscape to cut away stuff or just display one page per view.

In genereal, PDFs can be expected to have content reaching over the page borders, have you checked whether Inkscape has options to deal with this or opened a bug report there?

## Expectation management
> pdfarranger could try to remove the objects outside of the canvas of each page after the Split Pages… operation.

I think this is a complexity and needs a deeper understanding of PDFs which we don't want to have and maintain in PDF Arranger. (Just imagine an image or a text block crossing the split border. And every now and then a new PDF feature turns up which I have never heard of before.)

I'll leave that open for a few weeks and see if there is any more comments but I guess this will end up as wontfix, sorry.

tmaier:
I understand that "hiding" is currently expected behaviour, but it would be a great improvement of the feature to properly support "removing" content.

[...]

Issue #1090, opened 2024-05-11
Add ability to export single/multiple pages as images


archisman-panigrahi:
Describe the bug
----------------

Right now, pdfarranger can only export as pdf. Sometimes one has to extract a single file from a pdf and use it in image format.
For that, right now one has to extract the page(s) as pdf and then manually convert them to image format.

Having the option to export as png/jpg will reduce the last step.

To Reproduce
------------

1.  Select some pages
2.  Right click, and click export.
3.  All the exported pages are saved in pdf format. Even if I try to save as `filename.png`, it saves as `filename.png.pdf`.

Expected behavior
-----------------

Step 3 should have some way to produce png/jpg.

Input files
-----------

NA

Screenshots
-----------

NA

System and Versions
-------------------

-   PDF Arranger version and where you get it - 1.10.1, Ubuntu 24.04 APT package
-   pikepdf version and where you get it - 8.7.1 APT package
-   OS name and version - Ubuntu 24.04

dreua:
Hi and thanks for the suggestion! I, however, think that this is a rather niche use case. If you do this on a regular basis there are probably better ways to automate this with some scripting. If anyone wants to work on this feature please let us know :wink: 

As per our usual procedure with such feature requests: Let's give this 3 weeks. If nothing productive comes up it should be closed and tagged `stale feature`.

dreua:
Another request in discussions: https://github.com/pdfarranger/pdfarranger/discussions/1145
Feel free to use the :+1: reaction to show interest and Subscribe to get notifications in case there should be any development.

mak448a:
Please reopen this

dreua:
@mak448a Reopening this wouldn't change a thing but clutter the list of "open" features. There are projects with hundreds of open features but most of them are never actually care taken care of, therefore I don't understand the request to "reopen" this. Anyone can comment or leave their :+1:  here, no matter if it is open or closed. For implementation we need someone to implement and maintain the feature on all supported platforms. If someone willing to do that turns up or a relevant discussion takes place here I'm happy to reopen this feature :wink:

Heavygrass:
Hi,
I'd like to request export to PNG/JPG also.
It would be a very usefull feature!
Thanks

PR #1188, opened 2025-01-26
Add support for exporting to png, jpg and rasterized pdf

kbengs:


kbengs:
Export to images was requested in #1090 and #1145

Export to rasterized pdf is a way to remove the cropped parts from a pdf. This was requested for example in #1131
It will "preserve" the metadata but remove everything else. It can thus be used to remove any hidden content in a pdf and possibly make the file size smaller.

The settings are in the preferences dialog: "Pixels/inch", "Optimize" and "Convert to greyscale"

Does someone have comments on this?

dreua:
> any hidden content

We should be careful with the wording imo, hidden content can be in metadata and I believe also in image alpha channels in special cases. Of course we don't accept any liability by license but still I wouldn't want any confidential information of our users leaked due to incorrect communication on our side.

I guess it is good to give users what they want/need while I'm personally not a big fan of rasterization as it is bad for accessibility, quality, size/bandwidth, search-ability. Lets hope that our users will use it responsibly and only in those cases where it is necessary :)

That's just some high level comments, I haven't looked at the code or tried it.

kbengs:
>not a big fan of rasterization

I am not really a fan either, but there are cases when it make sense. When the content is already rasterized it should be fine to export to rasterized pdf (scanned pdf -> split booklet). If the "Pixels/inch" is set to the same as it was scanned to, the exported quality should be fine.

Yes i guess some hidden content could be in the pixel data even though there is no alpha channel in the exported pdf here. I changed the wording a bit. :-) We could of course choose to remove all metadata too.

PR #1190, opened 2025-02-03
Add "open with" entry on Windows

dreua:


dreua:
Works great on Windows 10, it would be nice if someone could test on 11 but I'd say this is good to be merged even without that test. Fixed the icon while I was at it, it always looked bad to me in the Start menu.

The dogtail tests are failing I'm pretty sure that has nothing to do with the changes here.

kbengs:
Works fine on Windows 11 also.

PR #1204, opened 2025-02-26
Fix #1202: make optional the saving of window geom on exit

### LINKED ISSUE #1202: Add config file option to disable saving of window geometry to config file
I like to keep my config files under revision control.  
Pdfarranger is a bit aggressive in saving even window geometry back to the config file.  On linux, config files typically have "permanent" preferences, and transient changes like window size are stored in a "state" file instead.
Rather than separate the transient from the permanent, it would be sufficient to simply make the saving of window geometry on exit a configurable option.
I will propose a PR that implements this feature.

jeffkowalski:
Introduce a config option "save-window-geometry", which defaults to
"true", and controls whether transient window properties will be written
to the config file upon exit.

By setting the option to "false" in the config file, transient window
properties will not be saved, and the config file will be modified only
if there are other (more important) preference changes.

### DIFF
--- a/pdfarranger/config.py
+++ b/pdfarranger/config.py
@@ -117,6 +117,9 @@ class Config(object):
         self.data.read(Config._config_file(domain))
         if 'preferences' not in self.data:
             self.data.add_section('preferences')
+        preferences = self.data['preferences']
+        if 'save-window-geometry' not in preferences:
+            preferences['save-window-geometry'] = 'true'
         if 'print-settings' not in self.data:
             self.data.add_section('print-settings')
         if 'image-export' not in self.data:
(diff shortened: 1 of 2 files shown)


jeffkowalski:
see PR https://github.com/pdfarranger/pdfarranger/pull/1204

dreua (commented):
Looks okay to me, it adds a few lines for a very specific use case :shrug:

jeffkowalski:
It is a narrow problem, but the change tries to better adhere PDF Arranger to XDG specs.  Program state should be stored outside the config file.

In the XDG Base Directory Specification, "xdg config" refers to the directory where user-specific configuration files should be stored, while "xdg state" refers to the directory for storing user-specific state data, which is essentially temporary data that an application needs to maintain its current working state between restarts, but isn't as important or portable as regular user data like configuration settings; essentially, "config" is for settings that affect how an application behaves, while "state" is for data needed to restore the application to its current position when it restarts.

PR #1205, opened 2025-02-26
ignore pdfarranger.egg-info in git

jeffkowalski:
The egg-info directory is not source code.  It is the product of building using pip[3x].


dreua:
Ideally your commit messages would match the repository style which starts with an uppercase letter.

Issue #1215, opened 2025-04-10
Webp support


AnyaVanelli989:
First of all, thank you for creating and maintaining PDF Arranger. It's an incredibly useful tool, and I truly appreciate the effort that goes into its development.

I would like to suggest adding support for the .webp image format. This format is becoming increasingly popular for web images due to its excellent compression and quality balance. Currently, PDF Arranger supports various image formats, but .webp compatibility would make it even more versatile, especially for users dealing with modern image standards.

For example:

Users often need to merge .webp files with PDFs or other image types, and adding this support would streamline the process.

It could open the tool to a wider audience who frequently work with .webp files.

Thank you for considering this suggestion! I am happy to provide further details if needed.

Best regards,Anya Vanelli

dreua:
Hi, thanks for kind words and the suggestion!

The image import functionality is provided by the python library img2pdf, therefore the best way would be to suggest and maybe implement it there. I've been thinking the same thing for SVG images recently. From my (rather short) research it looks like converting svg to pdf is not that easy and the best way I found was using inkscape to create a PDF. I haven't looked into webp and PDF, specifically whether it is possible to embed it without transformation. (SVGs need to be transformed, either rasterized - which is bad for quality - or with an advanced tool like inkscape capable of converting the SVG elements to PDF elements.)

https://gitlab.mister-muffin.de/josch/img2pdf

dreua:
Disregard my first answer, sorry. I just tried opening a webp as cli argument like `pdfarranger my.webp` and it just worked. Dragging it into the gui throws an error, that is probably ours to fix. This is on Fedora linux with libwebp installed (should be present on almost all desktop linux systems). We'd need to check if additional checks or packaging will be necessary on Mac/Windows. Could someone try it there?

Background: img2pdf used pillow for image formats and that already [supports webp](https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#webp).

dreua:
Argh I had multiple versions installed. Updated results: The rpm installation on Fedora supports webp regardless of how you import it. (There is just one specific webp which causes an error in img2pdf more on that later.)

The Flatpak installation doesn't support webp at all, it just says "Unknown file format" in a small dialog box as well as the console no matter how you try to import it. (png works fine for example)

What platform installation method and version are you using @AnyaVanelli989 ?

dreua:
[...]

PR #1220, opened 2025-04-15
Disable strict mimetype checking

dreua:
This should fix the import of webp images on Windows and Flatpak installations.

Closes: https://github.com/pdfarranger/pdfarranger/issues/1215


kbengs:
Also confirmed on Windows 10. I guess it should work on Windows 11 also. But I can test it next week.

kbengs (approved):

PR #1231, opened 2025-05-26
Increase timeout

dreua:
Github runs often fail inexplicably and succeed on rerun, let's try with a 10x increased timeout.


kbengs:
I rerun it twice and it worked both times. :-) Maybe this is the fix. Okay to merge imo.

PR #1242, opened 2025-06-25
Update windows build instructions and remove darkdetect

kbengs:


dreua (approved):
Thanks for the explanation, that sounds good, I just wanted to know the reason and figuring this out on my own would be much harder than asking ;)

Code looks good to me, can be merged if tested successfully on windows.

pdfarranger/pdfarranger.py
```diff
@@ -351,9 +351,19 @@ def __resource_path(path):
 
     def set_color_scheme(self):
         if Handy:
+            win_dark = False
+            if os.name == 'nt':
+                # Read from Registry whether Windows use dark theme
+                p = "Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize"
+                try:
+                    key = OpenKey(HKEY_CURRENT_USER, p)
```
dreua, line 359: Style nitpick: one letter variable names are often discouraged in favor of longer names which give the reader some context. Maybe the string could be inlined here (breaking the statement in multiple lines if necessary). But I do not insist on this being changed, it is close enough to the definition to not cause confusion. Thanks a lot for your contributions :)

dreua:
Has anyone tested the darkmode switching with this on windows? I'd like to merge this now as we already got another PR fixing the cx-freeze name.

PR #1245, opened 2025-07-03
Improve undo

kbengs:
This will make it easier to see what is changed when going back in undo history.

### DIFF
--- a/pdfarranger/undo.py
+++ b/pdfarranger/undo.py
@@ -23,6 +23,17 @@ only store snapshots of the GtkListStore object, not of
 the whole PDF files.
 """
 
+from gi.repository import GObject
+from dataclasses import dataclass
+from .core import Page
+
+
+@dataclass
+class State:
+    label: str
+    pages: list[Page]
+    selection: list[int]
+
 
 class Manager(object):
     """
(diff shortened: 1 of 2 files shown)


pdfarranger/undo.py
```diff
@@ -51,23 +53,36 @@ def commit(self, label):
...
+        1. The label of the action
+        2. The pages
+        3. Which page numbers are selected
+        4. The vertical adjustment percentual value
+        """
+        pages = [row[0].duplicate(False) for row in self.model]
+        s = self.app.iconview.get_selected_items()
+        selection = [path.get_indices()[0] for path in s]
+        vadj_percent = self.app.vadj_percent_handler()
+        return (self.label, pages, selection, vadj_percent)
```
dreua, line 74: Maybe that could be something like a small `@dataclass` instead of a tuple of four values just put together in a tuple. When reviewing this, it really is harder to understand and there is a risk of mixing stuff up and issues when changing/extending those tuples in the future.

(I know we are having code like this in a lot of places, around here. But maybe we could start with cleaner code when adding/modifying stuff.)

dreua (approved):
I haven't tested this but I don't see anything obviously wrong in the code. It is hard to understand at some places (mostly due to the original code, not the changes though).

kbengs:
I made it use `@dataclass`. Does it look okay?

PR #1266, opened 2025-09-14
docs(metainfo): Add link to source repo

salim-b:
cf. https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines#url


dreua (approved):
lgtm, thanks!

PR #1267, opened 2025-09-15
Fix booklet

kbengs:


kbengs:
#1269 should fix the failing crop-white-border test. It should go in before this PR.

PR #1269, opened 2025-09-27
Fix crop-white-borders for pages with overlays/underlays

kbengs:


dreua (approved):
Not tested by me at all or understood in depth but I'm fine with it being merged :) Thanks for your ongoing development efforts :+1: 

(I'm kind of in a process of adjusting my own expectation of code review in order to give you faster feedback. It is very possible that my previous standards or idea of a review were far too high and I don't want this to hold up development.)

PR #1276, opened 2025-10-17
Support resizing of pasted overlay/underlay

kbengs:


kbengs:
To more easy set the right size of the overlay:

resize-pasted-page

PR #1279, opened 2025-10-22
Remove iconview selection_changed event connection

kbengs:


dreua (approved):
Code looks good and readable to me (even more than before with the simplifications).

Not sure what the risk of hidden bugs is, I mean if you missed a way of introduced change then it would misbehave, wouldn't it? Might be hard to find that later but the speedup is sweet; probably worth the risks, therefore approved.

PR #1281, opened 2025-10-29
Update Portuguese translations in pt_PT.po

horus68:
Small fixes to pt-PT language

### DIFF
--- a/po/pt_PT.po
+++ b/po/pt_PT.po
@@ -130,7 +130,7 @@ msgstr "Recortar e adicionar margens"
 
 #: pdfarranger/pageutils.py:942
 msgid "Crop Margins"
-msgstr ""Recortar margens"
+msgstr "Recortar margens"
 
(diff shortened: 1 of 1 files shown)


po/pt_PT.po
```diff
@@ -129,7 +130,7 @@ msgstr "Recortar e adicionar margens"
 
 #: pdfarranger/pageutils.py:942
 msgid "Crop Margins"
-msgstr "Margens de recorte"
+msgstr ""Recortar margens"
```
dreua: There is an extra quote here.


```
  po/pt_PT.po:133: keyword "Recortar" unknown
  po/pt_PT.po:133:18: syntax error
  po/pt_PT.po:134: end-of-line within string
  msgfmt: found 3 fatal errors
  ```

dreua (changes requested):

po/pt_PT.po
horus68: fixed

PR #1288, opened 2025-11-23
Add support for searching of text in PDF

kbengs:
Does look like this:
Screenshot_20251123_170539

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -206,6 +206,7 @@ class PdfArrangerTest(unittest.TestCase):
 
     def _mainmenu(self, action):
         mainmenu = self._app().child(roleName="toggle button", name="Menu")
+        mainmenu.grabFocus()
         self._wait_cond(lambda: mainmenu.sensitive)
         mainmenu.click()
         if not isinstance(action, str):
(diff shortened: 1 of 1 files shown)


pdfarranger/pageutils.py
```diff
@@ -1059,6 +1059,190 @@ def get_offset(self):
...
+            doc = self.pdfqueue[page.nfile - 1]
+            poppler_page = doc.get_page(page.npage - 1)
+        else:
+            # Page has layers -> for simplicity export and search in exported pdf.
+            # This is slower but easier than recalculating rectangle coordinates.
+            page.angle = 0
+            page.scale = 1
+            page.crop = Sides()
+            page.hide = Sides()
+            doc, buf = get_in_memory_poppler_doc([page], self.pdfqueue)
```
dreua: Maybe we can use _ instead of buf here to make codacy happy.

pdfarranger/pageutils.py
```diff
@@ -1059,6 +1059,190 @@ def get_offset(self):
...
+                ):
+                to_remove.append(num)
+        for num in reversed(to_remove):
+            del rectangles[num]
+        return rectangles
+
+    def close(self, _widget):
+        self.clear_find_results(unselect_all=False)
+
+
```
dreua: This file is getting quite long, how would you feel about splitting it into multiple files? (I haven't looked deeper into whether that is possible.)

kbengs: Yes it's long. But I am not sure how we should split it. I think that will be for another PR.

kbengs:
I think it is ready.

PR #1293, opened 2025-12-13
Run Windows build with console printing

dreua:


mffm-NET:
This just make me ask "do we have a CLI to pipe in operations" via a batch script or like similar API calling?

kbengs:
The app can take pdf files as arguments. pdfarranger -v print software versions. The Windows msi can take a argument that will make it install to programFiles. That's all I am aware of.

This PR is about getting debug info available in terminal in Windows. Same as you will have when it is run from a terminal i Linux.

PR #1301, opened 2025-12-24
Map MIME type image/jp2 to extension .jp2

kbengs:
jp2 mimetype added to cpython Nov 18, 2024. I'm not sure in which python version it is available. At least on python 3.13.5 it works fine.


kbengs:
I looked some more into it. MIME type is added in python 3.14.0. It can still work before that because the operating system can have other locations which are searched also. On Debian a list is found in /etc/mime.types which makes it work on my machine with python 3.13.5

PR #1304, opened 2025-12-31
Add pyproject.toml file

kbengs:


kbengs:
This was done because with pip 25.1 this warning is printed when installing pdfarranger:
"DEPRECATION: Building 'pdfarranger' using the legacy setup.py bdist_wheel mechanism, which will be removed in a future version. pip 25.3 will enforce this behaviour change."

If the PR is merged the warning is gone on pip 25.1. But pdfarranger will no longer install on older distros such as Debian 11 because of too old setuptools available.

If the PR is not merged the warning is gone when pip is 25.3. This is because pip defaults to "PEP 517 mode". PDF Arranger still installs fine.

Therefore I currently think it would be best to not add a pyproject.toml file (yet) and close this PR. Do you @dreua have a opinion?

More info is found here: https://github.com/pypa/pip/issues/6334

kbengs:
pip was upgraded with `~/myenv/bin/pip3 install --upgrade pip` (following our readme first) pip will fetch the needed setuptools in venv. PDF Arranger can't be installed without a virtual environment on a modern distribution anyway (with pip), as far as I know.

I don't know how this affects building of deb, rpm.. packages on distributions. Are you thinking this would be a problem also?

Anyway we can put this PR on hold if you think so? As I already tested it will work fine to install PDF Arranger also without the pyproject file when pip is >= 25.3.

PR #1351, opened 2026-03-10
Add continuous PDF preview feature

m4rtxn:
This feature allows a continuous preview of the final document (vertical scroll style panel). This panel updates in real time as the user adds, deletes, or reorders pages. A button has also been added to the toolbar to enable or disable this preview.

image

### DIFF
--- a/data/pdfarranger.ui
+++ b/data/pdfarranger.ui
@@ -115,6 +115,24 @@ along with PDF Arranger.  If not, see <http://www.gnu.org/licenses/>.
           </object>
         </child>
         <child>
+          <object class="GtkToggleButton" id="preview_toggle_button">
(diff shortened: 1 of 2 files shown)


kbengs:
Sorry but I vote for not merging this. I think the app is handling the preview so well already.

mffm-NET:
this would only make sense when have a page view "full screen" and the vertical is used as navigation. 
Making the window of the App small enough to just view one page horizontal would just make the "read vertical problem obsolete as it kind of emulate it than
But I must say a "preview" of how double page layouts for brochure layouts would look were two pages are viewed together without the middle space, but edge to edge would be maybe useful. First and last page of a document would need then be shown separately (cover Pages)

PR #1364, opened 2026-04-10
Improvements to PaperSizeWidget

kbengs:
This was discussed in #1359
New look to the left and old to the right.
image


dreua (approved):
UI generation code looks messy before and afterwards, can be merged from my side :wink:

PR #1365, opened 2026-04-10
Fix false "unsaved" state on accidental micro-drag

### LINKED ISSUE #1363: False "unsaved" state on accidental micro-drag - accessibility concern for motor-impaired users
First of all, thanks for maintaining pdfarranger :-)

I ran into a usability issue that I think is worth fixing, especially for users with reduced motor precision.

**What happens:**

When a user clicks on a page thumbnail and moves the mouse by just a few pixels before releasing (accidental micro-drag), pdfarranger marks the document as modified and shows a "Save changes?" dialog on close, even though no page was actually reordered.

This is particularly confusing for elderly users or people with tremor, for whom a slight unintentional mouse movement during a click is common. The unexpected save dialog creates confusion ("Did I accidentally change something?") and can lead to data loss if the user clicks "Discard" out of panic.

**Root cause (`pdfarranger.py`, line 1808):**

`set_unsaved(True)` is called unconditionally when any internal DnD gesture completes, before checking whether the page order actually changed:

```python
self.set_unsaved(True)   # fires before any position check
```

This also conflicts with GNOME HIG ("only present a save dialog if the document has actually been modified") and ISO 9241-110 Principle 4 (error tolerance).

**Proposed fix:** Compare page order before and after the operation:

```python
old_order = [model.get_value(model.get_iter(Gtk.TreePath(i)), 0).filename
             for i in range(model.iter_n_children(None))]
# ... existing insert/remove operations ...
new_order = [model.get_value(model.get_iter(Gtk.TreePath(i)), 0).filename
             for i in range(model.iter_n_children(None))]
if old_order != new_order:
    self.undomanager.commit("Move" if move else "Copy")
    self.set_unsaved(True)
```

**Steps to reproduce:**
1. Open any multi-page PDF
2. Click a page thumbnail, move mouse 3-5 px, release without intentional reorder
3. Close pdfarranger -> save dialog appears despite no change

oxidworks:
Fixes #1363

When an internal drag-and-drop gesture completed without actually reordering pages (e.g. a micro-drag caused by motor impairment or hand tremor), `set_unsaved(True)` was called unconditionally, before checking whether the page order had changed at all.

This fix captures the page order before and after the DnD operation and only commits to the undo history and marks the document as unsaved if the order actually changed.

### DIFF
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -2234,6 +2234,13 @@ class PdfArranger(Gtk.Application):
         target = selection_data.get_target().name()
         if target == 'MODEL_ROW_INTERN':
             move = context.get_selected_action() & Gdk.DragAction.MOVE
+            if move and ref_to:
+                indices = sorted(int(p) for p in data)
+                target_idx = ref_to.get_path().get_indices()[0]
+                insert_pos = target_idx if before else target_idx + 1
+                continuous = len(indices) == indices[-1] - indices[0] + 1
+                if continuous and indices[0] <= insert_pos <= indices[-1] + 1:
+                    return
             self.undomanager.commit("Move" if move else "Copy")
             self.set_unsaved(True)
             data.sort(key=int, reverse=not before)
--- a/pdfarranger/pdfarranger.py
+++ b/pdfarranger/pdfarranger.py
@@ -2214,6 +2214,15 @@ class PdfArranger(Gtk.Application):
             return
         selection_data.set(selection_data.get_target(), 8, data.encode())
 
+    @staticmethod
+    def _is_noop_move(data, ref_to, before):
+        """Check if a move would not change page order."""
+        indices = sorted(int(p) for p in data)
+        target_idx = ref_to.get_path().get_indices()[0]
+        insert_pos = target_idx if before else target_idx + 1
+        continuous = len(indices) == indices[-1] - indices[0] + 1
+        return continuous and indices[0] <= insert_pos <= indices[-1] + 1
+
     def iv_dnd_received_data(self, iconview, context, _x, _y,
                              selection_data, _target_id, etime):
         """Handles received data by drag and drop in iconview"""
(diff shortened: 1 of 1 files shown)


kbengs:
That does not work for some reason. Does it work for you?

Another way to fix it I can think of is to compare self.model[npage][0].__repr__() instead.

oxidworks:
You're right, thanks for catching that. I traced through it and the reason my check never fires on `move` is that `model[i][0]` returns the `Page` object. `Page` in [core.py:324](https://github.com/pdfarranger/pdfarranger/blob/main/pdfarranger/core.py#L324) doesn't define `__eq__`, so list comparison falls back to identity. The `move` branch calls `model.remove(...)` after the inserts, which gives the rows new Python wrappers. So `old_order != new_order` is always `True` even on a no-op drag, and `set_unsaved(True)` still runs.

Comparing `__repr__()` works because [Page.__repr__](https://github.com/pdfarranger/pdfarranger/blob/main/pdfarranger/core.py#L339-L342) serializes the full state (nfile, npage, zoom, angle, scale, crop, hide, size, description, layerpages). It's stable across row recreation. I'll push an updated commit using that approach.

kbengs:
The unsaved flag work fine now but it does not store the undo state. That's because the undomanager.commit need to be before the action is done...

Maybe the approach need to be: Check before the action: If the moved page range is continuous, if it is a move action and the drop position is within the page range. Is so return and do nothing.

oxidworks:
Rewrote the fix based on your suggestion. It now checks before the action whether the moved page range is continuous and the drop position falls within that range. If so, it returns early. This keeps undomanager.commit() before the action as required and avoids the undo state issue.

dreua:
Now with the new version I have a hard time understanding what the code does. Could that be made more clear somehow? At least a comment or refactoring into a method with appropriate name might help in the future to understand what the code is supposed to achieve. On the other hand I don't want to gatekeep to much and the same is true for many bits of the existing code :thinking:

oxidworks:
The new lines are not covered by tests. I could extract the check into a helper method and add a unit test for it. Should I, or is it fine without?

oxidworks:
Extracted the check into _is_noop_move() so the intent is clear from the name.

kbengs:
That looks good to me. I don't think a unit test is necessary.

kbengs:
Thanks for the fix!

PR #1370, opened 2026-04-19
Transform Content, Montage, Add Text (pdftl integration)

### LINKED ISSUE #1367: Add Rotation by Degrees (and perhaps other advanced transformations)
Currently, PDF Arranger supports basic rotation options (90°, 180°, 270°) and simple transformations like cropping and resizing. However, it does not support more advanced use cases, such as rotating pages by arbitrary degrees (e.g., 45°, 30°) or applying skew transformations. These features are available in tools like Inkscape and would significantly improve PDF Arranger’s flexibility, especially for users working with scanned documents that require post-processing.

I’m unsure if the libraries used by PDF Arranger support these features. If not, I understand that implementation would be difficult. However, I want to emphasize that there is currently no tool within the GNOME ecosystem that provides these post-processing capabilities. For example, GNOME Document Scanner itself does not support such operations. This makes adding these features to PDF Arranger particularly valuable for the ecosystem.

pdftl-dev:
This PR introduces three new features using the (optional) pdftl library.

* Transform Content: An interactive dialog to shift, scale, and spin page content, with a graphical preview and mouse-drag/mousewheel support.

* Montage: Combine multiple selected pages onto a singe page with adjustable rows, columns, and gutters.

* Add Text: adjustable font size, opacity, rotation, and positioning. Supports variables like {page}, {total}, {date}, allowing Bates stamping.

If pdftl is not installed, the associated menu actions are disabled.

Fixes #1367 (arbitrary rotation) and #1368 (montage) 

To install pdftl with the minimum dependencies for this: `pip install "pdftl[add-text]"`

### DIFF
--- a/data/menu.ui
+++ b/data/menu.ui
@@ -178,6 +178,10 @@ along with PDF Arranger.  If not, see <http://www.gnu.org/licenses/>.
           <attribute name="action">win.rotate</attribute>
           <attribute name="target" type="i">90</attribute>
         </item>
+	<item>
+          <attribute name="label" translatable="yes">_Transform content…</attribute>
+          <attribute name="action">win.place</attribute>
+	</item>
         <item>
           <attribute name="label" translatable="yes">_Page Size…</attribute>
           <attribute name="action">win.page-size</attribute>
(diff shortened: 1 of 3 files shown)


dreua:
@kbengs what do you think? I'm currently against merging this.

PR #1372, opened 2026-04-21
Rebuild PDF outlines on export (single-file, export_doc path only)

### LINKED ISSUE #339: PDFArranger delete Bookmarks
Hello,
with the "Foxit Reader" for Linux, i create bookmarks in a pdf.
After this i must insert a new pdf - for this i use PDFArranger.
If i then open the pdf with "Foxit Reader" again, all my bookmarks are deleted.
Is this a bug of PDFArranger or a problem about the pdf standard?

Greetings from germany
Heiko

pdftl-dev:
Bookmarks from the source PDF(s) are now remapped to reflect the new page positions rather than being silently dropped. For this PR, this is applied only in the export_doc path (i.e. when start_with_empty is set), for single-file output.

The new `exporter_outlines.py` module handles:

- Bookmarks pointing to reordered, subsetted, or duplicated pages
- Named destinations (remapped with a fN- prefix to avoid collisions when merging multiple files)
- Multi-file merges (bookmarks from all source files are combined)
- Nested bookmarks (parents with no valid destination are retained if they have surviving children)
- GoTo actions as well as direct destinations

Fixes #339, #1348

EDIT:
This option is used when "Preserve document information from the first file opened" is unchecked in the preferences dialog.

### DIFF
--- a/pdfarranger/config.py
+++ b/pdfarranger/config.py
@@ -351,8 +351,12 @@ class Config(object):
         if self.has_pikepdf8:
             frame4 = Gtk.Frame(label=_("Saving/exporting to single file"), margin=8)
             cb_retain = Gtk.CheckButton(
-                label=_("Retain document-level information of the first opened file"),
+                label=_("Preserve document information from the first file opened"),
                 margin=8)
+            cb_retain.set_tooltip_text("\n".join([
+                _("When checked: use document properties from the first file opened."),
+                _("When unchecked: merge bookmarks from all documents.")
+            ]))
             cb_retain.set_active(not self.start_with_empty())
             frame4.add(cb_retain)
             d.vbox.pack_start(frame4, False, False, 8)
--- a/pdfarranger/exporter.py
+++ b/pdfarranger/exporter.py
@@ -400,6 +400,10 @@ def export_doc(pdf_input, pages, mdata, files_out, quit_flag, test_mode=False):
         return
     if isinstance(files_out[0], str):
         # Only needed when saving to file, not when printing
+        if len(files_out) == 1:
+            # Imported here to avoid circular import with exporter_outlines
+            from . import exporter_outlines
+            exporter_outlines.rebuild_outlines(pdf_input, pdf_output, pages)
         mdata = metadata.merge_doc(mdata, pdf_input)
     if len(files_out) > 1:
         for n, page in enumerate(pdf_output.pages):
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Upgrade pip
       run: pip install --upgrade pip
     - name: Install
-      run: pip3 install -v --upgrade .[image] coverage
+      run: pip3 install -v --upgrade .[image] coverage pytest
     - name: AppStream
       run: appstreamcli validate $HOME/.local/share/metainfo/com.github.jeromerobert.pdfarranger.metainfo.xml
     - name: Validate Desktop File
@@ -25,9 +25,11 @@ jobs:
       run: python3 -X tracemalloc -u -m unittest -v -f tests.test
     - name: Exporter Tests and Coverage
       run: python3 -m coverage run --data-file=.coverage.exporter -m unittest -v -f tests.test_exporter
+    - name: Exporter Outlines Tests and Coverage
+      run: python3 -m coverage run --data-file=.coverage.exporter_outlines -m pytest -v tests/test_exporter_outlines.py
     - name: Core Tests and Coverage
       run: python3 -m coverage run --data-file=.coverage.core -m unittest -v -f tests.test_core
-    # Convert to lcov because it's compatible with codecov AND can be combine
+    # Convert to lcov because it's compatible with codecov AND can be combined
     - name: Convert to lcov
       run: python3 -m coverage combine && python3 -m coverage lcov
     # Upload for later use (see upload-to-codecov job)
@@ -44,11 +46,13 @@ jobs:
     steps:
     - uses: actions/checkout@v6
     - name: Install
-      run: pip3 install -v .[image]
+      run: pip3 install -v .[image] coverage && pip3 install -v --ignore-installed pytest
     - name: Dogtail Tests and Coverage
       run: python3 -X tracemalloc -u -m unittest -v -f tests.test
     - name: Exporter Tests and Coverage
       run: python3 -m coverage run --data-file=.coverage.exporter -m unittest -v -f tests.test_exporter
+    - name: Exporter Outlines Tests and Coverage
+      run: python3 -m coverage run --data-file=.coverage.exporter_outlines -m pytest -v tests/test_exporter_outlines.py
     - name: Core Tests and Coverage
       run: python3 -m coverage run --data-file=.coverage.core -m unittest -v -f tests.test_core
     - name: Convert to lcov
--- a/.github/workflows/manual.yml
+++ b/.github/workflows/manual.yml
@@ -49,3 +49,5 @@ jobs:
         run: python3 -X tracemalloc -u -m unittest -v -f tests.test
       - name: Exporter Tests
         run: python3 -X tracemalloc -u -m unittest -v -f tests.test_exporter
+      - name: Exporter Outlines Tests
+        run: python3 -X tracemalloc -u -m unittest -v -f tests.test_exporter_outlines
--- a/pdfarranger/exporter_outlines.py
+++ b/pdfarranger/exporter_outlines.py
@@ -1,0 +1,212 @@
+# pdfarranger is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+"""Rebuild PDF outlines after page reorder, subset, or merge."""
+
+import warnings
+import pikepdf
+import decimal
+
+
+class OutlineRemapper:
+    """
+    Maps source-file bookmark destinations to their new locations in the output PDF.
+
+    Constructed once per export with the full list of input PDFs and the output
+    page sequence. Internally builds:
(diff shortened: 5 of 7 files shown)


kbengs:
Thanks for looking into this. Some random thoughts:

Does this make "export_doc_job" obsolete? https://github.com/pdfarranger/pdfarranger/pull/462

With "Retain document level information of the first opened file" unchecked, I would expect that nothing is preserved. So some change would be needed here. https://github.com/pdfarranger/pdfarranger/commit/453f19996e8570719d6e70d00f59044220874606

I see that links on overlay pages are still lost. (copy paste a page with links as overlay on a blank page) Can something be done to that?

Does this work with pikepdf >= 6 ?

I will look more as soon as I have time.

pdftl-dev:
This PR does not make the `job` export path obsolete. `job` is supposed to preserve hyperlinks for the first document, and this PR does not have that feature. It only targets document outlines, aka bookmarks. 

A similar approach should work for merging with hyperlinks.  There is code in pdftl which could be viewed as proof of concept for this. But that is more work. 

So `job` could be replaced in the future, maybe, if outlines and hyperlinks are the only things we care about. 

There are other document-level structures to consider too like OCGs, forms, etc which `job` might retain from the first document, and this PR's path ignores. Merging logic of some sort for all this could all be implemented in python, in principle, but it's all extra work. Copying from one input only (mimicking `job`) would probably be easier, as a stop-gap.

On your other questions:

- Does 'drop hyperlinks, preserve outlines' tally with your 'link on overlay' observation or is that something else?

- Agreed, the preferences UI label was step with behaviour, I've updated that now. 

- I made a couple of compatibility fixes and the exporter_outlines tests now pass with pikepdf 6.0.1. Will other versions get tested in CI?

kbengs:
> Does 'drop hyperlinks, preserve outlines' tally with your 'link on overlay' observation or is that something else?

With this pdf:
https://files.danfoss.com/download/Drives/Vacon%2020%20Complete%20Manual-DPD00716H1-UK.pdf?id=%2Fsites%2FDDS%2DCZSK%2FShared%20Documents%2FLiterature%5FCustomers%2FVACON%2FVACON%2020%2FVacon%2020%20Complete%20Manual%2DDPD00716H1%2DUK%2Epdf&parent=%2Fsites%2FDDS%2DCZSK%2FShared%20Documents%2FLiterature%5FCustomers%2FVACON%2FVACON%2020

1. Open pdf
2.  Select all
3. Change page size with "scale & add margins" (this will add the pages as overlay)

-> the outlines are lost


1. Open pdf
2. Select all
3. Change page size with "Scale" 

-> the outlines links go to slightly wrong location

pikepdf-10.5.1 is tested in the "build"
https://github.com/pdfarranger/pdfarranger/actions/runs/24875264319/job/73082133822?pr=1372#step:9:1246

pdftl-dev:
Ah, I'll look into that at some point. Does the `job` path retain outlines after "scale & add margins"? I would expect adding pages as overlays to destroy outlines, always. Maybe the issue is that scaling and adding margins shouldn't use this technique? It isn't needed, you can inject a CTM and change the mediabox.

Good point on the scaling though, I will need to fix it.

kbengs:
> Does the job path retain outlines after "scale & add margins"? 

No they are lost then also.

> Maybe the issue is that scaling and adding margins shouldn't use this technique?

You might be right.

The snap fails from time to time. It might work next time.

The test_exporter_outlines.py does not run on CI or am I wrong?

For now I don't see more issues.

pdftl-dev:
> The test_exporter_outlines.py does not run on CI or am I wrong?

You're right - I updated the CI workflow and I think the tests should now run. 

> For now I don't see more issues.

Great.

Only thing I noticed is that your test file had all outlines collapsed, and that was getting lost on save. So I fixed that: outline state (style, open/closed) should now be copied over.

kbengs:
I can prepare a updated docker for the `build-pikepdf-6` which has `pytest` installed if that is needed?

For the `build` you probably only need to add `python3-pytest` to install.yml

pdfarranger/exporter.py
```diff
@@ -398,6 +398,10 @@ def export_doc(pdf_input, pages, mdata, files_out, quit_flag, test_mode=False):
     _copy_n_transform(pdf_input, pdf_output, pages, quit_flag)
     if quit_flag is not None and quit_flag.is_set():
         return
+    if len(files_out) == 1:
```
kbengs: You could move this to the `if isinstance(files_out[0], str)`, then it will not run when printing.

pdfarranger/config.py
```diff
@@ -351,8 +351,12 @@ def preferences_dialog(self, parent, handy_available):
         if self.has_pikepdf8:
             frame4 = Gtk.Frame(label=_("Saving/exporting to single file"), margin=8)
             cb_retain = Gtk.CheckButton(
-                label=_("Retain document-level information of the first opened file"),
+                label=_("Preserve document information from the first file only"),
                 margin=8)
+            cb_retain.set_tooltip_text("\n".join([
+                _("When checked: use document properties from the first file."),
```
kbengs: Should it be "...first opened file." to be more clear?

kbengs (commented):
We'll se if manual.yml is successful after merge.

kbengs (approved):
Looks good to me

kbengs:
Is it okay to merge this?

pdftl-dev:
I think so

kbengs:
For anyone wondering how to use this feature: uncheck "Preserve document information from the first file opened" in the preferences dialog.

kbengs:
@pdftl-dev I noticed for this PDF the outlines are not saved
https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://files.danfoss.com/download/Drives/Vacon-NXS-NXP-User-Manual-DPD01218F-FI.pdf&ved=2ahUKEwjL14irpdSUAxWXQVUIHcAsGEMQFnoECA4QAQ&usg=AOvVaw3ZSGG0YaVjkPMoOgDqVBT1

pdftl-dev:
@kbengs true. ~~I think it is not a spec compliant PDF file, but anyway~~ I will submit a PR to fix this.

dreua:
> For anyone wondering how to use this feature: uncheck "Preserve document information from the first file opened" in the preferences dialog.

Is that supposed to be the final solution or do we consider this a testing phase and there will be a more clear solution later? At the moment I don't think those behaviors and the need to uncheck the checkbox will be clear to the average user.

dreua:
Okay, there already is a section in the manual, this is good: https://github.com/pdfarranger/pdfarranger/wiki/User-Manual#preserving-outlines-bookmarks-and-links

However I don't think that the function of this PR is really in there, is it? I mean the benefit of having the checkbox unchecked for only one file where it can preserve more outlines than otherwise.

pdftl-dev:
> Okay, there already is a section in the manual, this is good: https://github.com/pdfarranger/pdfarranger/wiki/User-Manual#preserving-outlines-bookmarks-and-links
> 
> However I don't think that the function of this PR is really in there, is it? I mean the benefit of having the checkbox unchecked for only one file where it can preserve more outlines than otherwise.

@dreua I agree, the UI is strange. This reflects the strangeness of the code: we have two export paths which do different things, with some overlap between them, and neither is optimal in every case. It is not clear to me what an ideal UI would be here. For example, would it make sense to have a radio selection with three options, the current two plus 'auto' which uses the 'box ticked' path if there is one input file, and 'box unticked' path otherwise? Or some other heuristic to decide which is better? In any event, we would want to find better wording to convey what the options actually mean.

dreua:
> It is not clear to me what an ideal UI would be here.

Then this should be discussed first in my opinion. Before starting to code. This code is not there for itself but to serve a purpose, to be used by users who need to understand what it is doing, otherwise it is not useful at all in my opinion.

kbengs:
I did not see it as the final solution when merging it, but still good enough. If something is not good we can always improve it. By time we usually see better ways of how to do things. We might also get feedback from users.

The radio button that pdftl-dev suggests could be an improvement. Then maybe also a "do not preserve anything" alternative should be there. Any comments @dreua on that?

Later I would hope that we could get rid of either the job-path or the other.

pdftl-dev:
I wonder if we can add UI elements to the save dialog? That seems the natural place for a 'save mode' selection to me. EDIT seems do-able... is it a good idea?
```diff
     def choose_export_pdf_name(self, exportmode):
-        """Handles choosing a name for exporting """
+        """Handles choosing a name for exporting with an added export options dropdown"""
         title = _('Save As…') if exportmode == 'ALL_TO_SINGLE' else _('Export…')
 
-        chooser = Gtk.FileChooserNative.new(title=title,
-                                        parent=self.window,
-                                        action=Gtk.FileChooserAction.SAVE,
-                                        accept_label=_("_Save"),
-                                        cancel_label=_("_Cancel"))
+        # 1. Switch to Gtk.FileChooserDialog to allow extra widgets
+        chooser = Gtk.FileChooserDialog(
+            title=title,
+            parent=self.window,
+            action=Gtk.FileChooserAction.SAVE,
+            buttons=(
+                _("_Cancel"), Gtk.ResponseType.CANCEL,
+                _("_Save"), Gtk.ResponseType.ACCEPT
+            )
+        )
         chooser.set_do_overwrite_confirmation(True)
+
+        # 2. Create the Dropdown (ComboBoxText)
+        dropdown_box = Gtk.Box(orientation=Gtk.Orientation.HORIZONTAL, spacing=10)
+        dropdown_label = Gtk.Label(label=_("Output bookmarks, hyperlinks etc:"))
+
+        mode_combo = Gtk.ComboBoxText()
+        # Append pairs of (internal_id, user_visible_label)
+        mode_combo.append('USE_PREFS', _('Default (per user preferences)'))
+        mode_combo.append('USE_FIRST', _('Transfer from first PDF opened'))
+        mode_combo.append('USE_ALL', _('Transfer bookmarks (and maybe hyperlinks) from all input PDF files'))
+        mode_combo.append('DISCARD', _('Discard'))
+
+        mode_combo.set_active_id('USE_PREFS')
+
+        # (Optional) Update filters dynamically if the user changes the dropdown
+        def on_mode_changed(combo):
+            selected_id = combo.get_active_id()
+            # You can clear and re-add filters here if you want them to react instantly
+            pass
+        mode_combo.connect("changed", on_mode_changed)
+
+        dropdown_box.pack_start(dropdown_label, False, False, 0)
+        dropdown_box.pack_start(mode_combo, True, True, 0)
+        dropdown_box.show_all()
+
+        # 3. Inject the dropdown into the dialog
+        chooser.set_extra_widget(dropdown_box)
+
```

dreua:
> I wonder if we can add UI elements to the save dialog?

That is problematic because we might use system specific file dialogs on Windows/mac, also for Flatpak file portals (not currently in use but may be necessary in the future) that customization is not possbible. It is unfortunate, I wish there was a better way to do it, too.

dreua:
If the current state is meant as a open beta for this feature that is fine with me, that wasn't clear from reading this PR.

PR #1373, opened 2026-04-21
Add "Transform content" dialog to shift, scale and spin page content

### LINKED ISSUE #1367: Add Rotation by Degrees (and perhaps other advanced transformations)

pdftl-dev:
Fixes #1367 (with no additional dependencies)

### DIFF
--- a/pdfarranger/pageutils.py
+++ b/pdfarranger/pageutils.py
@@ -21,11 +21,13 @@ import locale
 
 from math import pi
 
-from .core import Sides, Dims, PDFRenderer
+from .core import Sides, Dims, PDFRenderer, PlaceTransform
 from .exporter import get_in_memory_poppler_doc
 
 _ = gettext.gettext
 
+PT_PER_MM = 72 / 25.4
+MM_PER_PT = 25.4 / 72
 
 def scale(model, selection, factor):
     """Set the scale factor of a selection of pages."""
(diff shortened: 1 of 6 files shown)


kbengs:
Interesting. I just had a quick look.

* Can the rotation be in smaller steps than 1 degree?
* Imo rotation in + angle should go to the right. That would match behavior on Gimp.
* Maybe the shift in mm?

I'll look more into it later but trying to focus on the two other PRs first.

pdfarranger/pageutils.py
```diff
@@ -1166,3 +1166,140 @@ def on_changed(self, *args):
...
+            self.shift_x.set_value(self._start_shift_x + dx / zoom)
+            self.shift_y.set_value(self._start_shift_y - dy / zoom)
+            return True
+        return False
+
+    def _on_scroll(self, widget, event):
+        target_widget = self.spin if (event.state & Gdk.ModifierType.SHIFT_MASK) else self.scale
+        step = target_widget.get_adjustment().get_step_increment()
+        current = target_widget.get_value()
+        if event.direction == Gdk.ScrollDirection.UP:
```
kbengs, line 1284: My micro-step-scrolling mouse will emit `Gdk.ScrollDirection.SMOOTH`
Then you have to `up = event.get_scroll_deltas()[2] > 0`

You can have a look at sw_scroll_event() in DrawingAreaWidget

pdfarranger/pageutils.py
```diff
@@ -1166,3 +1168,145 @@ def on_changed(self, *args):
...
+            dx = event.x - self._drag_start_x
+            dy = event.y - self._drag_start_y
+            zoom = self.preview_page.zoom
+            self.shift_x.set_value(self._start_shift_x + (dx / zoom) * MM_PER_PT)
+            self.shift_y.set_value(self._start_shift_y - (dy / zoom) * MM_PER_PT)
+            return True
+        return False
+
+    def _on_scroll(self, widget, event):
+        target_widget = self.spin if (event.state & Gdk.ModifierType.SHIFT_MASK) else self.scale
```
kbengs, line 1281: shift is used in iconview and DrawingAreaWidget to scroll view horizontally. I think Alt is not used for anything.

kbengs (commented):
The transformations will get lost on the export_doc_job path, any thoughts on that?

The preview has some glitches, were it does not update the view. I haven't found out exactly how to reproduce it.

Anyway, a impressive PR!

kbengs (commented):
The transformations will get lost on the export_doc_job path, any thoughts on that?

The preview has some glitches, were it does not update the view. I haven't found out exactly how to reproduce it.

Anyway, a impressive PR!

pdftl-dev:
> The transformations will get lost on the export_doc_job path, any thoughts on that?

How about we disable the menu item if the `job` checkbox  is ticked? A corner case would be if that box is ticked after transforming... guess we would have to disallow that.

> The preview has some glitches, were it does not update the view. I haven't found out exactly how to reproduce it.

Oh,  maybe there is a better way. I wondered about faking the preview (rendering once or on zoom out only, then using 'local' graphics transformations), which might be faster and maybe less glitchy. Don' know if it's feasible yet.

kbengs:
Can it be made to work on the export_doc_job path also?

pdftl-dev:
I found additional bugs and am a bit discouraged about whether it is feasible to implement this feature properly. It seems like it would require much more extensive changes than I anticipated.

The problem is that other operations can affect this one, and the operations do not commute. pdfarranger doesn't store a list of operations, it stores parameters needed for the final export only, and these are applied in some canonical order. To keep this model and add this transformation feature, we'd have to add logic to many of the existing pdfarranger operations to correct the transformation parameters when the user applies operations in a non-canonical order, so that the effect of using canonical order with the corrected parameters gives the correct result.

To see the effect of this sort of bug: rotate a page and then shift it right. Then rotate another page after shifting it right. Currently the effect is the same, but  it should not be.

kbengs:
Okay, I see. The Page class has the rotate() function which rotates crop, hide etc on every rotate action. I think rotation of the place parameters need to be added here as well.

pdftl-dev:
Yes. Also in every other operation which changes geometry.

pdftl-dev:
Closing for now at least.

Issue #257, opened 2020-08-04
[Feature request] Add support for encryption functionalities


Nokia808:
Hi.
QPDF contain very powerful encryption functions - see bellow link:
http://qpdf.sourceforge.net/files/qpdf-manual.html#ref.encryption-options

It is useful to have GUI for these functionalities through pdfarranger. 

I would like to suggest the following UI:

1) like "Edit properties", the proposed "Encryption" options should be in the same upper right gear list,
2) "Encryption" options should be just above "Edit properties" option,
3) clicking of proposed "Encryption" option should result in new window - as for "Edit properties" - but certainly should be larger in size,
4) within new window (in point 3) there should be choices of encryption option arranged as sub-option then bellow it it's choices as check boxes, where user can check only one of these boxes like, for ex, the following (please notice that I can not draw check boxes here):

"Key length":
40 bit | 128 bit | 256 bit

"Printing":
Allowed | Disallowed
Full | low               (Note: this line should be active only if "Allowed" is checked by user & key length is set to 128 or 256 bit. It should be inactive if user set key length to 48, or if select "Disallowed")

"Document modification":
Allowed | Disallowed

"Text/image extraction":
Allowed | Disallowed

"Annotations (Comments and form fill-in and signing)":
Allowed | Disallowed

"Rotation and reordering of pages":
Allowed | Disallowed

"Accessibility to visually impaired":
Allowed | Disallowed

"Rotation and reordering of pages":
Allowed | Diallowed

"Filling form fields":
Allowed | Disallowed

And bellow all the above, there should be 2 check boxes:
[Encrypt cleartext-metadata]
[Use AES encryption will be used instead of RC4 encryption]

Lastly & bellow all above, there should be field for passwords as following:
"User password":            (to be set by user)

"Owner password":        (to be set by user)

Kindly, notice that some of these functionalities should be freeze (not be available to user) or being active (available to user) depending on key length selected. That is to say, if user slelect 48 bit key length then the following should be freeze:
"Rotation and reordering of pages"
"Accessibility to visually impaired"
"Filling form fields"
[Encrypt cleartext-metadata]
[Use AES encryption will be used instead of RC4 encryption]

But if user select key length to be 128 bit then, all parameter should be active & available to user.
If user select key length to be 256 bit, then also all parameter should be active & available but [Use AES encryption will be used instead of RC4 encryption] check box must be enforcedly checked .....

Nokia808:
[...]

PR #1375, opened 2026-04-21
Add GUI to Control PDF Password Encryption

pirate0815:
In Regards to #1144 and #257 this PR implements the ability to set and clear password from pdf files.
If any initial opened pdf files contain a password, that is set as the pdf passwort for export.
The user can clear (no encryption applied when exporting) or set a new password via a dialog, the current password state is visualized via the icon of the password / encryption dialog button.

Currently all password protected exports are hard coded to use AES-256 (R6).


pirate0815:
With the last commit the unit test of the exporter fails.
Visually the files are identical, but the pdf version  is now 1.6 instead of 1.3 as it was previously and so the files do not match.

pdfarranger/exporter.py
```diff
@@ -410,19 +412,37 @@ def export_doc(pdf_input, pages, mdata, files_out, quit_flag, test_mode=False):
...
     else:
         if isinstance(files_out[0], str):
             if not test_mode:
                 _set_meta(mdata, pdf_input, pdf_output)
             _remove_unreferenced_resources(pdf_output)
         if test_mode:
-            pdf_output.save(files_out[0], qdf=True, static_id=True, compress_streams=False,
-                stream_decode_level=pikepdf.StreamDecodeLevel.all,
-                min_version=max_version,
-            )
```
kbengs, line 427: Would this be more clean like this:
```
encryption = pikepdf.Encryption(user=password, owner=password, R=6 if password else False
pdf_output.save(files_out[0], qdf=True, static_id=True, compress_streams=False,
                    stream_decode_level=pikepdf.StreamDecodeLevel.all,
                    min_version=max_version,
                    encryption=encryption
                )
```

or:
```
if password:
  encryption = pikepdf.Encryption(user=password, owner=password, R=6)
else:
  encryption = False
```

Same could be done on all places in your code.

pdfarranger/pdfarranger.py
```diff
@@ -3292,6 +3305,54 @@ def clear_find_results(self, unselect_all):
...
+    def get_password(self):
+        self.show_all()
+        result = self.run()
+        text = self.entry.props.text
+        self.destroy()
+        if result == Gtk.ResponseType.OK:
+            return text
+        else:
+            return self.password
+
```
kbengs, line 3381: one blank row here.

pdfarranger/pdfarranger.py
```diff
@@ -3292,6 +3305,54 @@ def clear_find_results(self, unselect_all):
...
+        if new_password:
+            self.password = new_password
+            icon = Gtk.Image.new_from_icon_name("channel-secure-symbolic", Gtk.IconSize.BUTTON)
+            button.set_image(icon)
+        else:
+            self.password = None
+            icon = Gtk.Image.new_from_icon_name("channel-insecure-symbolic", Gtk.IconSize.BUTTON)
+            button.set_image(icon)
+
+class EncryptionPasswordDialog(Gtk.Dialog):
```
kbengs, line 3347: The dialog is too small. On KDE I can't see the title of the dialog.

pdfarranger/pdfarranger.py
```diff
@@ -1483,6 +1489,13 @@ def on_action_open(self, _action, _param, _unknown):
                     if not added:
                         break
                 adder.commit(select_added=False, add_to_undomanager=True)
+
```
kbengs: This does not work with import or drag and drop. But maybe that is fine?

kbengs (commented):
Thanks.

pdfarranger/pdfarranger.py
pirate0815: I could not replicating the issue in a debian vm with kde (wayland + x11).
As a change i now set the entry width to 15 characters, but i do not know if it fixes the issue on your system.
Screenshot From 2026-04-23 08-56-44

pdfarranger/pdfarranger.py
pirate0815: I did not think about people using import / drag and drop with an "empty" application.
Fixed by setting a password from input files when the pdfque was empty before the new pages where added.

pdfarranger/exporter.py
pirate0815: cleaned up

pdfarranger/pdfarranger.py
pirate0815: Fixed

pirate0815:
I have tried to update the test for the exporter.
However the output of the pdf files has been finicky.
Would it be acceptable to use a visual comparison tool like diff-pdf instead of the line by line textual approach currently employed?
Maybe in conjunction with a comparing of document structure when convert to qpdf json.

pdfarranger/pdfarranger.py
kbengs: This is how it looks now

Image

pdfarranger/exporter.py
```diff
@@ -482,21 +497,22 @@ def export_doc_job(pdf_input: List[pikepdf.Pdf], files: List[List[str]], pages:
             _set_meta(mdata, pdf_input, outpdf)
             outpdf.pages.append(page)
             _remove_unreferenced_resources(outpdf)
-            outpdf.save(files_out[n], min_version=max_version)
+            outpdf.save(files_out[n], min_version=max_version, encryption=encryption)
     else:
         if isinstance(files_out[0], str) and not test_mode:
             _set_meta(mdata, [pdf_output], pdf_output)
-        job.write_pdf(pdf_output)
```
kbengs, line 493: Why can job.write_pdf be removed?

pdfarranger/pdfarranger.py
```diff
@@ -1633,11 +1642,15 @@ def on_action_import(self, _action, _param, _unknown):
             adder = PageAdder(self)
             filenames = chooser.get_filenames()
             filenames = reversed(filenames) if os.name == 'nt' else filenames
+            empty_before_import = len(self.pdfqueue) == 0
             for filename in filenames:
                 added = adder.addpages(filename)
                 if not added:
                     break
             adder.commit(select_added=False, add_to_undomanager=True)
+            # Try to set a password from the input files if no pdf was already opened
```
kbengs, line 1655: Wouldn't it be better to remove the empty_before_import test, and just always call set_password_from_input_files? (everywhere)

kbengs (commented):
Users who use the app to remove the password from the pdf will need to click the padlock and then ok. That is not very intuitive and will be a extra step that not all users will be happy with.

Should the set_password_from_input_files be optional, a setting in preferences? 

What do you think @dreua ?

pdfarranger/pdfarranger.py
pirate0815: I believe it would be unexpected for the user if after first opening a unencrypted pdf file (Import, Open, Drag and Drop, Pasting etc.) that adding another pdf with encryption would set the password.
This is similar to the current behavior where the exported pdf is only encrypted if the first document opened is also encrypted.

If you strongly believe that it would be better user experience to always apply encryption if any imported document is encrypted i would be happy to implement it that way.

pdfarranger/exporter.py
pirate0815: Removing job.write_pdf and replaced it by saving the pdf from job.create_pdf allowed me to me to set the encryption directly.
I had believed that the behavior from doing it that way would be equivalent in execution, however after looking at source code in pikepdf and qpdf it turns out that my belief was wrong.
The changes have been reverted and encryption / decryption when using qpdf jobs is now set in the json once more.

pdfarranger/pdfarranger.py
pirate0815: Hopefully increasing the size of the entry to 30 characters should fix it

pirate0815:
> Users who use the app to remove the password from the pdf will need to click the padlock and then ok. That is not very intuitive and will be a extra step that not all users will be happy with.

I have modified the interaction so that clicking on a closed padlock clears the password
clicking on the open password brings up the password change dialog.

Due to reverting the changes to job.write() the export test succeed once again.

pdfarranger/pdfarranger.py
```diff
@@ -3292,6 +3318,65 @@ def clear_find_results(self, unselect_all):
...
+        if new_password:
+            self.password = new_password
+            icon = Gtk.Image.new_from_icon_name("channel-secure-symbolic", Gtk.IconSize.BUTTON)
+            button.set_image(icon)
+        else:
+            self.password = None
+            icon = Gtk.Image.new_from_icon_name("channel-insecure-symbolic", Gtk.IconSize.BUTTON)
+            button.set_image(icon)
+
+    def set_password_from_input_files(self):
```
kbengs, line 3340: if user already set a password before opening the first file it should probably return?

pdfarranger/pdfarranger.py
kbengs: Ok, it is fine as you have done it.

pdfarranger/exporter.py
```diff
@@ -221,7 +221,7 @@ def _remove_unreferenced_resources(pdfdoc):
     try:
         pdfdoc.remove_unreferenced_resources()
     except RuntimeError:
-	# Catch "RuntimeError: operation for dictionary attempted on object of
```
kbengs, line 224: you should remove this as it is unrelated

pdfarranger/pdfarranger.py
kbengs: It's fine now.

kbengs (commented):
The test is failing with invalid password
https://github.com/pdfarranger/pdfarranger/actions/runs/24927957134/job/73003250424

I did not look into that.

dreua:
> What do you think @dreua ?

I don't have the resources to look at this in depth and haven't run it. Does it add a lock icon for encryption permanently to the tool bar? I'm don't think this function is important enough for such a prominent placement.

I have never used PDF Arranger for de- or encrypting PDFs as all of the PDFs I handle are not encrypted. (Maybe actually I removed password protection once or twice now that I think about it. And that is probably what most users will be doing.)

Finally, is encryption actually the correct term or is it more like password protection? I think the PDF standard has stuff like protecting from printing which is technically useless (shame on whoever came up with that). I'd like PDF Arranger to keep a correct and clear language about those things: If it is not encryption then it shall never bear that name.

kbengs:
The placement is prominent I agree. Maybe the padlock could be in the statusbar, all to the left or right? Then it maybe also would need to be in the menu, as users might not realize that the padlock is clickable.

To my understanding Encryption is actually what it does. I think "Password protection" is more common in UI. But I am fine with "Encryption".

dreua:
I think it should just go to the menu, maybe after the "Edit properties" item?

We don't have statistics obviously but from following this issue tracker since the very beginning I'd say it is just not important enough to be always on screen, I really can't remember much requests about passwords at all. I mean if it would only be shown if password protection is used in the current document then that would be fine by me, but having it on screen all the time is too distracting imo. Especially if we want to roughly follow the Gnome design principles.

pirate0815:
I have moved the action from a button to a (toggle) menu.
Adjusted tests so that they hopefully execute successfully.

pdfarranger/exporter.py
kbengs, line 457: How will this json["password"] behave now when the "encrypt" is set?

pdfarranger/pdfarranger.py
```diff
@@ -476,10 +478,15 @@ def __create_actions(self):
...
             ("find_next", self.searchbar_widget.find_next),
-            ("find_all", self.searchbar_widget.find_all),
+            ("find_all", self.searchbar_widget.find_all)
         ]
         self.window.add_action_entries(self.actions)
 
+        password_toggle = Gio.SimpleAction.new_stateful("password-setup", None, GLib.Variant.new_boolean(False))
+        password_toggle.connect("change-state", self.password_setup)
+        self.window.add_action(password_toggle)
+
```
kbengs, line 488: one extra blank row

pdfarranger/pdfarranger.py
```diff
@@ -476,10 +478,15 @@ def __create_actions(self):
             ("find", self.searchbar_widget.find),
             ("find_prev", self.searchbar_widget.find_prev),
             ("find_next", self.searchbar_widget.find_next),
-            ("find_all", self.searchbar_widget.find_all),
```
kbengs, line 480: this is not related to the PR

kbengs (commented):
I don't know how the old "password" will behave, do you?

Then could you fixup the commit history so that everything is in one or a few commits? (and force push)

kbengs:
The history clean up can be done at merging also if you feel that is an issue. But I'd like to understand if the json["password"] does anything anymore.

pirate0815:
Sorry for the late answer, i had personal stuff in my life.

From my understanding of qpdf json format (`qpdf --job-json-help` and https://qpdf.readthedocs.io/en/stable/qpdf-job.html#qpdf-job)  it is used as the "default" password for reading (if no per page password is set ) and if "decrypt" is not set for saving.

pdfarranger/pdfarranger.py
pirate0815: Removed

pdfarranger/pdfarranger.py
pirate0815: Reverted

kbengs (approved):

PR #1388, opened 2026-05-15
Add swap odd/even pages function

KarolS:
Added a function that swaps odd and even pages.

It's a common issue that when scanning a multipage document in a language that goes right-to-left using a two-page scan feature, the scanning software saves the left page first, which is incorrect. The resulting file has pages in the following order (assuming page 1 is the cover and was scanned as a single page):

    1 3 2 5 4 7 6 ...

This feature fixes this by swapping odd and even pages in a selected range of pages.

### DIFF
--- a/data/menu.ui
+++ b/data/menu.ui
@@ -203,6 +203,10 @@ along with PDF Arranger.  If not, see <http://www.gnu.org/licenses/>.
           <attribute name="action">win.reverse-order</attribute>
         </item>
         <item>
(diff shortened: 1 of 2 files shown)


kbengs:
Looks good to me. Only thing is that the menu could have the underline, maybe "S_wap Odd/Even", which allows the use of Alt + W. Don't know how important it is.

kbengs:
Thanks!

PR #1392, opened 2026-05-23
Update pt_BR.po

rondrond:
Updated pt_BR translation with new strings.


po/pt_BR.po
```diff
@@ -802,160 +859,167 @@ msgstr "Ampliar"
...
 msgstr "_Extrair"
 
-#: data/menu.ui:220 data/menu.ui:495
+#: data/menu.ui:307
+msgid "_Find"
+msgstr "_Encontrar"
+
+#: data/menu.ui:310
+msgid "_Find…"
+msgstr "_Encontrar..."
```
kbengs, line 928: Here you should use the elipsis character …

rondrond:
thanks, fixed it

PR #1404, opened 2026-05-27
Update Ukrainian translation

Dymkom:
- add missing row;
- improve translation,


dreua:
Should be squashed before/when merging

Dymkom:
@dreua, done

PR #1410, opened 2026-06-02
Update Simplified Chinese translation

Puqns67:


po/zh_CN.po
```diff
@@ -600,81 +744,104 @@ msgstr "拆分页面"
...
-#: pdfarranger/pageutils.py:333
+#: pdfarranger/config.py:357
+msgid "When checked: use document properties from the first file opened."
+msgstr ""
+
+#: pdfarranger/config.py:358
+msgid "When unchecked: merge bookmarks from all documents."
+msgstr ""
+
+#: pdfarranger/pageutils.py:150
```
dreua, line 835: Are those missing on purpose?

dreua (approved):

po/zh_CN.po
Puqns67: I couldn't find the exact location of these texts in pdfarranger, so I wasn't sure how to translate them.
However, I've now located them by checking the source code and have fixed the issue.

Puqns67:
I also discovered that pdfarranger on Gentoo has an issue where translations don't work. I've identified the cause and am currently waiting for a response from the Gentoo developers.

https://bugs.gentoo.org/976693

dreua:
If the Gentoo devs accept your patch then that is fine with me, if not then please create a new issue about it so we can discuss and solve this. (Wondering what Gentoo is doing so differently than all the other distros where this does not seem to be an issue. Not saying that our code here is perfect.)

PR #1416, opened 2026-06-14
Fix a missing iv_selection_changed case

kbengs:


dreua (approved):
Code looks good to me but there is a test case failing, not sure what that is

PR #1421, opened 2026-06-28
Handle cancelled export job

DieimesNunes:
## Summary

This pull request handles the case where an export job is cancelled before the pikepdf job object is created.

When `quit_flag` is already set, `_create_job()` may return `None`. Previously, `export_doc_job()` continued execution and attempted to call `create_pdf()` on `None`, causing an `AttributeError`.

This change returns early when no job is created.

## Testing

Added a regression test covering cancellation before writing the output file.

The exporter and core test suites were executed successfully:

```text
Ran 30 tests in 0.405s

OK
```

```text
Ran 21 tests in 0.011s

OK
```


kbengs:
I don't know how to reproduce the AttributeError, could you give us the steps?

PR #1430, opened 2026-07-29
feat: add command palette

M4R7iNP:
To improve keyboard-only workflows, I thought a command palette (often `CTRL+K` or `CTRL+SHIFT+P` in other programs) would be useful. I think it turned out pretty good?

I bound it to both shortcuts, but we can choose one?


kbengs:
The code looks good and seems to work well. I am okay with merging it. What do you @dreua think?

I don't know which shortcut is better. But shouldn't it be in the main menu also so that user can be aware that the command palette exists?

The test is failing for reasons not related to this PR.

PR #1431, opened 2026-07-31
Fix filename diacritics issue in Windows save dialog

kbengs:
If its ever needed, below another workaround i tested, which appeared to work with `chooser.set_filename(f)`:

```
    def get_short_path_name(long_name):
    """Converts a Windows path to 8.3-format (removes special characters)

    Works around pdfarranger issue #1420

    If NtfsDisable8dot3NameCreation is set in Registry it will however
    not convert but instead return the original filename.
    """
    output_buf_size = ctypes.windll.kernel32.GetShortPathNameW(long_name, None, 0)
    if output_buf_size == 0:
        return long_name
    output_buf = ctypes.create_unicode_buffer(output_buf_size)
    ctypes.windll.kernel32.GetShortPathNameW(long_name, output_buf, output_buf_size)
    return output_buf.value
```

However `set_current_folder()` and `set_current_name()` seems to work around the problem also.


kbengs:
@pdftl-dev the Outlines test seems to fail after Pikepdf 10.11.0. Would you like to look at it?

pdftl-dev:
> @pdftl-dev the Outlines test seems to fail after Pikepdf 10.11.0. Would you like to look at it?

See #1434

Issue #1265, opened 2025-09-11
Can't open or import files without the correct extension


Leonardo3305:
# PDF Arranger fails to open files from Google Drive via Nautilus (Gnome)
*(Original title, edit by @dreua)*

**Describe the bug**
----------------

PDF Arranger is unable to open PDF files located in cloud directories (specifically, Google Drive) mounted in the Fedora Nautilus file manager. When attempting to open a file from this location, the program displays an error message stating "No such file or directory: '/e-drive:// [...]".

This indicates that the application cannot correctly interpret the virtual paths provided by GVFS (GNOME Virtual File System), which is the technology used by Nautilus to integrate cloud services and other remote locations.

**To Reproduce**
------------

1. In Fedora (with the GNOME environment), integrate a Google Drive account in "Online Accounts".
2. Open the Nautilus file manager and navigate to the Google Drive folder, which appears in the side panel.
3. Try to open a PDF file by dragging it into the PDF Arranger window or by using the "Open with..." option.
4. See the error message being displayed.

**Expected behavior**
-----------------

PDF Arranger should be able to access the file through the GVFS abstraction layer and open it normally in its interface, just as it does with files stored in local directories on the hard drive.

**Input files**
-----------

The bug does not seem to be specific to a particular PDF file, but rather to its location. The error occurs with any PDF file that is accessed through a GVFS mount point in Nautilus.

**System and Versions**
-------------------

- PDF Arranger version and where you get it: 1.12.1, obtained via Flathub directly from the Gnome Store (Flatpak version).
- Uses: libqpdf 12.2.0, pikepdf 9.10.2, GTK 3.24.50, and Python 3.12.11.
- OS name and version: Fedora Linux 42 (Workstation Edition) with the GNOME desktop environment.

dreua:
> Have you tried mounting your google drive via rclone? I wonder if that would work.

(quoting myself from the related issue)

If it turns out to be reproducible with rclone I might be able to work on it, otherwise you have a workaround for your problem 😉 



> cloud directories (specifically, Google Drive)

If there are actually others besides Google Drive I'd also like to hear about it.

mffm-NET:
I once got a similar problem (on Win11 with other programs) with an abandoned G-Drive Folder from a second account. It didn't hold the true files even thay look alike and show thumbnails as copy but as a softlink file pointing to the cloud and/or google writer,calc... webtools.
[...]

Issue #1253, opened 2025-08-15
'Unknown file format' when dragging jpg from Google Drive mount on Ubuntu 24.04


therobyouknow:
Describe the bug
----------------

PDF arranger says with a pop-up 'Unknown file format' when dragging jpg from Google Drive mount on Ubuntu

To Reproduce
------------

1.  Go to a folder on google drive, standard mounted in Files in Ubuntu
2.  Select the jpgs
3.  Drag them into the PDF Arranger Window
4.  pop-up 'Unknown file format' 
5. Workaround: copy files to a local folder on the machine, drag them into PDF Arranger, no error. 

Expected behavior
-----------------

Be able to drag jpgs in directly from Google Drive folder.

Input files
-----------

Cannot provide the specific files due to sensitive info, however would expect same issue with other jpgs pasting in from google drive.

I think it's due to the implementation of Google Drive in Files, in particular the file path implementation - it's just not implemented - instead the Google File id is used, so you get a long alphanumeric string.

Screenshots
-----------

System and Versions
-------------------

-   PDF Arranger version 1.10 - not sure where I got it.
-   pikepdf version - how do I find this?
-   OS name and version: Ubuntu 24.04

dreua:
I don't feel like spending time on workarounds or implementation of any commercial non standard APIs, sorry. Fuse mounts and all the regular stuff you find on unix filesystems should work.

The more sarcastic side of me thinks that you should ask Google to implement support for PDF Arranger 😉

PR #1433, opened 2026-08-01
Allow open of pdf/images without extension on Linux

kbengs:


kbengs:
I verified that it fixes the issue with google drive files in #1265 and #1253

Gio worked on my Windows 11 machine but then not on Windows 10. It will read the mime type from the registry and apparently is not reliable enough. It also does not read the magic bytes on Windows. Therefore still using mimetypes on Windows.

PR #1434, opened 2026-08-02
fix for pikepdf 10.11 outlines changes

pdftl-dev:
## Fix bookmark styling loss on pikepdf ≥ 10.11.0

pikepdf 10.11.0 silently drops any non-standard outline dictionary key (`/C`, `/F`, etc.) set via `OutlineItem(obj=...)` or via direct `.obj[key] =` mutation while still inside `open_outline()`'s `with` block. `_insert_tree_node` relied on exactly that, so bookmark color/bold/italic (and any other custom keys) were silently lost on newer pikepdf, while destinations/titles/hierarchy still worked — masking the bug.

**Fix:** stop passing `obj=` at construction. Instead stash non-managed keys on the new item and reapply them via `item.obj[key] = value` only after the outline's `with` block fully closes — confirmed safe on both 10.5.1 and 10.11.0, so no version branching needed. `/A` is deliberately excluded since it's superseded by the already-remapped `destination=`.

Also fixed `test_styles_and_closed_state_preserved`'s source fixture, which had the same bug (mutating `.obj` before `.append()`'s enclosing `with` block closed).


pdftl-dev:
This PR was prepared with AI assistance

